PR #19, opened 2017-08-22
Added http error handler and changed the error propagation logic

### LINKED ISSUE #12: Create http error handler middleware

lmammino:
This PR adds an HTTP error handler based on the package http-errors, which allows to throw http exceptions from any part of the code and the handler will be able to create a proper response for the user.

Also the error propagation logic in middlewares has been changed to keep propagating the error to any middleware regardless if any of them handled the error or not.

This enables more flexible behaviours where multiple error handlers can contribute to change the response (e.g. add headers, serialization, etc.)

Thanks to @dkatavic for the precious contribution

Closes #12


src/middy.js
```diff
@@ -72,7 +72,7 @@ const runErrorMiddlewares = (middlewares, instance, done) => {
   const runNext = (err) => {
     try {
       if (!err) {
-        return done()
+        instance.__handledError = true
```
peterjcaulfield: can you explain this prop? I can't see it documented anywhere?

lmammino: Good point. It's not even easy to explain.

What happens with this new logic is that, in case of error, the execution is moved to the error middlewares stack. In the previous version, as soon as one error middleware handled the error (by not propagating the error in the `next()` function), the execution stopped and the response was returned.

Now we thought it makes more sense to keep executing all the other error middlewares so that they have a chance to keep decorating the response (e.g. add headers, serialize the body, etc.)

With this new approach we need a way to figure out if the error was somehow handled so that we can decide wether the execution global execution of the lambda needs to be considered failed (unhandled error) or successful (handled error). We use that flag to do that

lmammino: I am thinking now that we should probably reset that flag at some point. Let me review this before moving the PR forward

lmammino: Just updated the code @peterjcaulfield, @dkatavic! Please have  another look when you can.

lmammino:
@peterjcaulfield are you happy to proceed with the merge on this?

@techfort, do you want to have a peek as well?

lmammino:
@dkatavic can you have a look here when you have a minute?

dkatavic:
@lmammino Everything seems fine. I am merging

lmammino:
Awesome, thanks @dkatavic! 🇭🇷

PR #23, opened 2017-08-23
Feature/validator

techfort:
@lmammino the PR is kind of finished (maybe a bit of jsdoc'ing left to do) however, for some reason the `after` validator middleware is not returning anything so whenever i test for correctness of the response, i get undefined, can you take a look? just run `npm run test` to see what i'm talking about


src/middlewares/__tests__/validator.js
```diff
@@ -0,0 +1,130 @@
...
+const validator = require('../validator')
+
+describe('📦  Middleware Validator', () => {
+  test('It should validate an incoming object', () => {
+    const handler = middy((event, context, cb) => {
+      cb(null, event.body) // propagates the body as a response
+    })
+
+    const schema = {
+      required: ['headers', 'body'],
```
lmammino: Are we sure `body` is mandatory? it might be stripped in GET requests (as it happens with `pathParameters` and `queryStringParameters` when they are not available)

src/middlewares/validator.js
```diff
@@ -0,0 +1,31 @@
+const createError = require('http-errors')
+const Ajv = require('ajv')
+const ajvKeywords = require('ajv-keywords')
+
+const ajv = new Ajv({v5: true, $data: true, allErrors: true})
+ajvKeywords(ajv)
+
+module.exports = (inputSchema, outputSchema, options) => ({
```
lmammino: I am currently thinking that, in order to easily allow both `inputSchema` and `outputSchema` to be optional and to be more consistent with other middlewares, we should probably accept an option object that contains `inputSchema` and `outputSchema` as keys.

With this approach we can also expose some other configuration layers for `ajv` (e.g. by having a `ajvOptions` key). In such case we should initialize `ajv` inside the middleware of course.

src/middlewares/validator.js
```diff
@@ -0,0 +1,31 @@
...
+  before: (handler, next) => {
+    if (!inputSchema) {
+      return next()
+    }
+    const validate = ajv.compile(inputSchema)
```
lmammino: Would it make more sense, for performance, to pre-compile the schema, outside the `before` function (so that we can reuse across multiple calls)

src/middlewares/validator.js
```diff
@@ -0,0 +1,31 @@
...
+    if (!valid) {
+      throw new createError.BadRequest('Event object failed validation', validate.errors)
+    }
+    next()
+  },
+  after: (handler, next) => {
+    if (!outputSchema) {
+      return next()
+    }
+    const validate = ajv.compile(outputSchema)
```
lmammino: As above, we might consider to move this outside the `after` function

lmammino:
Also, before merging we should update the README (the handlebars version) by adding the middleware there and the [middlewares module file](/src/middlewares/index.js)

src/middlewares/__tests__/validator.js
techfort: @lmammino in reality this is just an example to test the strictness of the validation. I could have put in a completely arbitrary schema like `const schema = { properties: { foo: { type: 'date'}}}`. It does reflect the use that will be made of the `event`. `inputSchema` is completely up to the developer and the use case at hand.

lmammino:
I made a number of changes and I think this is now good to go.

Anyway I  had to merge this branch with the outstanding code in #19, so that should go first!

@peterjcaulfield and @dkatavic can you have a look as soon as you have a minute?

Thanks

src/middlewares/__tests__/validator.js
```diff
@@ -0,0 +1,125 @@
...
+    handler.use(validator({outputSchema: schema}))
+
+    handler({}, {}, (err, response) => {
+      expect(err).toBe(null)
+      expect(response).toBe(expectedResponse)
+    })
+  })
+
+  test('It should make requests with invalid responses fail with an Internal Server Error', () => {
+    const expectedResponse = {
```
dkatavic: In this case, we are not testing the response, so can we rename `expectedResponse` variable to something more accurate and move it in the handler definition (it doesn't need to be in the upper scope)?

lmammino: That is totally correct, my bad! Can you (or @techfort) please apply all the needed corrections?

lmammino:
Thanks for the review @dkatavic and to @techfort for the effort in making this happen :)

src/middlewares/__tests__/validator.js
lmammino: I will do these fixes now @dkatavic @techfort

PR #29, opened 2017-09-13
Feature/do not wait for empty callback

dkatavic:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.2.4",
+  "version": "0.3.0",
```
lmammino: let's keep this `0.2.5` for now. I now it's not strictly semantic versioning (as we are adding new features here and not just doing bug fixes), but I'd like to keep the version count "small" before an official release.

src/middlewares/__tests__/doNotWaitForEmptyEventLoop.js
```diff
@@ -0,0 +1,32 @@
+const middy = require('../../middy')
+const doNotWaitForEmptyEventLoop = require('../doNotWaitForEmptyEventLoop')
+
+describe('🥃 Do Not Wait For Empty Event Loop', () => {
```
lmammino: Great logo for this test!

docs/middlewares.md
```diff
@@ -237,3 +237,26 @@ handler(event, {}, (_, body) => {
   })
 })
 ```
+
+## [doNotWaitForEmptyEventLoop](/src/middlewares/doNotWaitForEmptyEventLoop.js)
+
+Sets context.callbackWaitsForEmptyEventLoop property to false. This will prevent lambda for timing out because of hanging connection
```
lmammino: please put context.callbackWaitsForEmptyEventLoop and false in backticks so that they will look like this:

`context.callbackWaitsForEmptyEventLoop`

`false`

src/middlewares/__tests__/doNotWaitForEmptyEventLoop.js
dkatavic: haha, thx

docs/middlewares.md
dkatavic: nice spot

lmammino:
Awesome @dkatavic! Take all the pleasure in proceeding with the merge 😂

Issue #39, opened 2017-10-02
EISDIR: illegal operation on a directory when doing a build


philipodev:
Got this error when trying to do a build. Tried simple googling but couldnt find a suitable solution.

By first guess is that you try to copy a file to `fonts` which is a folder and cant be overriden.
I think we need to copy like this instead:
`node_modules/jsdoc/templates/default/static/fonts/OpenSans-Bold-webfont.eot' -> 'docs/middy/0.2.11/fonts/OpenSans-Bold-webfont.eot'`

Node Version: `8.5.0`
NPM Version: `5.3.0`

OS: `macOS Sierra`

Terminal Output:
```
$ npm run build

> middy@0.2.11 build /Users/philipandersson/Desktop/middy
> npm run build:docs && npm run build:readme


> middy@0.2.11 build:docs /Users/philipandersson/Desktop/middy
> jsdoc --readme README.md --package package.json --destination docs src

fs.js:1919
  binding.copyFile(src, dest, flags);
          ^

Error: EISDIR: illegal operation on a directory, copyfile '/Users/philipandersson/Desktop/middy/node_modules/jsdoc/templates/default/static/fonts/OpenSans-Bold-webfont.eot' -> 'docs/middy/0.2.11/fonts'
    at Object.fs.copyFileSync (fs.js:1919:11)
    at /Users/philipandersson/Desktop/middy/node_modules/jsdoc/templates/default/publish.js:516:12
    at Array.forEach (<anonymous>)
    at Object.exports.publish (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/templates/default/publish.js:512:17)
    at Object.module.exports.cli.generateDocs (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/cli.js:448:35)
    at Object.module.exports.cli.processParseResults (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/cli.js:399:20)
    at module.exports.cli.main (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/cli.js:240:14)
    at Object.module.exports.cli.runCommand (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/cli.js:189:5)
    at /Users/philipandersson/Desktop/middy/node_modules/jsdoc/jsdoc.js:105:9
    at Object.<anonymous> (/Users/philipandersson/Desktop/middy/node_modules/jsdoc/jsdoc.js:106:3)
    at Module._compile (module.js:624:30)
[...]

Issue #38, opened 2017-10-02
CORS middleware


lmammino:
It would be great to have a middleware that automatically adds the headers needed for enabling CORS in the response object. For example:

```javascript
{
  headers: {
    'Access-Control-Allow-Origin': '*'
  }
}
```

The middleware needs to be configurable with an option to specify different origins (it will default to `*`).

Also, it will be good to have in place the following checks:

  - Add the header only if the current response looks like an API Gateway response
  - Don't override the header if it already exists

lmammino:
This can be considered closed by #40. Thanks again @philipodev !

See also #43 for possible improvements

PR #40, opened 2017-10-02
added cors middleware

philipodev:
Solving issue #38 

Thanks for a great repo and solution to a common problem regarding.
I have added a cors middleware with a simple option `origin` which defaults to `"*"`.
But left it open for maybe more options later (like `Access-Control-Allow-Headers` etc.)?

Also added a few tests. But somehow jest is a little bit fussy regarding my coverage.

One more note: I wasn't able to do the `precommit` (`build`) script since it threw me an error when copying some fonts to a folder. See #39

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.2.12",
+  "version": "0.2.13",
   "description": "The simple (but cool 😎) middleware engine for AWS lambda in Node.js",
   "main": "src/index.js",
   "scripts": {
@@ -50,5 +50,8 @@
     "http-errors": "^1.6.2",
     "qs": "^6.5.0",
     "querystring": "^0.2.0"
+  },
+  "jest": {
+    "collectCoverageFrom": ["src/**/*.js"]
   }
 }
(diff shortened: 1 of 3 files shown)


src/middlewares/__tests__/cors.js
```diff
@@ -0,0 +1,77 @@
+const middy = require('../../middy')
+const cors = require('../cors')
+
+describe('📦 Middleware CORS', () => {
+  test('It should default to "*"', () => {
```
lmammino: I would rename this test to something a bit more explicit like:

`Access-Control-Allow-Origin header should default to '*'`

src/middlewares/__tests__/cors.js
```diff
@@ -0,0 +1,77 @@
...
+    handler(event, {}, (_, response) => {
+      expect(response).toEqual({
+        headers: {
+          'Access-Control-Allow-Origin': '*'
+        }
+      })
+    })
+  })
+
+  test('It should not override already declared cors header', () => {
```
lmammino: Here as well might be more explicit something like:

`It should not override already declared Access-Control-Allow-Origin header`

philipodev:
Thank you @lmammino for this great opportunity and great project.

Thanks for the feedback. Very valid points. I have rebased and commited changes.

lmammino:
Fantastic work @philipodev, thanks to you for pushing this feature through so timely and with high quality code.

I'll version bump your changes and get this merged ASAP.

Thanks again

PR #53, opened 2017-11-26
Add typings definition to work better with typescript

joseSantacruz:
Hi,

The first version of the typings definition, this cover the typings of the core methods, after this is merged I'm going to start working on the middlewares typings.


src/middy.d.ts
```diff
@@ -0,0 +1,36 @@
+import {Callback, Context, Handler, ProxyResult} from 'aws-lambda';
+
+export interface IMiddy {
+  use: IMiddyUseFunction;
+  before: IMiddyMiddlewareFunction;
+  after: IMiddyMiddlewareFunction;
+  onError: IMiddyMiddlewareFunction;
+  __middlewares: object;
```
lmammino: I am not sure i would expose this one as it's supposed to be private (only exposed for testing purposes)

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.3.0",
   "description": "The simple (but cool 😎) middleware engine for AWS lambda in Node.js",
   "main": "src/index.js",
+  "types": "./src/middy.d.ts",
```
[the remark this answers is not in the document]
joseSantacruz: Done

package.json
```diff
@@ -48,6 +49,7 @@
     "regenerator-runtime": "^0.11.0"
   },
   "dependencies": {
+    "@types/aws-lambda": "0.0.21",
```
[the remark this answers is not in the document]
joseSantacruz: No, because Dev dependencies are not installed as default so the dev cannot have access to the full type definition

src/middy.d.ts
joseSantacruz: Done

typescript/middy.d.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface IMiddyMiddlewareObject {
+  before?: IMiddyMiddlewareFunction;
+  after?: IMiddyMiddlewareFunction;
+  onError?: IMiddyMiddlewareFunction;
+}
+
+type IMiddyMiddlewareFunction = (
+  handler: IHandlerLambda,
+  next: IMiddyNextFunction
+) => void;
```
lmammino: This, from version 0.3.0, also supports promises as return value (as an alternative to calling the callback directly). Is there anyway we can make this explicit?

typescript/middy.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type IMiddyNextFunction = () => void;
```
lmammino: This function has an optional argument which is an error.

package.json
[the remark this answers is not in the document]
lmammino: Makes sense, thanks for clarifying this.

lmammino:
I am updating the current JSDoc in https://github.com/middyjs/middy/pull/54. It might be helpful here as a reference. 😉

joseSantacruz:
I just made the changes to add Promise on a Middleware function and the error parameter on the next callback

lmammino:
Awesome, thanks :)

I will version bump to 0.3.2 and get this released!

PR #56, opened 2017-12-07
Fix root export on typings and update AWS lambda types

joseSantacruz:
I found an issue using the default export on the typescript types, so I refactor the types to create a signature of the middy default function and add a Namespace with the same name to the rest of the types, so you can use the following imports:

```typescript
import * as middy from 'middy';
import {IMiddy} from 'middy';
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ typings/
 .tern-port
 .DS_Store
 docs/
+/.idea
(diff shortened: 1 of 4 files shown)


lmammino:
Thanks a million @joseSantacruz and sorry if I wasn't super responsive to this 😉

PR #59, opened 2017-12-15
Added support for async await in handler

### LINKED ISSUE #52: Support for async/await in handler
Since version `0.3.0` we have support for `async/await` in middlewares.

We could potentially apply the same approach to the handler. Since the handler is invoked and managed by the middy wrapper, we could support different syntaxes including handlers returning promises and consequently `async/await` handlers.

These are the two syntaxes I envision might be cool to support:

```javascript
const handler = middy((event, context) => {
  return Promise.resolve({
    statusCode: 200,
    body: 'hello'
  })
})
```

and, with `async/await`:

```javascript
const handler = middy(async (event, context) => {
  return {
    statusCode: 200,
    body: 'hello'
  }
})
```

Any thoughts?

I think this is feasible by addressing the handler runtime logic [here](https://github.com/middyjs/middy/blob/master/src/middy.js#L154-L168).

lmammino:
Closes #52 

TODO:

- [x] Add docs and examples
- [x] Update typings for typescript
- [x] Version bump

**READY FOR REVIEW** 🤓

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,8 +283,6 @@ middy(async (event, context, callback) => {
 })
 ```
 
-We believe that the async/await version is easier to read and reason about.
-
 Of course, since AWS lambda runs on Node.js 6.10, you will need to transpile your `async/await` code (e.g. using [babel](https://babeljs.io/)).
 
 
--- a/README.md
+++ b/README.md
@@ -249,6 +249,45 @@ to the user.
 If no middleware manages the error, the lambda execution fails reporting the unmanaged error.
 
 
+### Async/Await support
+
+Middy allows you to write *async/await handlers*. In *async/await handlers* you don't
+have to invoke the callback but just return the output (in case of success) or
+throw an error (in case of failure).
+
+We believe that this feature makes handling asynchronous logic easier to reason about
+and asynchronous code easier to read.
+
+Take the following code as an example:
+
+```javascript
+middy(async (event, context) => {
+  await someAsyncStuff()
+  await someOtherAsyncStuff()
+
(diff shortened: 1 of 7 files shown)


lmammino:
@joseSantacruz can you please verify if the typescript changes here make sense?

README.md
```diff
@@ -249,6 +249,45 @@ to the user.
...
+    .then(() => {
+      return someOtherAsyncStuff()
+    })
+    .then(() => {
+      callback(null, {foo: bar})
+    })
+})
+```
+
+We believe that the async/await version is easier to read and reason about.
```
peterjcaulfield: This sentence is redundant based on sentence above I think.

src/__tests__/middy.js
```diff
@@ -405,6 +405,63 @@ describe('🛵  Middy test suite', () => {
...
+      throw new Error('bad stuff happened')
+    })
+
+    handler({}, {}, (err, response) => {
+      expect(err.message).toEqual('bad stuff happened')
+      endTest()
+    })
+  })
+
+  test('An handler that returns a non-promise should trigger an error', (endTest) => {
```
peterjcaulfield: typo. s/An handler/A handler

README.md
lmammino: Good point!

src/__tests__/middy.js
lmammino: Thanks

dkatavic: What about handling then-able objects?

PR #63, opened 2017-12-19
camelcase urlEncodeBodyParser

DavidWells:
Example has this camelcased

```
const { urlEncodedBodyParser, validator, httpErrorHandler } = require('middy/middlewares')
```

I think it should be camelcased right?

If so we should update the file name as well =)


lmammino:
Great catch David. I believe camel case is more readable. Would you like to update the filename and the docs (where needed) as well?

Thanks

DavidWells:
https://github.com/middyjs/middy/pull/65

PR #80, opened 2018-01-12
Feature/cache middleware

lmammino:
Adds a caching middleware

### DIFF
--- a/src/middlewares/cache.js
+++ b/src/middlewares/cache.js
@@ -22,7 +22,7 @@ module.exports = (opts) => {
           return options.getValue(cacheKey)
         })
         .then((cachedResponse) => {
-          if (cachedResponse) {
+          if (typeof cachedResponse !== 'undefined') {
             return handler.callback(null, cachedResponse)
           }
 
--- a/src/middlewares/cache.js
+++ b/src/middlewares/cache.js
@@ -1,0 +1,38 @@
+const { createHash } = require('crypto')
+
+module.exports = (opts) => {
+  const defaultStorage = {}
+  const defaults = {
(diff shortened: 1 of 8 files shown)


NPCRUS:
Correct me if I wrong please. Looks like this is sort of in-memory cache, and in this case we can't really apply such middleware for lambda functions, because containers will have different cache and other problems. Lambda may not save this between executions. It's better to use some sort of redis or memcache in case of using lambda.
So, maybe you should add some sort of notice to this middleware, that it's not for lambda usage

peterjcaulfield:
@NPCRUS I believe @lmammino is intending to leverage container re-use where possible.

NPCRUS:
@peterjcaulfield Thank you very much for this article, make sense!

lmammino:
@NPCRUS, if you check the documentation here and the example you can see that the middleware is broadly configurable. You can provide your own functions to calculate cache ids, save items in the cache and get them from the cache. It should be relatively easy to write those functions to adopt a remote caching backend if you need that.

src/middlewares/cache.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return ({
+    before: (handler, next) => {
+      options.calculateCacheId(handler.event)
+        .then((cacheKey) => {
+          currentCacheKey = cacheKey
+          return options.getValue(cacheKey)
+        })
+        .then((cachedResponse) => {
+          if (cachedResponse) {
```
dkatavic: can we change this with `if (cachedResponse === undefined) {`? . That way we gonna avoid false negatives when cache is returning `0`, `null` or empty string

lmammino: Good point :)

PR #81, opened 2018-01-12
Feature/lambda warmup middleware

NPCRUS:
Hello guys,

I decided to implement such middleware since i really need it built in your library.
Can you please provide some information how do you run tests and other such things if it's possible( didn't find any documentation about collaboration)

Hope it helps,
Best regards


src/middlewares/index.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   doNotWaitForEmptyEventLoop: require('./doNotWaitForEmptyEventLoop'),
   httpErrorHandler: require('./httpErrorHandler'),
   jsonBodyParser: require('./jsonBodyParser'),
+  lambdaIsWarmingUp: require('./lambdaIsWarmingUp'),
```
lmammino: can we rename this just `warmup`

src/middlewares/lambdaIsWarmingUp.js
```diff
@@ -0,0 +1,15 @@
+module.exports = (config) => {
+  // get isWarmingUp boolean from config object or use approach with serverless-warmup-plugin library
+  const isWarmingUp = event => (config && config.isWarmingUp) || event.source === 'serverless-warmup-plugin'
```
lmammino: I believe `isWarmingUp` should be a function that accepts `event` as an argument so I would change this part as follows (also following the way of handling default options in other middlewares):

```javascript
const defaults = {
  isWarmingUp: (event) => event.source === 'serverless-warmup-plugin'
}

module.exports = (opts) => ({
  before: (handler, next) => {
    const options = Object.assign({}, defaults, opts)
      if (isWarmingUp(handler.event)) {
        // ...
      }
    
    // ...
  }
})
```

src/middlewares/__tests__/lambdaIsWarmingUp.js
```diff
@@ -0,0 +1,45 @@
+const middy = require('../../middy')
+const lambdaIsWarmingUp = require('../lambdaIsWarmingUp')
+
+describe('🥃 Lambda Is Warming Up', () => {
+  test(`Should exit with 'warmup' if config.isWarmingUp boolean is provided`, () => {
+    const handler = middy((event, context, cb) => {
```
lmammino: This test needs to be changed considering that the `isWarmingUp` should be a function accepting `event`.

So we should not have an empty `event` object but an `event` with a field (e.g. `warmup: true`) from which we can deduct if the current invocation is for warm up or for actual business logic execution.

src/middlewares/lambdaIsWarmingUp.js
NPCRUS: I got the idea of passing function instead of boolean into middleware.
That's what I thought when write the issue.
Thanks for suggestion!

lmammino:
Will be merging this soon and release it as `0.7.0`. Thanks a million for your contribution @NPCRUS :)

PR #82, opened 2018-01-12
Fix urlEncodeBodyParser when Content-Type header contains charset

Jolg42:
Decode body when Content-Type contains more than 'application/x-www-form-urlencoded'.
Example: 'application/x-www-form-urlencoded; charset=UTF-8'

In current state urlEncodeBodyParser & validator middleware are useless if the origin is not exactly sending 'application/x-www-form-urlencoded'.
In my case I don't control the origin request unfortunately!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.6.4",
+  "version": "0.6.5",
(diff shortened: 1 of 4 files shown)


lmammino:
Very good catch. I will get this merged after #80

PR #85, opened 2018-01-14
Proper content-type header parsing

### LINKED ISSUE #84: properly parse content-type when it contains encoding info
As discussed in #82 and as [described by W3C](https://www.w3.org/International/articles/http-charset/index), the `Content-Type` header can contain encoding information.

This also affects #75 somehow

Middy middlewares should work seamlessly in this case.

lmammino:
As discussed in #82 and as [described by W3C](https://www.w3.org/International/articles/http-charset/index), the `Content-Type` header can contain encoding information.

This also affects #75 somehow

Middy middlewares should work seamlessly in this case.

Closes #84


src/middlewares/jsonBodyParser.js
```diff
@@ -1,12 +1,16 @@
...
+const contentType = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
-    if (handler.event.headers && handler.event.headers['Content-Type'] === 'application/json') {
-      try {
-        handler.event.body = JSON.parse(handler.event.body)
-      } catch (err) {
-        throw new createError.UnprocessableEntity('Content type defined as JSON but an invalid JSON was provided')
+    if (handler.event.headers && handler.event.headers['Content-Type']) {
```
vladgolubev: So what about lower-case? 

I found that official `aws-serverless-express` relies on lower-case

https://github.com/awslabs/aws-serverless-express/blob/master/index.js#L82-L83

lmammino: Express makes an extra step that *destructively* converts all the headers to lowercase (see [here](https://github.com/nodejs/node-v0.x-archive/issues/1954)).

At the moment, I am not very positive about introducing the same approach in middy as it will slow down the processing of every request (and mutate the event headers, with some unpredictable edge case — what if you have both `content-type` and `Content-Type`, or even `cOnTenT-TyPe`?).

If you want to make sure that specific headers in your app are normalized to the canonical representation, you can see the snippet of code presented in [the previous conversation](https://github.com/middyjs/middy/pull/75#issuecomment-357529352).

vladgolubev: Understand. To clarify, I meant not  *overwriting* headers, but just checking 2 names of the header.

But since this is not a normal situation, I'll probably end up with writing  some kind of `normalize-headers` middleware for my case

lmammino: Yeah, checking 2 headers would potentially work in most of the cases, but it's not implementing the standard properly anyway as `Content-type` (lowercase `t`) will be missed as well. On this respect, I suppose the best solution to be compliant with the spec is to normalize all the headers to lowercase, but that has a cost...

lmammino:
Thanks for the review @dkatavic and @vladgolubev, getting this one merged now and opened a new discussion about a possible idea on normalizing headers in #87. Please have a look

PR #89, opened 2018-01-17
HTTP Header Normalizer Middleware

### LINKED ISSUE #87: NormalizeHTTPHeaders middleware
Following several discussions with @vladgolubev and @dkatavic (see #85 and #75) I think we should have a middleware that makes sure that the headers in the lambda-proxy integration from API Gateway are always exposed in canonical format (dash-separated-camelcasing).

The idea of this middleware would be to go over all the available headers and reconstruct the object with all the keys properly normalized.

Thoughts?

lmammino:
TODO:

- [x] Increase coverage
- [x] Documentation

Closes #87

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -6,6 +6,7 @@
  - [cors](#cors)
  - [doNotWaitForEmptyEventLoop](#donotwaitforemptyeventloop)
  - [httpErrorHandler](#httperrorhandler)
+ - [httpHeaderNormalizer](#httpHeaderNormalizer)
  - [jsonBodyParser](#jsonbodyparser)
  - [s3KeyNormalizer](#s3keynormalizer)
  - [validator](#validator)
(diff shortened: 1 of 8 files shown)


docs/middlewares.md
```diff
@@ -127,6 +128,40 @@ handler({}, {}, (_, response) => {
 ```
 
 
+## [httpHeaderNormalizer](/src/middlewares/httpHeaderNormalizer.js)
+
+Normalizes HTTP header names to their canonical format. Very useful if clients are
+not using the canonical names of header (e.g. `content-type` as opposed to `Content-Type`).
```
vladgolubev: Should we mention middleware adds `rawHeaders` property?

lmammino:
missing headers from [wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields):

 - `X-ATT-DeviceId`
 - `X-UIDH`

missing from [iana.org](https://www.iana.org/assignments/message-headers/message-headers.xml):

 - `ALPN`
 - `C-PEP`
 - `C-PEP-Info`
 - `CalDAV-Timezones`
 - `Content-ID`
 - `DASL`
 - `DAV`
 - `GetProfile`
 - `HTTP2-Settings`
 - `MIME-Version`
 - `Optional-WWW-Authenticate`
 - `Sec-WebSocket-Accept`
 - `Sec-WebSocket-Extensions`
 - `Sec-WebSocket-Key`
 - `Sec-WebSocket-Protocol`
 - `Sec-WebSocket-Version`
 - `SLUG`
 - `TTL`

docs/middlewares.md
lmammino: Very good point, i will do that as well

PR #96, opened 2018-01-20
doNotWaitForEmptyEventLoop middleware: make configurable and make it run before handler, closes #93 

### LINKED ISSUE #93: Move 'doNotWaitForEmptyEventLoop' middleware from 'after' step to 'before'?
Today I added this middleware ⬇️   and I discovered my Lambdas were timing out when they didn't pass validation

https://github.com/middyjs/middy/blob/c9c8e70a154dc67090db1b10be7c0134afad0f71/src/middlewares/doNotWaitForEmptyEventLoop.js#L1-L3

It sets `false`  in `after` step, meaning after handler execution. But if validation fails, handler never runs, and I get a timeout. 

I changed `after` to `before` and it works as expected now.

Any risk making the same change in the framework itself? Was there a reason to do so specifically?

vladgolubev:


docs/middlewares.md
```diff
@@ -363,7 +363,15 @@ handler(event, {}, (_, body) => {
 
 ## [doNotWaitForEmptyEventLoop](/src/middlewares/doNotWaitForEmptyEventLoop.js)
 
-Sets `context.callbackWaitsForEmptyEventLoop` property to `false`. This will prevent lambda for timing out because of hanging connection
+Sets `context.callbackWaitsForEmptyEventLoop` property to `false`. This will prevent lambda for timing out because of open database connections, etc.
+
+### Options
+
+By default middleware sets the property to `false` in `before` step, meaning you can override it in handler to `true` if needed. You can set it in all steps with the options:
```
lmammino: I would change this

> By default middleware sets the property to `false` in `before` step

to 

> By default middleware sets the `callbackWaitsForEmptyEventLoop` property to `false` only in the `before` phase, ...

src/middlewares/doNotWaitForEmptyEventLoop.js
```diff
@@ -1,6 +1,20 @@
...
+module.exports = (opts) => {
+  const defaults = {
+    runOnBefore: true,
+    runOnAfter: false,
+    runOnError: false
+  }
+
+  const options = Object.assign({}, defaults, opts)
+
+  const disableEmptyEventLoopWait = (handler, next) => {
```
lmammino: I like this naming 👍

lmammino:
Merging now, thanks for your valuable contribution @vladgolubev !

Regarding the `/docs` folder, it is excluded because it is used by  jsdoc to generate the API documentation. The `middleware.md` file is the only one included from that folder.

Issue #90, opened 2018-01-17
Postgres connection pool example


DavidWells:
Here is an example of middy middleware connecting to postgres by @dschep 

```js
const pg = require('pg-promise')();
const {getParameter} = require('./ssmParameterStore');

const getDb = () => getParameter('db_password')
  .then((password) => pg({
    host: process.env.DB_HOST,
    port: Number(process.env.DB_PORT),
    database: process.env.DB_DATABASE,
    user: process.env.DB_USER,
    password,
  }));

const dbMiddleware = () => ({
  before: (handler) => getDb().then((db) => {
    handler.context.db = db; 
  }),
  after: (handler) => {
    handler.context.db.$pool.end();
  },
  onError: (handler) => {
    handler.context.db.$pool.end();
    throw handler.error;
  },
})


module.exports = {getDb, dbMiddleware};
```

We should start collecting examples in the wild and housing them somewhere =)

props to @alexdebrie for the find

DavidWells:
Is everyone open to doing something like this https://github.com/serverless/plugins#community-contributed-plugins for middleware we find?

The markdown table is automatically generated via https://github.com/serverless/plugins/blob/master/plugins.json

I can set this up and do a PR

lmammino:
That would be awesome @DavidWells :)

NPCRUS:
What is the purpose of ending the connection after handler, since lambda freeze the closure, so you can reuse existing connection.

I founded a lot of solutions, where connection is setup outside of handler and handler reuses it on every lambda call. And in middleware you can just check existing connection - is it timed out or not?

Best regards

dschep:
@NPCRUS that's correct. However the lambda was timing out. This could probably be rewritten to take advantage of the [`
doNotWaitForEmptyEventLoop`](https://github.com/middyjs/middy/blob/master/docs/middlewares.md#donotwaitforemptyeventloop) middleware to avoid having to close the pool. I just haven't gotten around to it as I'm not using this in a performance sensitive environment. 

Also, @DavidWells, you probably wanna refactor it to pull the creds from `process.env` or something so there isn't a dependency on my (not included) SSM parameter store wrapper :wink:

DavidWells:
@dschep do you want to publish the code/repo for this? I'm missing `'./ssmParameterStore'` file =)

Need something to kick off the list.

Happy to help you setup the same on https://github.com/dschep/lambda-decorators

[...]

PR #97, opened 2018-01-20
SSM Middleware

vladgolubev:
Originates from discussion in https://github.com/middyjs/middy/issues/90#issuecomment-358650963

Offtopic: I needed `invertObject` function and added the one from StackOverflow for now. What about adding `lodash`? The only concern, wouldn't it be too big for such a lightweight wrapper?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,5 +60,6 @@ typings/
 .tern-port
 .DS_Store
 docs/
+!docs/middlewares.md
 
 .idea/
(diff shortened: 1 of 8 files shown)


lmammino:
As you did, I prefer to keep the `invertObject` function locally and not to depend on lodash, if that's the only lodash function we need.

dschep:
Oh yay, now I don't have to maintain my middleware :grin:

PR #101, opened 2018-01-24
Adds normalizeHttpEvent Middleware

### LINKED ISSUE #35: Api Gateway proxy event normalizer middleware
A common issue with Api gateway proxy event is that the keys:

`pathParameters` and `queryStringParameters` are created in the event if (and only if) at least one parameter exists.

this will lead to have to do the following tedious checks in the code:

```
console.log(event.queryStringParameters ? event.queryStringParameters.foo : undefined)
```

I'd love to have a middleware that takes care to normalize the event in this case and make sure we always have those key (mapped to an empty object) so that we don't need to have ternary conditions or a bunch of if statements every time we want to access one of those values in the collection.

i-am-kenny:
This should allow the CORS header to be applied even if the request fails. I can easily add the other headers/tests, but I wasn't sure about some reasonable defaults.

Closes #35


lmammino:
Hey @i-am-kenny, are you ok with the above comment? Do you need any kind of help/support?

lmammino:
Merged, thanks for your precious contribution @i-am-kenny 😉

Issue #64, opened 2017-12-20
Improve Http Error messages


DavidWells:
I altered the `httpErrorHandler` to follow the JSON spec http://jsonapi.org/examples/#error-objects-basics.

Should this change get applied across the board? Or should I just use this as my own custom middleware?

Whats the default lambda integration you guys are using? (we typically use `lambda-proxy` as the serverless framework default)

Here is my altered error handler for nice errors =)

```js
const { HttpError } = require('http-errors')

module.exports = () => ({
  onError: (handler, next) => {
    if (handler.error instanceof HttpError) {
        // as per JSON spec http://jsonapi.org/examples/#error-objects-basics
        handler.response = {
          body: JSON.stringify({
            errors: [{
              status: handler.error.statusCode,
              message: handler.error.message,
              detail: handler.error.details
            }]
        })
      }
      return next()
    }

    return next(handler.error)
  }
})
```

It looks like: 

![image](https://user-images.githubusercontent.com/532272/34184696-0f4a5700-e4d6-11e7-8be6-041650f5af5a.png)

Before with `body: handler.error.message` (no json response)

![image](https://user-images.githubusercontent.com/532272/34184778-6ed6a020-e4d6-11e7-918b-61719891acab.png)

Love this freakin project! Great work 🎉

OlzhasAlexandrov:
For the unexpected 'E', don't you minify the code?
Isn't that the reason why the variable got untrackable?

Great to see an improvement in error tracking though.

DavidWells:
The error is from the response returning a string instead of JSON via the `lambda-proxy` integration of API gateway

lmammino:
Hello (again 😊) @davidwells!

Thanks for the interesting input. Honestly it's my first time seeing the JSON API standard. It sounds pretty interesting but i would like to understand more all the implications before deciding whether it makes sense to adopt it by default in Middy or not.

Regarding our API gateway integration we prefer to use the Lambda proxy as well, as it makes easy to create standard middlewares and integrates well will the serverless framework.

Also, regardless the JSON API spec, error reporting and response serialization are a bit of open discussion at the moment. I would love to find an easy way to keep them separated for higher flexibility (imagine an API that can support both JSON and XML based on the client accept header). So, if you have any thought or ideas on this i'd love any input 😉

DavidWells:
[...]

PR #104, opened 2018-01-28
Middleware HTTP Content Negotiation

### LINKED ISSUE #103: HTTP Content negotiation parser
I am currently working on a middleware that will simplify [HTTP content negotiation](https://tools.ietf.org/html/rfc7231#section-5.3) parsing and easily allow an API gateway Lambda function to support multiple charsets, encoding (e.g. compression algorithms), languages or content-types.

My current approach is to use the module [negotiator](http://npm.im/negotiator) (used also by express and koa) to parse the content negotiation headers and expand the current event.

For example, the middleware will parse the incoming `Accept` header and expose the resulting information in something like `event.preferredMediaTypes`.

A usage example might look like the following:

```javascript
const { httpContentNegotiation } = require('middy/middlewares')

const handler = middy((event, context, callback) => {
  if (event.preferredMediaTypes[0] === 'application/json') {
    return callback(null, {stastusCode: 200, body: JSON.stringify({foo: 'bar'})})
  }

  if (event.preferredMediaTypes[0] === 'application/xml') {
    return callback(null, {stastusCode: 200, body: '<foo>bar</foo>'})
  }

  return callback(null, {statusCode: 415, body: 'Unsupported Media Type'})
})

handler.use(httpContentNegotiation())

module.exports = { handler }
```

This middleware could be the basis to create another middleware that allows to automatically serialize the API Gateway response body into the user preferred format (e.g. JSON vs XML). The combination of both middlewares should be enough to close issue number #64 (Improve HTTP error messages), as it will be easy to automatically serialize the error in the expected format.

I hope I will have a PR ready soon, meanwhile feel free to throw here suggestions or ideas 😉

lmammino:
This PR implements a middleware that will simplify [HTTP content negotiation](https://tools.ietf.org/html/rfc7231#section-5.3) parsing and easily allow an API gateway Lambda function to support multiple charsets, encoding (e.g. compression algorithms), languages or content-types.

My current approach is to use the module [negotiator](http://npm.im/negotiator) (used also by express and koa) to parse the content negotiation headers and expand the current event with data that can be reused by other middlewares or in the handler.

For instance, this middleware can parse the incoming `Accept` header and expose the resulting information in something like `event.preferredMediaType`.

A usage example might look like the following:

```javascript
const { httpContentNegotiation } = require('middy/middlewares')

const handler = middy((event, context, callback) => {
  if (event.preferredMediaType === 'application/json') {
    return callback(null, {stastusCode: 200, body: JSON.stringify({foo: 'bar'})})
  }

  if (event.preferredMediaType === 'application/xml') {
    return callback(null, {stastusCode: 200, body: '<foo>bar</foo>'})
  }

  return callback(null, {statusCode: 406, body: 'Unsupported Media Type'})
})

handler.use(httpContentNegotiation())

module.exports = { handler }
```

This middleware could be the basis to create another middleware that allows to automatically serialize the API Gateway response body into the user preferred format (e.g. JSON vs XML). The combination of both middlewares should be enough to close issue number #64 (Improve HTTP error messages), as it will be easy to automatically serialize the error in the expected format.

Closes #103 

---

## TODO

- [x] Tests
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
   <a href="https://greenkeeper.io/">
     <img src="https://badges.greenkeeper.io/middyjs/middy.svg" alt="Greenkeeper badge"  style="max-width:100%;">
   </a>
-  <a href="https://gitter.im/middyjs">
+  <a href="https://gitter.im/middyjs/Lobby">
     <img src="https://badges.gitter.im/gitterHQ/gitter.svg" alt="Chat on Gitter"  style="max-width:100%;">
   </a>
 </p>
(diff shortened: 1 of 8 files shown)


lmammino:
Since no comment has been made, i am getting this one merged shortly

vladgolubev:
@lmammino  shouldn't this be `0.9.0` by semver?  I'd say this is a new solid feature

lmammino:
@vladgolubev Definitely true, but I'd like to release `0.9.0` together with a middleware that can automatically serialize an object sent as the response (this will provide even more value and help us to close #64).

PR #106, opened 2018-01-31
Add httpEventNormalizer to the list of middlewares

KnupMan:
httpEventNormalizer was missing at the list of the available middlewares. Added to avoid confusions, someone can have a thought that documentation for this one is missing.

Also fixed 2 links of available middlewares as they were broken.


lmammino:
Great catch, thanks a million!

PR #115, opened 2018-02-15
Add partialResponse middleware

kevinrambaud:
I've been using for a couple weeks a really great middleware on `koajs` that allow to add super easily filtering on response for api consumer. It is based on the `json-mask` package that itself inspire from Google partial response filtering notation (https://developers.google.com/+/web/api/rest/#partial-responses)

This PR includes

* the middleware itself
* the unit tests
* the appropriate documentation


kevinrambaud:
hey @lmammino ! great I updated middleware name. Regarding the exchange format note, as you mentioned it only supported json at the moment, but I'm curious to know what others formats you're thinking about.

lmammino:
Great, I will version bump this shortly and get it merged! Thanks for the patience in updating the naming to make it more consistent with the rest of the library.

Regarding other formats, I think Middy should be agnostic, or at least offer facilities to implement multiple "output serialization" formats where needed.

It's still common for many APIs to have to support other formats than JSON, like `xml` or even other formats like `yaml`, `html` or `plain text`.

Since we introduced a content negotiation parsing middleware, we now have an opportunity to create dedicated middlewares that take advantage of this information to serialize the output in one of the available formats.

One idea would be to always return plain objects from the main handler and let a dedicated output serializer middleware to take care of picking the right serialization based on the current request. This module would still work as in its core it operates on the response body (as an object). The part that we might eventually remove is only the serialization one, that will live in a dedicated middleware.

I hope I will have some time in the next few weeks to complete a draft implementation of a generic http response serializer I was trying to implement.

For now this is still a great addition to the framework, so thanks again for your valuable contribution.

PR #124, opened 2018-03-06
Use generic to define the type of the event and response 

joseSantacruz:
A little improvement to the type definition using typescript generics to define the type of the event and response to get better type definition. This change has default types so doesn't break any previous code.


README.md
```diff
@@ -515,11 +515,11 @@ documentation](/docs/middlewares.md)
 
 ## Api
 
-## 
+## Functions
```
lmammino: This will be auto-regenerated by our build process. For some reason the new version removes the titles here, but I don't see it as a big deal for now. I will try to address this in a separate PR

README.md
```diff
@@ -515,11 +515,11 @@ documentation](/docs/middlewares.md)
...
 
 * [middy(handler)](#middy) ⇒ [<code>middy</code>](#middy)
 
-## 
+## Typedefs
```
lmammino: Same here

PR #137, opened 2018-03-24
Fixed wrong path import

### LINKED ISSUE #136: Incorrect import path
https://github.com/middyjs/middy/blob/cc142c392604166b906fc3330a0f9b9e9e25d7d7/middlewares.d.ts#L3

nnyegaard:
Fixed wrong path import https://github.com/middyjs/middy/issues/136

fixes #136


nnyegaard:
Of course, just need to get python installed on the windows machine here. Will update this PR soon.

PR #145, opened 2018-04-01
Add option to validate only POST body in validator middleware

mafrosis:
Hey guys.

I'm automatically deriving POST body schemas from a swagger doc, but I'm unable to use the `validator` middleware since it validates the whole `event` object, and not just the `body` property.

This patch fixes it for me - but I'm not sure if the implementation is what you would like.

### DIFF
--- a/src/middlewares/validator.js
+++ b/src/middlewares/validator.js
@@ -7,7 +7,7 @@ let ajv
 let previousConstructorOptions
 const defaults = {v5: true, $data: true, allErrors: true}
 
-module.exports = ({inputSchema, outputSchema, ajvOptions}) => {
+module.exports = ({inputSchema, outputSchema, ajvOptions, inputBodyOnly}) => {
   const options = Object.assign({}, defaults, ajvOptions)
   lazyLoadAjv(options)
 
@@ -20,7 +20,9 @@ module.exports = ({inputSchema, outputSchema, ajvOptions}) => {
         return next()
       }
 
-      const valid = validateInput(handler.event)
+      // validate only the body payload, or the whole event object
+      const data = inputBodyOnly ? handler.event.body : handler.event
+      const valid = validateInput(data)
 
       if (!valid) {
         const error = new createError.BadRequest('Event object failed validation')


mafrosis:
I forgot to say, if you chaps approve of this approach then I'll add/update the tests!

lmammino:
Hello @mafrosis,
thanks for using middy and for taking the time for submitting a PR.

I am not really convinced about the need to change this and make the validator middleware more complicated.

I am not sure why you can't follow an approach like the following:


```javascript
// sample lambda handler

const validator = require('middy/src/middlewares/validator')
const urlEncodeBodyParser = require('middy/src/middlewares/urlEncodeBodyParser')

const myLambda = middy((event, context, callback) => {
  console.log('The input is valid 😛')
  return callback(null, 'OK')
})

const schema = {
  required: ['body'],
  properties: {
    body: {
      type: 'object',
      required: ['foo', 'bar'],
      properties: {
        foo: { type: 'string' },
        bar: { type: 'string' }
      }
    },
  }
}

myLambda
  .use(urlEncodeBodyParser())
  .use(validator({
    inputSchema: schema
  }))

module.exports = { myLambda }
```

Test code:

```javascript
const myLambda = require('./')

// test
const event = {
  headers: {
    'Content-Type': 'application/x-www-form-urlencoded'
  },
  body: 'foo="hello"&bar=world'
}

myLambda(event, {}, (err, response) => {
  console.log({ err, response })
})
```

Is it just a matter of preference (e.g. having a slightly more concise usage of the middleware) or do you have an actual limitation with the current implementation?

Can you please argument your use case more so that I can understand better your need?

lmammino:
Hello @mafrosis,
thanks for using middy and for taking the time for submitting a PR.

I am not really convinced about the need to change this and make the validator middleware more complicated.

I am not sure why you can't follow an approach like the following:


```javascript
// sample lambda handler

const validator = require('middy/src/middlewares/validator')
const urlEncodeBodyParser = require('middy/src/middlewares/urlEncodeBodyParser')

const myLambda = middy((event, context, callback) => {
  console.log('The input is valid 😛')
  return callback(null, 'OK')
})

const schema = {
  required: ['body'],
  properties: {
    body: {
      type: 'object',
      required: ['foo', 'bar'],
      properties: {
        foo: { type: 'string' },
        bar: { type: 'string' }
      }
    },
  }
}

myLambda
  .use(urlEncodeBodyParser())
  .use(validator({
    inputSchema: schema
  }))

module.exports = { myLambda }
```

Test code:

```javascript
const myLambda = require('./')

// test
const event = {
  headers: {
    'Content-Type': 'application/x-www-form-urlencoded'
  },
  body: 'foo="hello"&bar=world'
}

myLambda(event, {}, (err, response) => {
  console.log({ err, response })
})
```

Is it just a matter of preference (e.g. having a slightly more concise usage of the middleware) or do you have an actual limitation with the current implementation?

Can you please argument your use case more so that I can understand better your need?

mafrosis:
I completely understand your position. I guess I was surprised that the validator schema needed to include the top-level `body` wrapper, but on second thought you could also validate `headers` with the same schema? 

In my case it's a trivial change to our `swagger-to-jsonschema` tool to add the `body` wrapper, so I'm not too bothered if we close this P/R. (and also why I didn't bother with tests before this conversation!)

That said, in general I'm a fan of adding the complexity to libraries which (along with good documentation) then reduce the complexity of my application.

lmammino:
Definitely great point on moving the complexity where needed to keep the application as simple as possible. At the same time, keep in mind that this is a generic framework and i think it's a good practice to keep it simple, generic enough but also extensible for people with very specific needs. My advice would be to write your own specific `bodyValidator` middleware (maybe using the current generic validator as an example), then you should be able to easily re use it across your app, keeping the main app simpler and more declarative. If this sounds like an option i'll close the issue, but feel free to bounce other ideas as they can probably help the project and other readers.

Also if you end up writing your own middleware feel more than welcome to share it, it might be a great starting point for an article or a use case.

Thanks again for your involvement on the project.

mafrosis:
@lmammino All good with me, thanks for putting `middy` out there.

I'm already using a customised version of `httpErrorHandler` in our code, so it's not a problem to write our own `bodyValidator` as you suggest. The nice thing about this approach is that my specific case needs _a lot_ less code than a library version (not surprisingly):

    import createError from 'http-errors';
    import Ajv from 'ajv';
    import ajvKeywords from 'ajv-keywords';

    const validator = (schema) => ({
      before: (handler, next) => {
        let ajv = new Ajv({v5: true, $data: true, allErrors: true});
        ajvKeywords(ajv);

        const validator = ajv.compile(schema);

        if (!validator(handler.event.body)) {
          const error = new createError.BadRequest('Event object failed validation');
          error.details = validator.errors;
          throw error;
        }

        return next();
      },
    });

    export default validator;

I can also now easily extend this to include the custom error messaging rather than outputting the generic message `Event object failed validation`.

PR #155, opened 2018-04-16
Adds error details if there

thomasedwards:
Using the validation middleware, when validation fails you get this error:

    Event object failed validation

But the error message is added to the error, but just not returned. This adds it to the end of the string, if it’s there.

### DIFF
--- a/src/middlewares/__tests__/httpErrorHandler.js
+++ b/src/middlewares/__tests__/httpErrorHandler.js
@@ -20,6 +20,25 @@ describe('📦 Middleware Http Error Handler', () => {
     })
   })
 
+  test('It should create a response for HTTP errors, with details', () => {
+    const handler = middy((event, context, cb) => {
+      const error = new createError.UnprocessableEntity()
+      error.details = 'Try writing tests that pass'
+      throw error
+    })
+
+    handler
+      .use(httpErrorHandler())
+
+    // run the handler
+    handler({}, {}, (_, response) => {
+      expect(response).toEqual({
+        statusCode: 422,
+        body: 'Unprocessable Entity: Try writing tests that pass'
+      })
+    })
+  })
+
   test('It should NOT handle non HTTP errors', () => {
     const handler = middy((event, context, cb) => {
       throw new Error('non-http error')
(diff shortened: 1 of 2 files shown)


thomasedwards:
Have realised there’s a chance that this might print `Event object failed validation: [object Object]` – will see if I can work something out...

lmammino:
This is an interesting PR but, on principle, I am personally against this approach. The reason is that, in production you might end up disclosing sensible information if you don't limit the error message scope.

The current implementation should expose only a generic error message, describing the error type, which is probably a good default. Of course, if you need something more specific you can write and use a custom error handler.

Another approach would be to make the handler support **debug mode** (returns the full error) or **production mode** (return only the message). This would be interesting but make the current implementation more complicated and I prefer to keep the default middlewares as simple as possible.

Let me know what are your thoughts on this.

thomasedwards:
Thanks for the reply!

I should of course added the problem I was actually trying to solve...

1. In development, when using the validation middleware, I would like to see the error stack the validator kicks back. If I have a 100-line long schema, it’s hard to know where the problem might be.
2. In production, I would like to be able to log the error to Sentry. I know that the message is not always suitable for the end user and that validation should take place before the request to Lambda is made – but it would be great to know if there is a bug somewhere.

At the moment I manually edit the validation middleware in the `node_modules` folder locally to print the error out in the console while debugging.

Happy to write something and PR it if you think this is something you’d consider.

Thanks! 😎

lmammino:
Maybe logging the error might be a good idea, what do you think of an approach like the following (warning: untested code ahead):

```javascript
// src/middlewares/httpErrorHandler.js
const { HttpError } = require('http-errors')

module.exports = () => {
  
  const defaults = {
    logger: console.error
  }

  const options = Object.assign({}, defaults, opts)
  
  return ({
    onError: (handler, next) => {
      if (handler.error instanceof HttpError) {
        if (typeof options.logger === 'function') {
          options.logger(handler.error)
        }

        handler.response = {
          statusCode: handler.error.statusCode,
          body: handler.error.message
        }

        return next()
      }

      return next(handler.error)
    }
  })
}
```

This way, by default, the error will be logged with `console.error`, but you can pass `false` to disable the logging or a custom logging function if you need to do something more specific.

Personally, I feel this might be a good balance between flexibility and not adding too much complexity.

I'll submit a PR later with this to see if reaches a broader consensus

PR #156, opened 2018-04-18
Added logger option in httpErrorHandler middleware

lmammino:
This implements a `logger` option in the `httpErrorHandler` middleware as discussed in #155

This also changes the way the exception is checked to avoid issues with environments in which multiple versions of `http-errors` are installed (see [comment](https://github.com/middyjs/middy/pull/156#discussion_r182543535)).

CC: @thomasedwards

### DIFF
--- a/src/middlewares/httpErrorHandler.js
+++ b/src/middlewares/httpErrorHandler.js
@@ -1,16 +1,28 @@
 const { HttpError } = require('http-errors')
 
-module.exports = () => ({
-  onError: (handler, next) => {
-    if (handler.error instanceof HttpError) {
-      handler.response = {
-        statusCode: handler.error.statusCode,
-        body: handler.error.message
+module.exports = (opts) => {
+  const defaults = {
+    logger: console.error
+  }
+
+  const options = Object.assign({}, defaults, opts)
+
+  return ({
+    onError: (handler, next) => {
+      if (handler.error instanceof HttpError) {
+        if (typeof options.logger === 'function') {
+          options.logger(handler.error)
+        }
+
+        handler.response = {
+          statusCode: handler.error.statusCode,
+          body: handler.error.message
+        }
+
+        return next()
       }
 
-      return next()
+      return next(handler.error)
     }
-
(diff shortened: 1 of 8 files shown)


src/middlewares/httpErrorHandler.js
```diff
@@ -1,16 +1,28 @@
...
+module.exports = (opts) => {
+  const defaults = {
+    logger: console.error
+  }
+
+  const options = Object.assign({}, defaults, opts)
+
+  return ({
+    onError: (handler, next) => {
+      if (handler.error instanceof HttpError) {
```
vladgolubev: Just remembered, I had a bug recently when 2 versions of `http-error` library were present in the project.

An error was thrown by version `1.6.3` meanwhile version `1.6.2` existed as a middy dependency.

So `instanceof HttpError` didn't work as `require('http-errors')` resolved to a different package.

Maybe there is a better way to do this check? If you agree let's open a new issue for discussion

vladgolubev: By the way, it was solved by locking down to 1 version of `http-errors` using yarn resolutions https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

lmammino: This is a very interesting point. Any way we can replicate the issue? I am going to have a better look at the HttpError package docs to see if I can find anything useful...

lmammino: This might do the trick:

```javascript
if (e.constructor.super_ && e.constructor.super_.name === 'HttpError') { 
  // ...
}
```

lmammino: @vladgolubev,  I am going to add this change too, so we don't need to deal with a separate PR

thomasedwards:
Great! Looks like it would be quite useful in all sorts of circumstances.

src/middlewares/httpErrorHandler.js
vladgolubev: @lmammino wow, didn't know about this

PR #161, opened 2018-04-22
feat: add in i18n support for validator

willfarrell:
Includes: ajv i18n support (Default: `en`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "types": "./index.d.ts",
   "scripts": {
-    "test:lint": "eslint --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
+    "test:lint": "eslint --fix --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
     "test:typings": "typings-tester --config tsconfig.json index.d.ts middlewares.d.ts",
     "test:unit": "jest --verbose --coverage",
     "test:unit:watch": "jest --verbose --coverage --watch",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.12.1",
+  "version": "0.13.0",
   "description": "🛵 The stylish Node.js middleware engine for AWS Lambda",
   "main": "./index.js",
   "files": [
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "types": "./index.d.ts",
   "scripts": {
-    "test:lint": "eslint --fix --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
+    "test:lint": "eslint --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
     "test:typings": "typings-tester --config tsconfig.json index.d.ts middlewares.d.ts",
     "test:unit": "jest --verbose --coverage",
     "test:unit:watch": "jest --verbose --coverage --watch",
@@ -66,7 +66,6 @@
   "dependencies": {
     "@types/aws-lambda": "^8.10.1",
     "@types/http-errors": "^1.6.1",
-    "accept-language": "^3.0.18",
     "ajv": "^6.0.0",
     "ajv-i18n": "^3.1.0",
     "ajv-keywords": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,9 @@
   "dependencies": {
     "@types/aws-lambda": "^8.10.1",
     "@types/http-errors": "^1.6.1",
+    "accept-language": "^3.0.18",
     "ajv": "^6.0.0",
+    "ajv-i18n": "^3.1.0",
     "ajv-keywords": "^3.0.0",
     "content-type": "^1.0.4",
     "http-errors": "^1.6.2",
--- a/src/middlewares/validator.js
+++ b/src/middlewares/validator.js
@@ -8,11 +8,10 @@ let ajv
 let previousConstructorOptions
 const defaults = {
   v5: true,
-  format: 'full',
   coerceTypes: 'array', // important for query string params
   allErrors: true,
   useDefaults: true,
-  $data: true           // required for ajv-keywords
+  $data: true // required for ajv-keywords
 }
 
 module.exports = ({ inputSchema, outputSchema, ajvOptions, errorFormat = errors => errors }) => {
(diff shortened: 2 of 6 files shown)


src/middlewares/validator.js
```diff
@@ -1,13 +1,40 @@
...
+const defaults = {
+  v5: true,
+  coerceTypes: 'array', // important for query string params
+  allErrors: true,
+  useDefaults: true,
+  $data: true // required for ajv-keywords
+}
+
+module.exports = ({ inputSchema, outputSchema, ajvOptions, errorFormat = errors => errors }) => {
+  const acceptLanguage = require('accept-language')
```
lmammino: Instead re-doing this here, can we delegate somehow the work of processing the languages to the already available (and configurable) [HTTP Content Negotiation middleware](https://github.com/middyjs/middy/blob/master/docs/middlewares.md#httpcontentnegotiation)?

If we do that, we should just assume that the `event` object will contain a property called `preferredLanguage`.

src/middlewares/validator.js
```diff
@@ -1,13 +1,40 @@
...
-const defaults = {v5: true, $data: true, allErrors: true}
+const defaults = {
+  v5: true,
+  coerceTypes: 'array', // important for query string params
+  allErrors: true,
+  useDefaults: true,
+  $data: true // required for ajv-keywords
+}
+
+module.exports = ({ inputSchema, outputSchema, ajvOptions, errorFormat = errors => errors }) => {
```
lmammino: On one side, I like the idea of the `errorFormat` function. On the other, I would prefer to keep this middleware simpler and encourage people to combine it with the usage of the [HTTP Error Handler middleware](https://github.com/middyjs/middy/blob/master/docs/middlewares.md#httperrorhandler).

package.json
```diff
@@ -13,7 +13,7 @@
   ],
   "types": "./index.d.ts",
   "scripts": {
-    "test:lint": "eslint --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
+    "test:lint": "eslint --fix --ignore-pattern='node_modules/' --ignore-pattern='coverage/' --ignore-pattern='docs/' .",
```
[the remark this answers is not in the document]
willfarrell: Sorry, ment to remove this before commit. Habit from a pattern I use where my setup auto formats and lint on commit. No IDE settings or remembering to run lint before commit. Check out `lint-staged` and `husky`.

Will be removed in next commit.

src/middlewares/validator.js
willfarrell: I've been writing a `jsonapi` middleware that essentially replaces `httpErrorHandler`. At the time of writing it made sense to do the formatting here, there have been changes since that might make more sense to move it to another middleware. Let me get back to you on this one.

willfarrell:
Thanks for the great feedback @lmammino. Changes pushed, ready for next round.

src/middlewares/validator.js
```diff
@@ -1,13 +1,44 @@
...
+    'it',
+    'ja',
+    'nb',
+    'pl',
+    'pt-BR',
+    'ru',
+    'sk',
+    'sv',
+    'zh'
+  ])
```
lmammino: Still not sure it's a good approach to list the languages here and not having any opportunity to configure the selection

lmammino:
Hey @willfarrell and others. Please check my suggested implementation and let me know what you think.

I am assuming here that you will use the middleware in combination with the Content Negotiation Middleware, which, if attached before, should initialize the field `event.preferredLanguage`.

I am also supporting the possibility to redefine the default language if needed and normalising the Portuguese language.

If you like this approach, I'll make sure to update the documentation and the typings

src/middlewares/validator.js
```diff
@@ -12,30 +11,31 @@ const defaults = {
...
-    'pl',
-    'pt-BR',
-    'ru',
-    'sk',
-    'sv',
-    'zh'
-  ])
-  if (!results.length) return 'en'
-  return results[0]
+const availableLanguages = Object.keys(ajvLocalize)
```
lmammino: This saves us the trouble of having an hardcoded list of languages

src/middlewares/validator.js
```diff
@@ -12,30 +11,31 @@ const defaults = {
...
+
+/* in ajv-i18n Portuguese is represented as pt-BR */
+const languageNormalizationMap = {
+  'pt': 'pt-BR',
+  'pt-br': 'pt-BR',
+  'pt_BR': 'pt-BR',
+  'pt_br': 'pt-BR'
+}
+
```
lmammino: Makes Portuguese less ambiguous and it will work with all possible variations.
Honestly I am not sure if it's a good idea to go down the loophole of dealing with this kind of things. So far, looking at `ajv-i18n`, only Portuguese was the outlier, so I thought it was easy enough to support all the possible variations.

src/middlewares/validator.js
```diff
@@ -56,9 +56,8 @@ module.exports = ({ inputSchema, outputSchema, ajvOptions }) => {
       if (!valid) {
         const error = new createError.BadRequest('Event object failed validation')
         handler.event.headers = Object.assign({}, handler.event.headers)
-        console.log(chooseLanguage(handler.event))
-        const locale = chooseLanguage(handler.event)
-        ajvLocalize[locale](validateInput.errors)
+        const language = chooseLanguage(handler.event, options.defaultLanguage)
```
lmammino: the default language can now be configured with `options.defaultLanguage`

src/middlewares/validator.js
```diff
@@ -1,13 +1,44 @@
...
-const {deepEqual} = require('assert')
+const ajvLocalize = require('ajv-i18n')
+const { deepEqual } = require('assert')
 
 let ajv
 let previousConstructorOptions
-const defaults = {v5: true, $data: true, allErrors: true}
+const defaults = {
+  v5: true,
+  coerceTypes: 'array', // important for query string params
```
vladgolubev: Hmm, isn't this a breaking change? `coerceTypes` is `false` by default

willfarrell: From a code standpoint, yes. From an end-user standpoint, no. Since this middleware is paired with a json schema, having `coerceTypes ` enabled would make requests more forgiving when a `body` is passed in and enable the ability to parse query strings making them more usable. There are no cases where existing code would fail where they once passed. For example; if someone is using the current defaults their schema would have `"type":"string"` for all their query string params, in this case, no coercion would happen. `coerceTypes` is only applied if the types don't match. I added this option to support the use case when you expect an array from the query string, but only one item was passed in.

PS I wrote the `coerceTypes: 'array'` feature specifically for using ajv in an API middleware like we are here.

willfarrell:
@lmammino I will have time Friday or Saturday to review your suggested changes. At first glance, they look good.

src/middlewares/validator.js
vladgolubev: > For example; if someone is using the current defaults their schema would have "type":"string" for all their query string params, in this case, no coercion would happen. 

Totally makes sense for me. And thanks for adding this, I see how it will be useful in my code already!

willfarrell: @vladgolubev Along these lines, you may also be interested in this PR: https://github.com/epoberezkin/ajv-keywords/pull/64
I just had a new keyword added, `transform`, that allows to you to do some sanitization on strings being passed in. We're drafting up next steps for it here: https://github.com/epoberezkin/ajv-keywords/issues/66

lmammino:
@willfarrell well done. I will try to get this one completed and merged tonight 😉

PR #176, opened 2018-05-13
Refer to root module to correctly bring in types

ossareh:
Hey there,

Currently when I build a project using middy I get the following reported back from `tsc`:

```
$ tsc -p .
node_modules/middy/middlewares.d.ts(4,19): error TS7016: Could not find a declaration file for module './src/middy'. '/Users/ossareh/dev/src/code.famfi.co/mono/famfi.co/mailchimp/node_modules/middy/src/middy.js' implicitly has an 'any' type.
  Try `npm install @types/middy` if it exists or add a new declaration (.d.ts) file containing `declare module 'middy';`
```

This patch fixes this issue:

 - middleware.d.ts references middy using the relative reference `./src/middy`;
 - this bypasses the types defined in `index.d.ts`; and thus causes the issue

### DIFF
--- a/middlewares.d.ts
+++ b/middlewares.d.ts
@@ -1,7 +1,7 @@
 import { SSM } from 'aws-sdk'
 import { Options as AjvOptions } from 'ajv'
 import { HttpError } from 'http-errors'
-import middy from './src/middy'
+import middy from './'
 
 interface ICorsOptions {
   origin: string;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.13.1",
+  "version": "0.13.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.13.1",
+  "version": "0.13.2",
   "description": "🛵 The stylish Node.js middleware engine for AWS Lambda",
   "main": "./index.js",
   "files": [


ossareh:
Since I cannot publish to the project's npm reference this shows up as an error here, however I pushed a change (not included in this PR) which shows the build works fine:

 - https://circleci.com/gh/ossareh/middy/2
 - [config.yml changes](https://github.com/ossareh/middy/commit/ee688a83c52017ad5b284d61e9950b0e292fb2f4)

I've removed this commit from my ossareh/middy/master - so as not to pollute the PR changeset.

lmammino:
I'll get this version bumped and published within few hours. Thanks a million for this fix! 👍

ossareh:
Really appreciate the speedy response @lmammino :)

lmammino:
Merged, the CI will publish the new version in few minutes! 👍 
I'd like to make you aware that we started to work toward version [`1.0.0`](https://github.com/middyjs/middy/milestone/2) where every middleware will be published as an independent module.

Nobody in the core team is an experienced typescript dev, so we'd appreciate any help in this area in case you want to be involved @ossareh :)

ossareh:
I'm definitely down to help out. The easiest way to get types is to write Typescript! :) Since it's pretty much a superset of JS it's very easy to get going, and you can progressively add more specific types as work on the project. 

Obviously as middy gains traction, consideration for a stable set of types up front increases the chance that the goals of v1 (make it easier for people to build a community of middlewares) is successful. So that counters the progressive improvement suggestion above.

Either way, if there's a venue to discuss this more formally let me know.

PR #177, opened 2018-05-15
Fix middy import in definition file

Quelu:
Allow to use middy with Typescript imports

### DIFF
--- a/middlewares.d.ts
+++ b/middlewares.d.ts
@@ -1,7 +1,7 @@
 import { SSM } from 'aws-sdk'
 import { Options as AjvOptions } from 'ajv'
 import { HttpError } from 'http-errors'
-import middy from './'
+import * as middy from './'
 
 interface ICorsOptions {
   origin: string;


ossareh:
@Quelu - can you give an example of the error you get from tsc without this change? 

From my local testing your change seems to be a noop.

Quelu:
I'm having the error `'middy/index' has no default export` when I try to compile with tsc 2.8.3

ossareh:
Hmm, odd. As I understand it (I'm not great at writing these type files for javascript projects) this is actually a default export: https://github.com/middyjs/middy/blob/master/index.d.ts#L41

In my own project this works just fine for me:

handler.ts
```
import middy from "middy";
```

package.json
```
	"dependencies": {
		... snip ...
		"middy": "^0.13.2"
	},
	"devDependencies": {
                 ... snip ...
		"ts-loader": "^4.3.0",
		"tslib": "^1.9.1",
		"typescript": "^2.8.3",
		"webpack": "^4.8.3",
		"webpack-cli": "^2.1.3"
	},
```

tsconfig.json
```
{
	"compilerOptions": {
		"allowSyntheticDefaultImports": true,
		"esModuleInterop": true,
		"experimentalDecorators": true,
		"importHelpers": true,
		"lib": [
			"es2018",
		],
		"moduleResolution": "node",
		"noEmitHelpers": true,
		"noEmitOnError": true,
		"noImplicitReturns": true,
		"target": "es6",
		"sourceMap": true,
		"strict": true,
	},
	"exclude": [
		"node_modules",
	]
}
```

Does your setup look somewhat similar?

Quelu:
I added the allowSyntheticDefaultImports in my tsconfig and It works now !
Thank you for the help ;)

ossareh:
Glad it was helpful @Quelu 👍

astroanu:
I tried what the documentation says for typescript. but i still get that problem.

ossareh:
@astroanu it would be helpful if you provided more information about your situation, for example your tsconfig.json would allow me to provide assistance.

PR #182, opened 2018-05-21
Typescript type improvements

ossareh:
### This is not ready to merge ###

I'm creating this as a reminder to myself to do the remaining work.

Specifically I need tests for the typescript stuff.

I think this means:

 - add typescript
 - add npm task to attempt compile of typescript test code / to validate types 
 - underlying project already has tests, so no need to duplicate them in typescript

ETA: 1 week

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:6.10
+      - image: circleci/node:8.11
 
     working_directory: ~/middy
 
@@ -13,7 +13,7 @@ jobs:
           keys:
           - v1-dependencies-{{ checksum "package.json" }}
           - v1-dependencies-
-      
+
       - run:
           name: configure NPM registry
           command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
@@ -26,11 +26,11 @@ jobs:
           paths:
             - node_modules
           key: v1-dependencies-{{ checksum "package.json" }}
-        
+
       - run:
           name: unit tests
           command: npm test
-          
+
       - run:
           name: coverage reporting
           command: npm run coverage:submit
--- a/package.json
+++ b/package.json
@@ -44,6 +44,9 @@
   },
   "homepage": "https://github.com/middyjs/middy#readme",
   "devDependencies": {
+    "@types/aws-lambda": "^8.10.1",
+    "@types/http-errors": "^1.6.1",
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:6.10
+      - image: circleci/node:8.11
```
vladgolubev: Maybe 8.10, same as in Lambda?

vladgolubev: Or leave 6.10 if middy wants to support 6.10 in Lambda for some time?

ossareh: I'm happy to change it to either, personally since I use the v.8.x line I'd like to ensure my code works on that line, not least because v.8.x uses a different es code that typescript is set to make it's compiled code compatible with.

@lmammino any thoughts on this? Is there someone else that feels very close to this topic that can provide some guidance?

lmammino: I agree, this should be 6.10 as we still support this version of Node.js

ossareh: Sweet, thank you @lmammino and @vladgolubev 

I'll make sure tsconfig targets the emcascript that 6.10 supports also.

lmammino: There's probably some way in circle CI to run the tests in both versions. If that's the case we should integrate v8.10 too. But I would consider this a separate issue

package.json
```diff
@@ -44,10 +44,13 @@
   },
   "homepage": "https://github.com/middyjs/middy#readme",
   "devDependencies": {
+    "@types/aws-lambda": "^8.10.1",
+    "@types/http-errors": "^1.6.1",
+    "@types/node": "^8.0.57",
     "aws-sdk": "2.221.1",
```
lmammino: I am not a TypeScript expert, but are you sure we have to move those to dev dependencies? If we do this they won't be installed while installing middy. So in a middy app, these types libraries won't be directly available unless the user installs them explicitely. Is this a normal behaviour for TypeScript apps/libraries?

ossareh: I'm about 90% sure on this; I'll find a way to verify this - thanks for calling it out.

This weekend was busy for me so I didn't get a chance to look at making these changes just yet. It'll happen this weekend for sure (Sunday is my OSS day).

ossareh:
I'm going to close this and open a couple of other PR's which implement the various small portions of this change.

lmammino:
Thanks, having separate PR will make it much easier to review and get things merged!

PR #183, opened 2018-05-22
Refactor SSM getParametersByPath to work recursively

tommy5dollar:
Hello,

I found out today (the hard way) that [SSM.getParametersByPath](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SSM.html#getParametersByPath-property) only returns up to 10 parameters per call.

It has a MaxResults property but you can't set this to higher than 10.

Therefore I've refactored the middleware to work recursively if there's a NextToken in the response (indicating that there are additional results).

Let me know if you foresee any issues or need me to change anything.

Thanks in advance.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.13.2",
+  "version": "0.13.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


tommy5dollar:
Hey, I've written an additional test to cover the new scenario.

Let me know if you see any other issues

Issue #166, opened 2018-04-25
lowercase headers by default in httpHeaderNormalizer middleware


lmammino:
I found out in multiple instances that "normalized" header names tend to be confusing and unexpected for developers. This is probably because most of the famous web frameworks in the JS landscape prefer to use a lowercased version of the headers name.

Although this would be a breaking change I would say that's an ideal change for version 1.0

We might still optionally produce a hashmap of "normalized" headers based on a boolean flag (but I would set that as `false` by default, to avoid extra computation when not explicitly needed).

As part of this issue, we should take care of updating all the other middlewares that rely on headers.

miki79:
I've tried to work on this issue, but I'm blocked as not really sure how you want to handle the headers of the others middlewares.

For example, if you run the middleware "html-json-body-parser" it expects the header to be "Content-Type", but:
- if you put the "http-header-normalizer" middleware before, the json parser won't work anymore.
- if you modify the json parser to check for "content-type", it will require to have the "header-normalizer" to run before it, so it creates a dependency for the json parser to the header-normalizer

lmammino:
Hello @miki79, glad to know you would like to contribute on this one.
 
My view (not necessarily the right one), is that we should make this a breaking change in respect to the current API (version 0.x) and always normalize to lowercase. I feel this is what happens in all the major frameworks in Node.js and that many people used middy under this assumption and bumped into problems.

Regarding compatibility with the current middlewares, we should update them and use the lowercased version of the headers. Of course, we should explicitly tell in the documentation that the middleware should be used in combination with the header normalizer middleware and add it also in all the code examples shipped with the documentation (maybe also as a peer dependency ???).

Additionally, we can make this dependent middlewares fallback to the canonical representation of every header. So the middlewares themselves won't fail if the client is passing the headers in the canonical format and the server is not using the normalization. This is very brittle but would make middlewares less dependent on the normalizer.

E.g.

```javascript
const contentTypeHeader = handler.event.headers['content-type'] || handler.event.headers['Content-Type']
```

The current list of middlewares that will be affected by this are:

- cors (maybe we should rename this one to http-cors...)
- http-content-negotiation
- http-json-body-parser
- http-urlencode-body-parser

Do you think it's a good idea to proceed under these assumptions?

PR #184, opened 2018-05-22
1.0.0 alpha

miki79:
#166 Modified the middleware "http-header-normalizer" to normalise to lowercase. There is an extra options now in the middleware, so you can decide to normalise to camelCase.

Modified the other middleware:
- cors
- http-content-negotiation
- http-json-body-parser
- http-urlencode-body-parser

to handle headers in lowercase or camel case.

  - [x] I added tests for my changes
  - [x] I updated the documentation to reflect my changes
  - [x] I updated the Typescript typings to reflect my changes
 
I didn't updated the documentation of the middlewares about the handling of both format of header, but I don't think it's necessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "middy-monorepo",
-	"version": "1.0.0-alpha.1",
+	"version": "1.0.0-alpha.3",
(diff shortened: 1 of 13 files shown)


miki79:
Missed an extra semicolon

PR #185, opened 2018-05-22
Http header normaliser to lowercase

### LINKED ISSUE #166: lowercase headers by default in httpHeaderNormalizer middleware

miki79:
Modified the middleware "http-header-normalizer" to normalise to lowercase. There is an extra options now in the middleware, so you can decide to normalise to camelCase.

Modified the other middleware:
- cors
- http-content-negotiation
- http-json-body-parser
- http-urlencode-body-parser

to handle headers in lowercase or camel case.

  - [x] I added tests for my changes
  - [x] I updated the documentation to reflect my changes
  - [x] I updated the Typescript typings to reflect my changes
 
I didn't updated the documentation of the middlewares about the handling of both format of header, but I don't think it's necessary.

Closes #166

### DIFF
--- a/packages/http-header-normalizer/index.js
+++ b/packages/http-header-normalizer/index.js
@@ -35,12 +35,12 @@ module.exports = (opts) => {
     return acc
   }, {})
 
-  const normalizeHeaderKey = (key, camelCase) => {
+  const normalizeHeaderKey = (key, canonical) => {
     if (exceptions[key.toLowerCase()]) {
       return exceptions[key.toLowerCase()]
     }
 
-    if (!camelCase) {
+    if (!canonical) {
       return key.toLowerCase()
     }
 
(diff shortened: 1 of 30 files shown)


packages/http-header-normalizer/index.js
```diff
@@ -49,7 +53,8 @@ module.exports = (opts) => {
   }
 
   const defaults = {
-    normalizeHeaderKey
+    normalizeHeaderKey,
+    camelCase: false
```
lmammino: can we rename this `camelCase` option to canonical?

It isn't really camel case since it will also have dashes (`-` ) in between

miki79:
yeah, I'm not very great with names, that's why I called camelCase. I changed now to canonical. Let me know if there is anything else you want me to change

PR #189, opened 2018-05-25
Add support for cache invalidation to the SSM middleware

theburningmonk:
Updated the `SSM` middleware to support cache invalidation through an optional `cacheExpiryInMillis` config option.

Also refactored the corresponding test a bit to make it easier to run test scenarios across multiple invocations.


theburningmonk:
Will reopen after solving conflicts

PR #190, opened 2018-05-25
add SSM cache invalidation

theburningmonk:
- support cache invalidation for the SSM middleware
- refactored the corresponding tests a bit to make it easier to run scenarios that involves multiple invocations

### DIFF
--- a/src/middlewares/__tests__/ssm.js
+++ b/src/middlewares/__tests__/ssm.js
@@ -3,7 +3,6 @@ jest.mock('aws-sdk')
 const {SSM} = require('aws-sdk')
 const middy = require('../../middy')
 const ssm = require('../ssm')
-const Promise = require('bluebird')
 
 describe('🔒 SSM Middleware', () => {
   const getParametersMock = jest.fn()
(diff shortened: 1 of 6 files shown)


src/middlewares/__tests__/ssm.js
```diff
@@ -35,9 +36,27 @@ describe('🔒 SSM Middleware', () => {
...
+              cb(error, {event, context, response})
+              resolve()
+            } catch (err) {
+              reject(err)
+            }
+          })
+        })
+      }).then(() => {
+        if (delay) {
+          return Promise.delay(delay)
```
lmammino: Are we adding `bluebird` only for this feature (`Promise.delay`)?

Can't we just do something like the following and get rid of `bluebird`?

```javascript
// ...
return new Promise((resolve, reject) => setTimeout(resolve, delay))
// ...
```

I know this is only for dev mode, but I would prefer to avoid to introduce new libs if they don't bring a substantial value.

theburningmonk: sure thing

lmammino:
Awesome :) I will version bump and get this merged shortly. I'll also port these changes to the work in progress 1.0.0 branch!

Thanks again @theburningmonk, great to have you contributing to middy!

theburningmonk:
@lmammino no probs, really appreciate the work you're doing with middy!

lmammino:
Thanks @theburningmonk, but it's not just me, lots of fantastic people are contributing every day to this project... I can barely keep up with it! 😛

PR #194, opened 2018-05-30
Allow cors middleware to specify multiple allowed origins

thejuan:
Changes the cors middleware options to accept an array of origins. If the request origin matches an origin in the options it is returned. Otherwise the first origin is returned.

### DIFF
--- a/src/middlewares/cors.js
+++ b/src/middlewares/cors.js
@@ -1,27 +1,24 @@
 const defaults = {
-  origin: '*', // Can also be an array of whitelisted origins
+  origin: '*',
+  origins: [],
   headers: null,
   credentials: false
 }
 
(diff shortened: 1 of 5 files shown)


src/middlewares/cors.js
```diff
@@ -1,15 +1,26 @@
 const defaults = {
-  origin: '*',
+  origin: '*', // Can also be an array of whitelisted origins
```
vladgolubev: Maybe leave `origin` property as it is, and introduce a new property `origins`? To not mix different data types in 1 property and avoid if/else checks for data types?

thejuan: Pushed changes to add origins

thejuan:
Will do.

PR #195, opened 2018-06-02
Multi Origin Cors

thejuan:
Forward port of feature

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-alpha.8"
+  "version": "1.0.0-alpha.9"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,8 @@
 {
 	"name": "middy-monorepo",
-	"version": "1.0.0-alpha.8",
+	"version": "1.0.0-alpha.9",
 	"lockfileVersion": 1,
+	"requires": true,
 	"dependencies": {
 		"@babel/code-frame": {
 			"version": "7.0.0-beta.46",
@@ -18,9 +19,9 @@
 			"integrity": "sha512-r4snW6Q8ICL3Y8hGzYJRvyG/+sc+kvkewXNedG9tQjoHmUFMwMSv/o45GWQUQswevGnWghiGkpRPivFfOuMsOA==",
 			"dev": true,
 			"requires": {
-				"chalk": "2.4.1",
-				"esutils": "2.0.2",
-				"js-tokens": "3.0.2"
+				"chalk": "^2.0.0",
+				"esutils": "^2.0.2",
+				"js-tokens": "^3.0.0"
 			},
 			"dependencies": {
 				"ansi-styles": {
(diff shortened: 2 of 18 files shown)


packages/http-cors/index.js
```diff
@@ -6,12 +6,19 @@ const defaults = {
 
 const getOrigin = (options, handler) => {
   handler.event.headers = handler.event.headers || {}
-
-  const origin = handler.event.headers['origin'] || handler.event.headers['Origin']
-  if (options.credentials && options.origin === '*' && origin) {
-    return origin
+  const incomingOrigin = handler.event.headers['origin'] || handler.event.headers['Origin']
```
vladgolubev: I believe we can omit `handler.event.headers['Origin']` as all headers are lower-cased?

thejuan: The original code on this branch checks both?

dkatavic: I think we can change the tests to use lowercase only, is that fine @lmammino ?

packages/http-cors/README.md
```diff
@@ -45,6 +45,7 @@ npm install --save @middy/http-cors
 ## Options
 
  - `origin` (string) (optional): origin to put in the header (default: "`*`")
+ - `origins` (array) (optional): An array of allowed origins. The incoming origin is matched against the list and is returned if present.
```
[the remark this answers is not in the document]
thejuan: I can find that file or that build target in this branch?

packages/http-cors/index.d.ts
```diff
@@ -2,6 +2,7 @@ import middy from '../core'
 
 interface ICorsOptions {
   origin: string;
+  origins: string[];
```
ossareh: this should be optional: `origins?: string[]` - to match up with the javascript: https://github.com/middyjs/middy/pull/195/files#diff-3f35fefc85b62544f23244fe1407a252R10

The current ICorsOptions interface is incorrect - I have a fix for that about to be submitted to PR - so no need to fix anything other than this one line imho.

lmammino: Then let's wait for the other PR to be merged and then we'll get this one updated (if needed) and merged too. What do you think? :)

ossareh: If 200 lands, then this won't pass the ci check, if this lands first then 200 will fail because the tests do not look for this value being optional. So I don't think we need to be too strategic around which is merged first :) 

I called it out primarily as I noticed it and wanted to let @thejuan know about my impending PR and how it may affect this work.

ossareh: yeah, this branch should be rebased off master and then any conflicts and test errors should be fixed.

thejuan:
Rebased and made origins optional in typescript definitions

ossareh:
@thejuan thanks!

As a side note I've just opened #204 as a note to myself to improve the tests associated with this middleware.

packages/http-cors/README.md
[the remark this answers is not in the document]
lmammino: This file was removed from this branch

PR #197, opened 2018-06-04
Node8 build

ossareh:
These changes enable building middy against node v6 (current) and node v8 (next). Additionally it leverages the work flow feature of CircleCI to ensure deploys only happen if the code base passes tests on *both* node versions.

This was discussed in #182 with @vladgolubev and @lmammino 

As with the original file we only trigger the deploy job if we're on master; we just utilize the circleci config to control this instead of using a shell script to check the branch name.

I chose not to run coverage for the `build_next` job as the underlying node version shouldn't make a difference to the output here. Also we only take the workspace from v6 and deploy that since nothing in the build vary's based on node version this should be just fine.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
+  build_node6:
     docker:
       - image: circleci/node:6.10
 
@@ -37,7 +37,7 @@ jobs:
           paths:
             - middy/*
 
-  build_next:
+  build_node8:
     docker:
       - image: circleci/node:8.10
 
@@ -94,13 +94,13 @@ workflows:
 
   build_test_deploy:
     jobs:
-        - build
-        - build_next
-        # we only deploy if build && build_next both succeed.
+        - build_node6
+        - build_node8
+        # we only deploy if build_node6 && build_node8 both succeed and if on master.
         - deploy:
             requires:
-                - build
-                - build_next
+                - build_node6
+                - build_node8
             filters:
                 branches:
                     only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,12 +11,9 @@ jobs:
 
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "package.json" }}
+          - v1-dependencies-v6-{{ checksum "package.json" }}
+          - v1-dependencies-v6-
           - v1-dependencies-
-      
-      - run:
-          name: configure NPM registry
-          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
 
       - run:
           name: install dependencies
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -26,18 +26,83 @@ jobs:
...
+      - checkout
+
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package.json" }}
+          - v1-dependencies-
+
+      - run:
+          name: configure NPM registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
```
vladgolubev: Hm, if you don't publish in this step, then no need for `NPM_TOKEN` in `build_next ` job?

.circleci/config.yml
```diff
@@ -26,18 +26,83 @@ jobs:
...
+
+  deploy:
+    docker:
+      - image: circleci/node:6.10
+
+    steps:
+        - attach_workspace:
+            at: ~/
+
+        -  run:
```
vladgolubev: So I guess `NPM_TOKEN ` should be set here, not in `build` and `build_next` ?

.circleci/config.yml
```diff
@@ -26,18 +26,83 @@ jobs:
...
+    # Yaml doesn't allow you to concatenate lists: https://github.com/yaml/yaml/issues/35
+    # Not in these steps:
+    #  - uploading code coverage
+    #  - saving workspace for the deploy job
+    steps:
+      - checkout
+
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package.json" }}
```
vladgolubev: What about making a cache key different for different node version? Like `v1-dependencies-node8-{{ checksum "package.json" }}` ?

ossareh: does it matter? I guess I don't know enough about node here - does the set of packages differ between v6 and v8?

vladgolubev: Set of packages no, but there is `npm install` , and node 6 and 8 have different versions of  `npm`  bundled which has different logic of installing dependencies, so file structure will differ so cache won't be very helpful if dependencies need to be reinstalled anyway

ossareh: Thanks for that info! I'll go ahead and make that change.

ossareh: I've made the change, I wasn't sure what to default to - in this case it'll default to `v1-dependencies-`; but I think that doesn't make sense since if the checksum process fails you'll still end up with a cache key of `v1-dependencies-vX-`. However, looking at the [caching docs on circleci](https://circleci.com/docs/2.0/caching/#full-example-of-saving-and-restoring-cache) the pattern is what I've implemented.. thoughts?

vladgolubev: Cool! Thanks for taking this into account! Defaulting to `v1-dependencies` as a last resort can remain as is, however, I doubt it will be used more than 1 time at all (prob only the very first run after merging this PR) so LGTM

ossareh: Yeah, actually it helped speed up one of my builds right after asking the question :)

lmammino:
This sounds great :)

.circleci/config.yml
```diff
@@ -25,19 +22,85 @@ jobs:
...
+      - image: circleci/node:6.10
+
+    steps:
+        - attach_workspace:
+            at: ~/
+
+        - run:
+            name: configure NPM registry
+            command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+
```
lmammino: I am not really sure about this. We use the authentication only to publish the package (so it will be useful only in the workflow part). I'll have to see if this needs to be moved to the workflow step just before deployment.

lmammino:
Merged, thanks a million @ossareh and @vladgolubev for your work on this!

This is super cool, let's see what happens shortly!  

screen shot 2018-06-06 at 09 28 52

lmammino:
We got stuck at publishing the git tag with the following prompt:

```
The authenticity of host 'github.com (192.30.253.112)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)? 
```

I'll troubleshoot this one shortly

lmammino:
For now fixed the prompt manually (circle ssh) let's see if this happens again!

ossareh:
Thanks for the follow up work @lmammino 👍

lmammino:
Thanks to you @ossareh, for investigating this option and proposing a PR :)

Issue #198, opened 2018-06-05
Using an async function with middy without callback throws error.


mellisdesigns:
**Summary**
Using async / await within the lambda handler function throws an error when not passing a callback into the `midi` constructor function. AWS documentation specifies that we don't need to pass a callback when using async / await

https://aws.amazon.com/blogs/compute/node-js-8-10-runtime-now-available-in-aws-lambda/

**Example Code**
*index.js*
``` js
const middy = require('middy')

const handler = middy(async (event, context) => {
  return {foo: 'bar'}
})

handler('asd', 'asd')
```

Running this throws an error:
```
yarn start
yarn run v1.7.0
$ node index.js
(node:25665) UnhandledPromiseRejectionWarning: TypeError: callback is not a function
    at terminate (/Users/mellisdesigns/Documents/Boilerplates/lambda-node-boilerplate/node_modules/middy/src/middy.js:151:16)
    at runNext (/Users/mellisdesigns/Documents/Boilerplates/lambda-node-boilerplate/node_modules/middy/src/middy.js:125:14)
    at runErrorMiddlewares (/Users/mellisdesigns/Documents/Boilerplates/lambda-node-boilerplate/node_modules/middy/src/middy.js:129:3)
    at errorHandler (/Users/mellisdesigns/Documents/Boilerplates/lambda-node-boilerplate/node_modules/middy/src/middy.js:159:14)
    at onHandlerError (/Users/mellisdesigns/Documents/Boilerplates/lambda-node-boilerplate/node_modules/middy/src/middy.js:167:16)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:697:11)
    at startup (internal/bootstrap/node.js:201:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:516:3)
(node:25665) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:25665) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
✨  Done in 0.15s.
```

Runtime Version
```
System:
    OS: macOS High Sierra 10.13.4
    CPU: x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
[...]

PR #199, opened 2018-06-05
bugfix - terminate function returns callback or response

mellisdesigns:
**Summary**
`terminate` function should be able to return the value or the callback to fully support async function flow.

This is my initial approach but very happy to discuss alternatives :+1:

**Issue**
#198

### DIFF
--- a/src/middy.js
+++ b/src/middy.js
@@ -148,10 +148,10 @@ const middy = (handler) => {
 
     const terminate = (err) => {
       if (err) {
-        return callback(err)
+        return callback ? callback(err) : err
       }
 
-      return callback(null, instance.response)
+      return callback ? callback(null, instance.response) : instance.response
     }
 
     const errorHandler = err => {


lmammino:
Hello @mellisdesigns! 👋 
At a first glance this makes a lot of sense. What I would love to see before merging this is a dedicated test (that would fail without your changes). Would you be happy to add it to this PR?

mellisdesigns:
Awesome @lmammino. Sounds like a plan! I will build out the test and update the pr. Thanks for your feedback! 👏

lmammino:
Looking forward to it and, most importantly, thanks to you for taking the time to investigate this and submitting a PR 😉

lmammino:
Hello @mellisdesigns,
I was spending some time to figure out which case you are trying to cover, but I couldn't.

We currently have a comprehensive test suite regarding async/await. Check it out here https://github.com/middyjs/middy/blob/master/src/__tests__/middy.js#L420-L510

mellisdesigns:
Hey @lmammino 

I think it relates to the idea that the middy handler should not have to pass in a callback.

https://github.com/middyjs/middy/blob/master/src/__tests__/middy.js#L420

If you check out the test above, although the function we pass in is async, when calling the handler, we still need to pass in a handler to middy `middy({}, {}, (err, response) => {})`

Ideally, it would be nice to do something like:

```
const handler = middy(async (event, context) => {
  return {some: 'response'}
})

const result = await handler({}, {})
console.log(result)
```

In terms of lambda execution, we are able to return promises directly instead of callbacks, because the middy handler is the first point of execution, returning like above would be ace!

mellisdesigns:
I need to create a few tests on my pr, which I am in the process off. You can see the failure happen with the following code.

```
const process = async (event, context) => Promise.resolve('hello world')

const handler = middy(process)

const result = await handler({}, {}) // will fail
console.log(result) // > hello world

```

lmammino:
Hello @mellisdesigns,
I was fiddling a bit with this PR tonight and I tried to add this test based on your last comment:

```javascript
  test('It should be possible to await a middyfied handler', async (endTest) => {
    const originalHandler = async (event, context) => Promise.resolve({some: 'response'})
    const handler = middy(originalHandler)

    const response = await handler({}, {})
    console.log({response}) // undefined
    expect(response).toEqual({some: 'response'})
    endTest()
  })
```

Unfortunately this fails, so I am not sure we are covering the case you described. Am I doing something wrong here?

mellisdesigns:
@lmammino, awesome, that is the exact case I am trying to cover, but really good point, my PR does not resolve this issue, I am trying to figure out where the response gets lost `{some: 'response'}` to `undefined`.

Just a note that I did try this test on master and it failed as well. `callback is undefined`. Will keep trying to figure out where the response gets lost. 👍

pocesar:
since the presence of `callback` (!= undefined) doesn't always mean it's a callback that returns something, it should try to check if the handler uses the async aproach (by using `event, context`) or by using callback approach (have the signature as `event, context, callback`). to know this before hand, you can use `callback.length` if `typeof callback === 'function'`. length return the number of parameters the function has, so if it's 3 === callback, else async (and should use the return type)

shroomist:
It has been a while, do we arrive at conclusion on the best way to approach this?
@rehangit @lmammino

lmammino:
@shroomist, this is still missing a test. I'd love to have that before proceeding with the merge. Would you like to contribute to the PR?

shroomist:
@lmammino  We are starting to rely heavily on `middy`, and having our lambdas `async` way.
It would be nice to be able to test them without callbacks.
I might look into getting this a test, although now mostly interested in `1.0.0-alpha` branch to getting `async` handlers to work.

lmammino:
I am happy to close this one if we have an equivalent/tested solution for 1.0.0.
I hope we'll get rid of 0.x in 1 or 2 months.

shroomist:
I made #314 solving this @lmammino

abdala:
I think maybe this should be closed, right? @lmammino @shroomist.

lmammino:
Yes, thanks!

PR #200, opened 2018-06-05
Fix IMiddy and ICorsOptions type definitions

ossareh:
As per #182 this fixes the types issues on IMiddy and ICorsOptions. `before`, `after`, and `onError` on IMiddy were incorrectly configured:

 - they did not take a function
 - they did not return themselves

While `ICorsOptions` itself was optional when `cors()` was being called, the individual properties of it were not.

The project already had the infrastructure to run tests but was missing a few things:

 - `@types/node` were not present, so `typings-tester` would fail
 - it was easy to miss this because `test:typings` was not part of the build workflow; I've now added it to `npm test`.

I have examples of the tests failing:

 - [tests against incorrect IMiddy](https://circleci.com/gh/ossareh/middy/37)
 - [tests against incorrect ICorsOptions](https://circleci.com/gh/ossareh/middy/36)

These have slightly differing package.json setups which allow them to target the specific tests:

 - branch: [before_after_onerror_failure](https://github.com/middyjs/middy/compare/master...ossareh:before_after_onerror_failure)
 - branch [cors_opts_failure](https://github.com/middyjs/middy/compare/master...ossareh:cors_opts_failure)

I chose to use the name: `foo.types.ts` to indicate that the these files are verifying types. Additionally I chose to locate them in `__tests__` because that seemed like the most correct place. Finally, I also chose to use the jest notation (and thus include `@types/jest`) as that allows folks to just copy and paste their tests into `foo.types.ts` and they'll be picked up by `typings-tester`.

I believe we have the minimum here to constitute a basis to ask folks to contribute accurate types to the project now. However I decided not to implement tests for all of the types because that would have made the commit much larger. I'm happy to contribute a few types each weekend though :)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,9 +1,23 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "lib": ["es2015"],
-    "target": "es2015"
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "lib": [
+      "es2018",
+    ],
+    "moduleResolution": "node",
+    "noEmitHelpers": true,
+    "noEmitOnError": true,
+    "noImplicitReturns": true,
+    "target": "es5",
+    "sourceMap": true,
+    "strict": true,
   },
+  "exclude": [
+    "node_modules",
+  ],
   "files": [
     "index.d.ts",
     "middlewares.d.ts"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
     "noEmitHelpers": true,
     "noEmitOnError": true,
     "noImplicitReturns": true,
-    "target": "es5",
+    "target": "es2015",
     "sourceMap": true,
     "strict": true,
   },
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -1,9 +1,23 @@
...
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "lib": [
+      "es2018",
+    ],
+    "moduleResolution": "node",
+    "noEmitHelpers": true,
+    "noEmitOnError": true,
+    "noImplicitReturns": true,
+    "target": "es5",
```
nnyegaard: Is there a reason to change the target from es2015 to es5?

nnyegaard:
Expect for the one question everything looks good. This is a good addition!

tsconfig.json
ossareh: oh, good catch. This was not meant to change.

Context: This branch started off as two features, this feature but then also another one which made the tests run on node v8.x - during that development I had an issue where the tests were not running on v6, I changed the value, it didn't make a difference, and I forgot to change it base.

The issue was that I'd switched the entire project over to using ts-jest at some point and it complained about some of the javascript within the project, but only on v6 - so I figured something about targeting or babel was wrong, and I was iterating through that.

I'll go ahead and change it back to es2015.

lmammino:
Thanks for the review @nnyegaard! I'll have this version bumped and merged shortly

PR #202, opened 2018-06-06
Check for case-insensitive content-type property

Quelu:
The headers field name are case-insensitive as mentioned in [Section 4.2, "Message Headers"](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

### DIFF
--- a/src/middlewares/jsonBodyParser.js
+++ b/src/middlewares/jsonBodyParser.js
@@ -3,13 +3,16 @@ const contentType = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
-    if (handler.event.headers && handler.event.headers['Content-Type']) {
-      const { type } = contentType.parse(handler.event.headers['Content-Type'])
-      if (type === 'application/json') {
-        try {
-          handler.event.body = JSON.parse(handler.event.body)
-        } catch (err) {
-          throw new createError.UnprocessableEntity('Content type defined as JSON but an invalid JSON was provided')
+    if (handler.event.headers) {
+      const contentTypeHeader = handler.event.headers['Content-Type'] || handler.event.headers['content-type']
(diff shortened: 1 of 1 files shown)


ossareh:
Just a thought on this: since there is no standard on header capitalization, and since there are effectively infinite headers, why not auto inject the header normalizer if you use any middleware? Or, if that is too aggressive, perhaps allow middlewares to declare that they read headers, and then automatically inject the header normalizer?

Quelu:
What a dumbass I am! Didn't see you already have a httpHeaderNormalizer. I totally agree on this, 

But I also found [@middy/http-header-normalizer](https://www.npmjs.com/package/@middy/http-header-normalizer) that adds the `canonical` parameter

Thank you for the help ;)

ossareh:
@Quelu just a heads up - 1.0.0-alpha is new stuff that the middy team have been working on. I'm not sure how compatible it is with the 0.x packages that the master branch points at.

lmammino:
Thanks for opening this issue @Quelu!

Header normalization has been one of the main conversations around middy. In version 0.x we took the unconventional approach to have middlewares only check for canonical header names and suggest to use the header normalizer middleware to be able to support all the possible case variations of that.

This is a bit counter-intuitive from what almost all the other frameworks (and node.js itself) are doing, so many people had issues with that. For this reason, version 1.0 is going to change this into the more intuitive approach of having middlewares check for either canonical and lowercased headers names. The header normalizer middleware will lowercase all the headers and expose the original headers in a map called `rawHeaders`.

I hope this is clear and helpful, depending on which version of middy you are willing to use in the near future.

I am going to close this issue, but feel free to re-open it if you feel it still deserves some clarification.

PR #205, opened 2018-06-12
Add input output logger middleware

### LINKED ISSUE #174: InputOutputLogger middleware
We should have a middleware that, once attached, logs the input (event) and the output (response) once attached.

By default, it should use `console.log`, but the logging function should be customisable by an option.

We might also consider merging this with the Error Logger discussed in #170

kevinrambaud:
As discussed in #174 this is an basic implementation of input-output middleware. Some of you mentioned the possibility also to log errors, but considering the error middleware flow, I wasn't sure that it was something doable.

Closes #174

### DIFF
--- a/packages/input-output-logger/__tests__/index.js
+++ b/packages/input-output-logger/__tests__/index.js
@@ -1,0 +1,21 @@
+const middy = require('../../core')
+const inputOutputLogger = require('../')
+
+describe('📦 Middleware Input Output Logger', () => {
+  test('It should log event and response', () => {
+    const logger = jest.fn()
+
+    const handler = middy((event, context, cb) => {
+      cb(null, { message: 'hello world' })
+    })
+
+    handler
+      .use(inputOutputLogger({ logger }))
+
+    // run the handler
+    handler({ foo: 'bar', fuu: 'baz' }, {}, (_, response) => {
+      expect(logger).toHaveBeenCalledWith({ foo: 'bar', fuu: 'baz' })
+      expect(logger).toHaveBeenCalledWith({ message: 'hello world' })
+    })
+  })
+})
--- a/packages/input-output-logger/__tests__/index.js
+++ b/packages/input-output-logger/__tests__/index.js
@@ -14,8 +14,8 @@ describe('📦 Middleware Input Output Logger', () => {
 
     // run the handler
     handler({ foo: 'bar', fuu: 'baz' }, {}, (_, response) => {
-      expect(logger).toHaveBeenCalledWith({ foo: 'bar', fuu: 'baz' })
-      expect(logger).toHaveBeenCalledWith({ message: 'hello world' })
+      expect(logger).toHaveBeenCalledWith({ event: {foo: 'bar', fuu: 'baz'} })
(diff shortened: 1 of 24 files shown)


packages/input-output-logger/__tests__/index.js
```diff
@@ -0,0 +1,21 @@
...
+    const handler = middy((event, context, cb) => {
+      cb(null, { message: 'hello world' })
+    })
+
+    handler
+      .use(inputOutputLogger({ logger }))
+
+    // run the handler
+    handler({ foo: 'bar', fuu: 'baz' }, {}, (_, response) => {
+      expect(logger).toHaveBeenCalledWith({ foo: 'bar', fuu: 'baz' })
```
lmammino: I love the `fuu` 😛

packages/input-output-logger/index.js
```diff
@@ -0,0 +1,24 @@
...
+  const defaults = {
+    logger: console.log
+  }
+
+  const options = Object.assign({}, defaults, opts)
+
+  return ({
+    before: (handler, next) => {
+      if (typeof options.logger === 'function') {
+        options.logger(handler.event)
```
lmammino: I wonder if we should make clear that this log line concerns the event, by logging something like:

```javascript
options.logger('event', handler.event)
```

or 

```javascript
options.logger({ event: handler.event })
```

What do you guys think?

CC: @dkatavic @vladgolubev @nnyegaard @ossareh  ?

packages/input-output-logger/index.js
```diff
@@ -0,0 +1,24 @@
...
+      }
+
+      return next()
+    },
+    after: (handler, next) => {
+      if (typeof options.logger === 'function') {
+        options.logger(handler.response)
```
lmammino: Same as previous comment, maybe we should make clear that the log line concerns the response.

packages/input-output-logger/index.js
```diff
@@ -0,0 +1,24 @@
+module.exports = (opts) => {
+  const defaults = {
+    logger: console.log
```
lmammino: Should we *JSONStringify* by default?

Something like:

```javascript
const defaults = {
  logger: (data) => console.log(JSON.stringify(data, null, 2))
}
```

What do you think?

lmammino: The goal here is that the output would be more readable

lmammino:
@kevinrambaud, I merged with current `1.0.0-alpha`, so remember to pull in case you want to add any change :)

packages/input-output-logger/__tests__/index.js
kevinrambaud: Ahah

packages/input-output-logger/index.js
kevinrambaud: I'll check the output in Cloudwatch but it seems to be a good idea.

packages/input-output-logger/index.js
ossareh: I prefer option 1, but really with type destructuring in function arguments being so normal now I don't think it really matters either way?

PR #206, opened 2018-06-13
Adds GitHub PR and Issue templates

### LINKED ISSUE #127: httpErrorHandler not working correctly
Hey @middyjs. Love the project. I'll keep an eye on tickets and chip in where I can help.

It appears to me that `httpErrorHandler` isn't working correctly. Serverless is returning a 200 when I throw an Error. Consider the code in the middleware:

https://github.com/middyjs/middy/blob/master/src/middlewares/httpErrorHandler.js#L5

And this test code:

    import HttpError from 'http-errors';
    import createError from 'http-errors';

    console.log( new createError.UnprocessableEntity() instanceof HttpError );
    // false
    console.log( new createError.UnprocessableEntity() instanceof createError.HttpError );
    // true

Which to me suggests a bug.. But I may just be misusing the library. (Warning: I am not a node guy).

Cheers!

lmammino:
Closes #127 

This is a first stab at it. Let me know what you think

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,32 @@
+(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)
+
+What are the steps to reproduce this issue?
+-------------------------------------------
+1. …
+2. …
+3. …
+
+What happens?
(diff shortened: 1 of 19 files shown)


kevinrambaud:
It looks good to me, it contains all the information we might need.

PR #209, opened 2018-06-25
fix: ICorsOptions now has origins option

### LINKED ISSUE #204: [Typescript]: Cors tests are written poorly
Can someone assign this to me please?

Note: The tests here are written poorly. For each option that is added to ICorsOptions one must update each test and add a new test. In fact, we can just do two tests:

 1. where cors is called without an argument
 2. where cors is called with a `{}`

The second case would catch all cases where the properties of ICorsOptions were not defined as optional.

ossareh:
Fixes #204 as well as defining ICorsOptions.origins which was introduced with #194 but not included in #200 for some reason; I must have branched before that landed and didn't update.

I've verified the test fix, if I run the tests without the changes to middlewares.d.ts I get:

```
$npm run test:typings

> middy@0.15.6 test:typings /Users/ossareh/dev/src/github.com/ossareh/middy
> typings-tester --config tsconfig.json index.d.ts middlewares.d.ts src/__tests__/*.types.ts src/**/__tests__/*types.ts

Semantic: src/middlewares/__tests__/cors.types.ts (21, 9) Argument of type '{ origins: string[]; }' is not assignable to parameter of type 'ICorsOptions | undefined'.
  Object literal may only specify known properties, and 'origins' does not exist in type 'ICorsOptions | undefined'.
```

### DIFF
--- a/middlewares.d.ts
+++ b/middlewares.d.ts
@@ -5,6 +5,7 @@ import middy from './'
 
 interface ICorsOptions {
   origin?: string;
+  origins?: string[];
   headers?: string;
   credentials?: boolean;
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.15.6",
+  "version": "0.15.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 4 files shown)


lmammino:
Thanks :)

Issue #175, opened 2018-05-11
Make `next` and `callback` callable only once


lmammino:
This is a common "safety" feature in many middleware frameworks. It makes the `next` (and in your case the lambda `callback`) functions callable only once.

This allows avoiding any side effect that might happen when invoking the function more than once by mistake.

I think implementing this in middy might be very beneficial and that it might increase the value proposition of the framework by also adding the safety as a value proposition to the user.

Implementing this is not complex, but I'd like to collect some opinions on whether you think this is going to be useful or not.

@vladgolubev @dkatavic @padraigobrien

dkatavic:
Great point

For the next one, I definitely agree. My preferred scenario when calling same next function 2 times from the middleware would be
* First call to `next` is gonna activate following middlewares/handler
* Second/subsequent calls should not activate following middlewares/handler. I still want to be warned of this behavior, so logging an error/warning should be a must


From the callback point of view, calling callback multiple times is an AWS Lambda specific behavior. Currently, AWS Lambda just ignores any subsequent call without logging. As a middy user, I would like to be notified when the callback is called multiple times, but I wouldn't like to stop these calls, I would let AWS to decide how to handle that issue.

Does this make sense?

lmammino:
Makes sense and good to know you appreciate the idea :)

I'll add this to the goals for version 1.0.0

lmammino:
This is currently work in progress in [this branch](https://github.com/middyjs/middy/tree/feature/next-callback-callable-once), assigning it to myself.

lmammino:
Closed by 8526ed1e6739c9fe090f8cfbd1c1962165233297

PR #210, opened 2018-06-29
Fix a bug where handler callback is called more than once

mbfisher:
Thanks for your effort on this package, it's helped me refactor a lot of async setup code into middlewares 👍.

By passing a callback to the handler *and* waiting for a returned
promise to resolve, onHandlerSuccess and onHandlerError can be called
twice. Wrapping them with `once` fixes this.

PS Installing `once` has rewritten a bunch of `package-lock.json`, which I'm guessing is because NPM is using a new format in more recent versions; I'm using 6.1.0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ docs/
 !docs/middlewares.md
 
 .idea/
+.vscode/
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       "integrity": "sha1-lr3GtD4TSCASumaRsQGEktOWIsw=",
       "dev": true,
       "requires": {
-        "chalk": "2.4.1",
-        "esutils": "2.0.2",
-        "js-tokens": "3.0.2"
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^3.0.0"
       },
       "dependencies": {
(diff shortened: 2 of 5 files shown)


lmammino:
Hello @mbfisher and thanks for this PR :)

This is something we actually wanted to have as part of the work we are doing for v.1.0.0 (#175).

One thing we need before to be able to proceed with a merge is some unit tests. Ideally, we should have a test that fails with the current code on master (the bugged one) and should instead work on this branch. 

It might be also good to test that 2 separate invocations of the handler will still behave properly firing the right function only once.

Is this something you are willing to work on?

mbfisher:
Sure! I had a quick look at tests when I made the change but I was at work and didn't have much time. I'm on holiday this week to I'll have a look this morning. 👍

mbfisher:
I've pushed a test commit. I can confirm that the tests fail if I remove the `once` wrappers. Let's see what CI makes of it!

I don't normally like using timers in tests, but it seems the best way to replicate the actual execution environment. Let me know if you have any suggestions!

lmammino:
Hey @mbfisher, sorry for the late reply. I have been travelling a lot lately and only now I finally have some time to spend on middy :)

This looks very good. I'll version bump this PR and get it merged. I'll also try to port these changes to the 1.0.0 branch!

PR #224, opened 2018-08-21
bugfix - ssm middleware - setToContext doesn't work when cache option is enabled

theburningmonk:
**Summary**
In the `ssm` middleware, when the `cache` option is enabled we skip the call to SSM but it also skips the step to set the params onto the context.

This wasn't caught in the tests as the tests didn't verify that the properties are set on the invocation context.

**Changes**

- fixed some ssm plugin tests which weren't validating that values are set in the context
- changed places where mockReset is used where it  should be mockClear (since we just want to reset the invocation count)
- cache the SSM response, and apply the cached params even when skip calling SSM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.16.0",
+  "version": "0.16.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


lmammino:
Awesome @theburningmonk :)

I am going to version bump and get this merges. We should also port these changes to v1.0.0, would you have some time to do a PR for that as well?

theburningmonk:
@lmammino sure, let me do a PR to the 1.0.0 branch as well

PR #232, opened 2018-09-04
 Add secretsManager middleware TypeScript definition + entry in Available middlewares docs section

hsz:


theburningmonk:
@hsz thanks for adding these, total oversight on my part! Not sure which is gonna get merged in first, but could you add the new config option that was added in #233 if it gets added in first?

hsz:
@theburningmonk I have accidently removed my fork. So let's keep it as it is or I'll create a new PR. Your call.

lmammino:
Thanks @hsz, let's get #233 done and make sure we have updated typings before merging this. I guess, since #233 is close to be merged, it should be faster/easier to go down this path.

lmammino:
#233 was merged.
I see now the origin repo for this PR does not exist anymore. @hsz, can you please open a new PR with these changes and maybe include also the new option added in #233?

PR #233, opened 2018-09-18
Fail lambda if shouldFail flag provided

sdomagala:
Hello,
It seemed to me that checking if each secret is available at start of lambda function is a bit too much if you are in situation where missing secrets mean there is no point for lambda to run.

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -487,7 +487,7 @@ For each secret, you also provide the name under which its value should be added
 - `secrets` (object) : Map of secrets to fetch from Secrets Manager, where the key is the destination, and value is secret name in Secrets Manager.
   Example: `{secrets: {RDS_LOGIN: 'dev/rds_login'}}`
 - `awsSdkOptions` (object) (optional): Options to pass to AWS.SecretsManager class constructor.
-- `throwOnFailedCall` (boolean) (optional): Defaults to `false`. Set it to `true` if you want your lambda to fail in case call to AWS Secrets Manager fails (secrets don't exist or internal error)
+- `throwOnFailedCall` (boolean) (optional): Defaults to `false`. Set it to `true` if you want your lambda to fail in case call to AWS Secrets Manager fails (secrets don't exist or internal error). It will only print error if secrets are already cached.
 
 NOTES:
 * Lambda is required to have IAM permission for `secretsmanager:GetSecretValue` action
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -487,6 +487,7 @@ For each secret, you also provide the name under which its value should be added
 - `secrets` (object) : Map of secrets to fetch from Secrets Manager, where the key is the destination, and value is secret name in Secrets Manager.
   Example: `{secrets: {RDS_LOGIN: 'dev/rds_login'}}`
 - `awsSdkOptions` (object) (optional): Options to pass to AWS.SecretsManager class constructor.
+- `shouldFail` (boolean) (optional): Defaults to `false`. Set it to `true` if you want your lambda to fail in case call to AWS Secrets Manager fails (secrets don't exist or internal error)
 
 NOTES:
 * Lambda is required to have IAM permission for `secretsmanager:GetSecretValue` action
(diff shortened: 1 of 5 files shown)


src/middlewares/__tests__/secretsManager.js
```diff
@@ -52,6 +52,7 @@ describe('🔒 SecretsManager Middleware', () => {
       promise = promise.then(() => {
         return new Promise((resolve, reject) => {
           handler(event, context, (error, response) => {
+            if (error) reject(error)
```
vladgolubev: Should there be `return` before this `reject`?

sdomagala: Yes, I must've deleted it when cleaning up.
Thanks!

docs/middlewares.md
```diff
@@ -487,6 +487,7 @@ For each secret, you also provide the name under which its value should be added
 - `secrets` (object) : Map of secrets to fetch from Secrets Manager, where the key is the destination, and value is secret name in Secrets Manager.
   Example: `{secrets: {RDS_LOGIN: 'dev/rds_login'}}`
 - `awsSdkOptions` (object) (optional): Options to pass to AWS.SecretsManager class constructor.
+- `shouldFail` (boolean) (optional): Defaults to `false`. Set it to `true` if you want your lambda to fail in case call to AWS Secrets Manager fails (secrets don't exist or internal error)
```
vladgolubev: Maybe name it more explicitly to better convey the intent?

Like `throwOnMissingParam` or `ignoreMissingParams`

src/middlewares/__tests__/secretsManager.js
```diff
@@ -217,6 +218,56 @@ describe('🔒 SecretsManager Middleware', () => {
...
+    })
+    return testScenario({
+      mockResponse: {},
+      middlewareOptions: {
+        secrets: {
+          KEY_NAME: 'failed_call'
+        }
+      },
+      callbacks: [
+        (_, {context}) => {
```
vladgolubev: Unused `context` variable here?

sdomagala: eslint didn't mention it so I thought that's not an issue but sure, I'll delete it

docs/middlewares.md
sdomagala: Yeah I couldn't think of any good name, your proposal doesn't cover errors other than missing param in store - how about `throwOnFailedCall`?

vladgolubev: Hmm, agree, doesn't cover throttling errors etc, so another thing I can think of is to borrow [bail](https://jestjs.io/docs/en/cli#bail) name from jest

sdomagala: Hmm, I think `throw` would be more descriptive here, `bail` is not so common

sdomagala:
Thanks @vladgolubev for quick review!

src/middlewares/secretsManager.js
```diff
@@ -57,6 +58,9 @@ module.exports = opts => {
           // if we already have a cached secrets, then reset the timestamp so we don't
           // keep retrying on every invocation which can cause performance problems
           // when there's temporary problems with Secrets Manager
+          if (options.throwOnFailedCall) {
```
theburningmonk: I think we should only throw on the first request, if you have a cached value already you can continue with execution still. If we want to be smarter then we can build in some mechanism for the app developer to flush the cache - for example, by adding a function to the context object to `flushCachedSecrets`, but let's not do it in this PR.

sdomagala: Hey @theburningmonk, 
I didn't thought of it, it's fixed now, lambda will fail only if first call didn't succeed

theburningmonk:
@sdomagala thanks for making the change, LGTM!

lmammino:
Hey Guys, sorry I have been a bit absent these days. Thanks for taking care of this and doing such a thorough review.

I'll version bump and get this merged.

lmammino:
@sdomagala, thanks for proposing this change. I tried to version bump (to version `0.18.0`) in your branch but it doesn't let me push. can you update `package.json` and `package-lock.json` and change the version to `0.18.0` so that we can get this merged and let the CI pipeline release the new version?

Thanks again :)

sdomagala:
Hey @lmammino, sure, on it!

sdomagala:
@lmammino, could you update package-lock yourself? I'm using yarn exclusively and get a lot of checksum errors when going the npm way, which then result in dependencies refresh.
I'll give you write access to this repo

PR #235, opened 2018-09-19
Add secretsManager middleware TypeScript definition + README fixes

hsz:
Ref discussion: #232

PR #236, opened 2018-09-19
update secrets manager with throwOnFailedCall flag

sdomagala:
1.0.0 doesn't seem to work on node v6.10 because of the issue reported here: https://github.com/jsdom/jsdom/issues/2304, at least for me.
I'd say it's not a regression from this pull request so I parked it for now.
Works like a charm on 8.10.0 though.

When it comes to #235, looks like @theburningmonk took care of it already so I just updated typings

Edit: CI passed for Node 6.10, maybe it's because of cache?


theburningmonk:
LGTM

lmammino:
Great job @sdomagala. Thanks a lot for being so quick at this :)

I'll try to re-run the node.js 6 build without cache to see if we bump into the issue you mentioned, then i'll see to get this version bumped and merged.

PR #239, opened 2018-10-02
Support normalized 'content-type' for jsonBodyParser()

DavidWells:
Just got bit by this because I didn't also have the `.use(httpHeaderNormalizer())` right above `.use(jsonBodyParser())`

This will streamline the amount of middleware needed for JSON based APIs.

```
// this looked fine but was broken due to lowercase 'content-type' in my request.
exports.handler = middy(businessLogic)
  .use(jsonBodyParser())
  .use(validator({ inputSchema: schema.input }))
  .use(httpErrorHandler())
```

```
// using httpHeaderNormalizer before fixed it but it's doing alot more than I need.
exports.handler = middy(businessLogic)
  .use(httpHeaderNormalizer())
  .use(jsonBodyParser())
  .use(validator({ inputSchema: schema.input }))
  .use(httpErrorHandler())
```

I think `jsonBodyParser` should normalize it's own `content-type` header =)

What do you think?

### DIFF
--- a/src/middlewares/jsonBodyParser.js
+++ b/src/middlewares/jsonBodyParser.js
@@ -1,13 +1,15 @@
 const createError = require('http-errors')
-const contentType = require('content-type')
+const contentTypeLib = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
     const { headers } = handler.event
-    // normalize header
-    const headerContentType = headers['Content-Type'] || headers['content-type']
-    if (headers && headerContentType) {
-      const { type } = contentType.parse(headerContentType)
+    if (!headers) {
+      return next()
+    }
+    const contentType = headers['Content-Type'] || headers['content-type']
+    if (contentType) {
+      const { type } = contentTypeLib.parse(contentType)
       if (type === 'application/json') {
         try {
           handler.event.body = JSON.parse(handler.event.body)
--- a/src/middlewares/jsonBodyParser.js
+++ b/src/middlewares/jsonBodyParser.js
@@ -3,8 +3,11 @@ const contentType = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
-    if (handler.event.headers && handler.event.headers['Content-Type']) {
-      const { type } = contentType.parse(handler.event.headers['Content-Type'])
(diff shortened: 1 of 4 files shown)


theburningmonk:
This makes sense to me, these middlewares should be self-contained and avoid implicit inter-dependencies. 👍

theburningmonk:
@DavidWells can you update the corresponding tests which are failing on CI?

lmammino:
Hey @DavidWells, thanks for this PR.
I am not against this, but I would love to have the documentation updated to make clear that if you don't use either one of the 2 supported casing then the middleware is still not going to work, so if you need to support "varied" casing, then it's better to use the header normalizer.

src/middlewares/jsonBodyParser.js
```diff
@@ -3,8 +3,11 @@ const contentType = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
-    if (handler.event.headers && handler.event.headers['Content-Type']) {
-      const { type } = contentType.parse(handler.event.headers['Content-Type'])
+    const { headers } = handler.event
+    // normalize header
+    const headerContentType = headers['Content-Type'] || headers['content-type']
```
lmammino: you should probably check that `headers` is not undefined here as well

src/middlewares/jsonBodyParser.js
```diff
@@ -3,8 +3,11 @@ const contentType = require('content-type')
 
 module.exports = () => ({
   before: (handler, next) => {
-    if (handler.event.headers && handler.event.headers['Content-Type']) {
-      const { type } = contentType.parse(handler.event.headers['Content-Type'])
+    const { headers } = handler.event
```
lmammino: Maybe we can add an early return after this line if `headers` is `undefined` and remove any other check below. This should help us to keep the code simple.

src/middlewares/jsonBodyParser.js
DavidWells: Updated!

src/middlewares/jsonBodyParser.js
DavidWells: updated

PR #246, opened 2018-10-23
Added FunctionShield middleware

0xh0b0:
We would like to contribute FunctionShield middleware - a very easy way to harden AWS Lambda execution environment.
https://www.jeremydaly.com/serverless-security-with-functionshield/
https://www.puresec.io/function-shield

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -239,6 +239,23 @@ module.exports = {
 };
 ```
 
+```
+START RequestId: 3523b3e7-d783-11e8-bf40-6d5f1256964c Version: $LATEST
+[TOKEN VERIFICATION] license is OK
+{"function_shield":true,"policy":"read_write_tmp","details":{"path":"/tmp/test"},"mode":"block"}
+2018-10-24 14:51:58.361 (+03:00)        3523b3e7-d783-11e8-bf40-6d5f1256964c    cannot open file in /tmp folder
+{"function_shield":true,"policy":"create_child_process","details":{"path":"/bin/uname"},"mode":"block"}
+2018-10-24 14:51:58.366 (+03:00)        3523b3e7-d783-11e8-bf40-6d5f1256964c    cannot create child process
+{"function_shield":true,"policy":"outbound_connectivity","details":{"host":"nodejs.org"},"mode":"alert"}
+{"function_shield":true,"policy":"outbound_connectivity","details":{"ip":"104.20.22.46"},"mode":"alert"}
+2018-10-24 14:51:58.416 (+03:00)        3523b3e7-d783-11e8-bf40-6d5f1256964c    successfully created outbound connection
+{"function_shield":true,"policy":"read_handler","details":{"path":"/var/task/handler.js"},"mode":"block"}
+2018-10-24 14:51:58.416 (+03:00)        3523b3e7-d783-11e8-bf40-6d5f1256964c    cannot open handler file
+END RequestId: 3523b3e7-d783-11e8-bf40-6d5f1256964c
+REPORT RequestId: 3523b3e7-d783-11e8-bf40-6d5f1256964c  Duration: 611.64 ms     Billed Duration: 700 ms         Memory Size: 1024 MB    Max Memory Used: 73 MB  
+```
+
+
 ## [httpContentNegotiation](/src/middlewares/httpContentNegotiation.js)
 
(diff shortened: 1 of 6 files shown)


docs/middlewares.md
```diff
@@ -140,6 +141,104 @@ handler(event, context, (_, response) => {
...
+    console.error('cannot create outbound connection');
+  }
+  try {
+    let fd = fs.openSync('/var/task/handler.js', 'r');
+    fs.closeSync(fd);
+    console.log('successfully opened handler file');
+  } catch (e) {
+    console.error('cannot open handler file');
+  }
+}
```
theburningmonk: I think this sample usage is quite confusing, looking at it I find myself asking "what is this sample code trying to show me?". I don't think it's necessary, and I just want to see the initialization bit below. The sample code should be minimal and shows me how to use this middleware, not what function shield does, which there should be explained in the description above, maybe with a link to a more in-depth article or the product landing page.

src/middlewares/functionShield.js
```diff
@@ -0,0 +1,22 @@
...
+        read_write_tmp: 'block',
+        create_child_process: 'block',
+        read_handler: 'block'
+      },
+      disable_analytics: false,
+      token: process.env.FUNCTION_SHIELD_TOKEN || handler.context.FUNCTION_SHIELD_TOKEN
+    }
+    const options = merge(defaults, opts)
+
+    functionShield.configure(options)
```
theburningmonk: this would run `functionShield.configure` on every invocation, how heavy is this call? 

is it worth doing it on every invocation? would it be enough to just do it during coldstart?

src/middlewares/__tests__/functionShield.js
```diff
@@ -0,0 +1,18 @@
+const middy = require('../../middy')
+const functionShield = require('../functionShield')
+
+describe('📦  Middleware FunctionShield', () => {
```
theburningmonk: can you add other tests to show that it blocks HTTP requests, spawning child processes, and accessing /tmp folder as well?

docs/middlewares.md
```diff
@@ -140,6 +141,104 @@ handler(event, context, (_, response) => {
 })
 ```
 
+## [functionShield](/src/middlewares/functionShield.js)
+
+Hardens AWS Lambda execution environment:
+* By monitoring (or blocking) outbound network traffic from your function, you can be certain that your data is never leaked
```
theburningmonk: `outbound network traffic to public internet` to be specific? or mention that traffic to AWS services are not affected, that'd be my first question when I see this sentence

docs/middlewares.md
```diff
@@ -140,6 +141,104 @@ handler(event, context, (_, response) => {
...
+* By disabling read/write operations on the /tmp/ directory, you can make your function truly ephemeral
```
theburningmonk: `you can make your function truly ephemeral` is a bold claim, and untrue (there's all kinds of other things I can do in my code to make my function stateful), and more importantly it's not all that important to me as a developer.

Being ephemeral is means to an end, and in this particular case I'd be more incentivized to use this feature if you explain why using /tmp folder is a bad idea, and why it should be disabled, maybe pointing to an article?

src/middlewares/__tests__/functionShield.js
0xh0b0: It is not easy to do.
FunctionShield adopted to run in AWS Lambda container and requires token.
1. We should use `lambci/lambda` images instead of `circleci/node` images
2. We should put FUNCTION_SHIELD_TOKEN to circleci environment variables
3. We should find a workaround to run tests locally (probably in docker container with FUNCTION_SHIELD_TOKEN env. variable)
What do you think?

vladgolubev: IMO tests for HTTP request blocking and child process are out of the scope of the middleware, it should not repeat what is tested inside function shield package.

theburningmonk: @vladgolubev good call, at that point we're not testing the middleware anymore.

theburningmonk:
@lmammino can you have a look and merge if you're happy?

theburningmonk:
@0xh0b0 can you do another PR for this feature but against the `1.0.0-alpha` branch?

0xh0b0:
> @0xh0b0 can you do another PR for this feature but against the 1.0.0-alpha branch?
Sure, I'll do PR next week

lmammino:
Awesome. I'll try to get this merged sometime this evening.

Thank you so much for this addition to middy

PR #259, opened 2018-12-04
Updates S3 event version to allow event versions of 2.x

kenleytomlin:
## Description

The s3KeyNormalizer middleware was not running because the version check was failing. AWS have updated S3 events and are now using a new version of 2.1. This PR updates the version number to run the normalizer when the version is 2.1. [AWS docs for S3 events](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.20.2",
+  "version": "0.20.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.20.2",
+  "version": "0.20.3",
   "description": "🛵 The stylish Node.js middleware engine for AWS Lambda",
   "main": "./index.js",
   "files": [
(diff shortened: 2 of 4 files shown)


vladgolubev:
@kenleytomlin  do you know did they switch 100% customers to the version 2.1? Is this the version in all regions? Does it makes sense to support version 2.0 in the middleware for some time? Or even better check is the version is `2.x` to avoid sudden failures in the future?

kenleytomlin:
@vladgolubev `The following example shows version 2.0 of the event message structure, which is no longer used by Amazon S3.` I do not know 100% if there are any regions which are still using 2.0. I would expect this to be highlighted in the documentation.

I also question if the check for the version is correct because this broke our code and prevented the key normalizer from running.

vladgolubev:
@kenleytomlin the only thing added in the 2.1 is this property:

```
"glacierEventData": {
	"restoreEventData": {
		"lifecycleRestorationExpiryTime": "The time, in ISO-8601 format, for example, 1970-01-01T00:00:00.000Z, of Restore Expiry",
		"lifecycleRestoreStorageClass": "Source storage class for restore"
	}
}
```

I'd say it's safe to assume AWS follows semver and we should just check is the version starts with `2.`. Would you mind making this addition, Kenley?

kenleytomlin:
@vladgolubev Yep no problem I will add a check for `2.x` for the event version.

lmammino:
Done and also forward-ported the changes to 1.0.0-alpha

lmammino:
Thanks to you @kenleytomlin :)

PR #261, opened 2018-12-04
docs: fix broken link

### LINKED ISSUE #260: Broken link in the website docs
Hi guys,

I think the repo of the website `middy.js.org` is private otherwise I would have sent a PR! 

There's a broken link in the docs:

```
More details on creating middlewares
Check the ------> code for existing middlewares <------- to see more examples on how to write a middleware.
```

screen shot 2018-12-04 at 18 42 11

Currently pointing to:
```
https://middy.js.org/src/middlewares
```

instead of:
```
https://github.com/middyjs/middy/blob/master/docs/middlewares.md
```

Thanks!

andreafalzetti:
First contribution 🎉

This fixes #260.

### DIFF
--- a/README.md.hb
+++ b/README.md.hb
@@ -496,7 +496,7 @@ methods to allow you to quickly hook-in simple inline middlewares.
 
 ### More details on creating middlewares
 
-Check the [code for existing middlewares](/src/middlewares) to see more examples
+Check the [code for existing middlewares](/docs/middlewares.md) to see more examples
 on how to write a middleware.
 
 


README.md.hb
```diff
@@ -496,7 +496,7 @@ methods to allow you to quickly hook-in simple inline middlewares.
 
 ### More details on creating middlewares
 
-Check the [code for existing middlewares](/src/middlewares) to see more examples
+Check the [code for existing middlewares](https://github.com/middyjs/middy/blob/master/docs/middlewares.md) to see more examples
```
ChristianMurphy: `/docs/middlewares` should route to the same place.
While still giving markdown some flexibility on choosing how to construct the url on GithHub vs the docs site.

andreafalzetti: @ChristianMurphy Thanks for your feedback! I am not familiar with GH pages but sounds like a better approach. I will update the PR. Don't I need the `.md` ext at the end like `(/src/middlewares.md)`?

ChristianMurphy: Including the file type should be fine, and appears to be used commonly in the GitHub docs.
https://blog.github.com/2016-12-05-relative-links-for-github-pages
Keep it. :+1:

andreafalzetti: I landed on the same page but it's not clear if GitHub pages will fallback on `.md`. Anyway, I've updated my PR. :)

andreafalzetti:
@vladgolubev I am not familiar with your CI setup but it seems like the deploy is broken - and the change does not appear on the website :/

vladgolubev:
@andreafalzetti yeap, I forgot to bump the version. Will fix it now

PR #265, opened 2018-12-21
Add onChange handler to ssm + audit packages

sdomagala:
Hello,
In this PR I'd like to add functionality to automatically trigger handler when ssm parameter is regenerated - we use this data in other parts of code and need to be able to regenerate if any parameter changed.

While at it, I saw multiple vulnerabilities doing npm audit so I thought it would be sensible to make it a part of pipeline so nothing slips to production anymore (added only to Node8.10 as 6.10 doesn't have that command yet) .

### Edit
Actually, npm audit is not available in Circle CI, had to revert this change

Cheers

### DIFF
--- a/src/middlewares/__tests__/ssm.js
+++ b/src/middlewares/__tests__/ssm.js
@@ -9,12 +9,15 @@ describe('🔒 SSM Middleware', () => {
   SSM.prototype.getParameters = getParametersMock
   const getParametersByPathMock = jest.fn()
   SSM.prototype.getParametersByPath = getParametersByPathMock
+  const onChange = jest.fn()
 
   beforeEach(() => {
     getParametersMock.mockReset()
     getParametersMock.mockClear()
     getParametersByPathMock.mockReset()
     getParametersByPathMock.mockClear()
+    onChange.mockReset()
+    onChange.mockClear()
     delete process.env.KEY_NAME
   })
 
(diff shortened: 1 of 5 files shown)


src/middlewares/__tests__/ssm.js
```diff
@@ -189,6 +217,39 @@ describe('🔒 SSM Middleware', () => {
...
+        },
+        cache: true,
+        onChange: onChange,
+        cacheExpiryInMillis: 10,
+        setToContext: true,
+        paramsLoaded: false
+      },
+      callbacks: [
+        (_, {context}) => {
+          expect(onChange.mock.calls.length).toEqual(1)
```
vladgolubev: Just FYI there is [.toHaveBeenCalledTimes(number)](https://jestjs.io/docs/en/expect#tohavebeencalledtimesnumber) API:)

sdomagala: I knew there must be something like this but couldn't find it - changed :)

lmammino:
I love this feature :)
Thanks @sdomagala!

Once merged would you be willing to port this change to the 1.x branch?

sdomagala:
Thanks for quick response 😉 
Holiday may be a bit busy but if I can squeeze an hour or two then I'll definitely do it!

PR #272, opened 2019-01-16
Add string template option to ssm middleware

aheissenberger:
**use case:**
* retrieve different options from parameter store based on aws api key
* api key is dynamic set in the context of each request

**implementation**
* only activated when needed with new option `stringTemplates : true`
* uses a syntax well known with ES6 Template literals `${VARIABLENAME}`
* full access to all objects in `handler`
* only requests with identical requests to parameter store are cached

**Examples:** 
`{names: {DB_URL: '/dev/${event.body.service}/options'}}` 
`{paths: {DB_: '/${context.stage}/service/db'}}`

**TODO:**
I need help to implement a test which will check if the cache is ignored when the value of a variable changes. I failed to manipulate the context or the event during two sequential calls:
https://github.com/aheissenberger/middy/blob/25ed9586acfd986224280c1d1a4f57a575212213/src/middlewares/__tests__/ssm.js#L449

### DIFF
--- a/src/middlewares/ssm.js
+++ b/src/middlewares/ssm.js
@@ -1,3 +1,4 @@
+const {createHash} = require('crypto')
 let ssmInstance
 
 module.exports = opts => {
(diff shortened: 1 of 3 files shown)


lmammino:
Hello @aheissenberger and thanks for submitting this PR.
Sorry for the late reply but I am currently traveling for work. I'll try to review this PR and see if I can address your test issue (hopefully some good coffee and good connectivity will be on my side :P)

lmammino:
Hello @aheissenberger and thanks for submitting this PR.
Sorry for the late reply but I am currently traveling for work. I'll try to review this PR and see if I can address your test issue (hopefully some good coffee and good connectivity will be on my side :P)

src/middlewares/ssm.js
```diff
@@ -192,3 +209,34 @@ const invertObject = obj =>
...
+ * Convert string to template string and replace variables
+ * @param {String} template Template string with variable placeholder allowed by ES6 Template Literals
+ * @param {Object} Values to be used by the template
+ */
+const StringTemplate = function (template, params) {
+  const names = Object.keys(params)
+  const vals = Object.values(params)
+  // eslint-disable-next-line no-new-func
+  return new Function(...names, `return \`${template}\`;`)(...vals)
+}
```
lmammino: This approach is interesting. First time I am seeing it!

lmammino: Only one note in terms of naming: maybe we should call this something like `resolveStringTemplate`.
This should be more explicit. Also since this is not used as a "class" but as a function I would prefer to keep the first letter lowercase.

theburningmonk:
I like the flexibility that's proposed here, I agree with @lmammino on supporting functions over special syntax. My main concern over this proposal is that if the parameter path have to be decided based on invocation event/context then it forces you to have very permissive IAM policies.

I suppose if that's your use case then you gotta do what you gotta do to make it work, but if you can provide some context around the use case that'd be helpful for me.

lmammino:
@aheissenberger, I just realized that I probably didn't make my point clear.

I am super happy to get this merged if you would like to apply the changes to use functions rather than templates and maybe also drop a note in the middleware docs about trying not to have too open security policies.

Let me know if you want to carry these ones over yourself or if you need support :)

willfarrell:
@aheissenberger Just wanted to follow up on this. Are you still interested in adding this functionality to middy?

willfarrell:
Please re-open if you'd like to re-visit this.

PR #281, opened 2019-01-29
Add stringifyBody middleware

ciacka:
I did simple middleware to stringify response's body.


theburningmonk:
@lmammino I'm in favour of a more general purpose response body serializer too. HTTP has a rich content negotiation mechanism, for the core I'd like something that respects that content negotiation protocol and apply the appropriate serialization.

Issue #275, opened 2019-01-20
[TypeScript] Context type with ssm middleware injecting in context


JuHwon:
I would like to use the ssm middleware to inject parameters into context like the following

```typescript
export const handler = middy(handlerFunction)
    .use(ssm({
        cache: true,
        cacheExpiryInMillis: 3 * 60 * 1000,
        setToContext: true,
        names: {
            user_table: '/nobox/dev/user_table'
        }
    }))
```

When  using typescript i am not allowed to access `context.user_table` since it is not defined in the `Context` type of the aws-lambda types package obviously. 

So i thought to enhance the context interface like that:

```typescript
interface HandlerContext extends Context { user_table: string }
type HandlerFunction = (event: APIGatewayProxyEvent, context: HandlerContext) => Promise<APIGatewayProxyResult>

const handlerFunction: HandlerFunction = async (event, context) => {
    console.log("user_table: ", context.user_table)
    return {
        statusCode: 500,
        body: JSON.stringify({ message: "Not Implemented yet." })
    }
}
```

Though when doing that i do get the following typescript error at `middy(handlerfunction)`
```
Argument of type 'HandlerFunction' is not assignable to parameter of type 'Handler<any, any> | AsyncHandler'.
  Type 'HandlerFunction' is not assignable to type 'Handler<any, any>'.
    Types of parameters 'context' and 'context' are incompatible.
      Property 'user_table' is missing in type 'Context' but required in type 'HandlerContext'.
```

The only fix i came up with is `middy(handlerFunction as unknown as APIGatewayProxyHandler)` which kinda feels wrong ¯\_(ツ)_/¯

Is there any better solution? I guess i am not the only one using the ssm middleware with context and typescript.

lmammino:
Hello @JuHwon, would you like to submit a PR with this fix? :)

pocesar:
the typings should rely on using generics, I tried adding them but they might break some people that are relying on non generic types like

```ts
declare type AsyncHandler<T, R> = ((event: T, context: Context, callback: Callback<R>) => void) | ((event: T, context: Context) => Promise<R>);
declare const middy: <C extends AsyncHandler<T extends infer T ? T : any, R extends infer R ? R : any>, T = APIGatewayProxyEvent, R = APIGatewayProxyResult>(handler: C) => middy.IMiddy<T, R>;
```
defaulting to API gateway (guess that's what most people use with serverless)

pocesar:
[...]

PR #283, opened 2019-02-02
Typescript typesafe, guards and generics

pocesar:
As per #275, it's a breaking change for people using Typescript, but it's as safe as it gets. the usage is 

```ts
import { APIGatewayProxyHandler, Context } from 'aws-lambda'

interface CustomContext extends Context {
  haha?: boolean // needs ? because doesn't exist in original context when TS tries to infer the type
}

const get = middy<APIGatewayProxyHandler>( async (event, context: CustomContext) => {
  return {
    statusCode: 200,
    body: ''
  }
}).use(cors()).after((s, next) => { s.context; next() })
```

since it's a breaking change (for the types only), renamed the interfaces to force people rewrite it

custom data on "event" would then be:

```ts
import { Handler, APIGatewayProxyEvent, Context } from 'aws-lambda'

middy<Handler<APIGatewayProxyEvent & { extra: string, field: string }, {}>>( async (event, context: CustomContext) => {
  event.extra
  event.field
  return {
  }
})
```
this way, the event can append data from middlewares that modify the event object, like `httpContentNegotiation` for example, that adds `preferredEncoding`, etc, to the event

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,41 +1,56 @@
-import { Callback, Context, Handler, ProxyResult } from 'aws-lambda';
+import {
+  Callback,
+  Context,
+  Handler
+} from 'aws-lambda';
 
-type AsyncHandler = (event: any, context: Context) => Promise<ProxyResult | object>;
+declare type EventType<T> =
+  T extends (event: infer EventArgType, context: Context, callback: Callback<any>) => void ? EventArgType :
+  T extends (event: infer EventArgType, context: Context) => Promise<any> ? EventArgType :
+  never;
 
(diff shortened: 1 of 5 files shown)


lmammino:
Thank @pocesar :)

Gonna version bump and get this merged.

pocesar:
sure, is the API the same?

lmammino:
Regarding TypeScript there shouldn't be any difference. Big change is in term of repo structure, it's a *monorepo* where every middleware will be published in its own package. Let me know if you have any trouble in figuring out the new structure.

PR #286, opened 2019-02-04
Removed functionShield from core

### LINKED ISSUE #282: functionShield middleware uses closed-source dependency with custom license/EULA
Hi there 👋 

I've really been enjoying using `middy`, but recently ran into a legal issue with one of it's dependencies.

After running a security and license scan on our project dependencies, we realized that the `functionShield` middleware that comes bundled with `middy` also includes the [`@puresec/function-shield`](https://www.npmjs.com/package/@puresec/function-shield) package. `@puresec/function-shield` is closed-source, and if you look at the license, it points to a [custom license/EULA](https://assets.puresec.io/agreements/FunctionShield-EULA(13Jan19).pdf).

This means that just by installing `middy` (and installing it's dependency `@puresec/function-shield`), you're legally agreeing to the terms in the PureSec license/EULA. Even if you don't use the `functionShield` middleware (as we don't), you're still downloading it and agreeing to the license.

Would it be possible to move `@puresec/function-shield` from the `dependencies` section of `middy`'s `package.json`, and move it into `peerDependencies`? That way npm won't install it automatically when installing `middy`. We can also add a note in the documentation that in order to use the `functionShield` middleware, you'll also need to install `@puresec/function-shield` and suggest to review the license before installing.

I'd be happy to submit a PR with the changes I suggested, please let me know what you think.
Thanks!

lmammino:
Closes #282


theburningmonk:
For breaking changes like this, should we do it as part of v1 instead? 

If not, would the change be compatible when we do roll out v1?

lmammino:
Very good questions @theburningmonk. I am already referring to the middleware in v1 as an alternative installation. that one should be compatible as we didn't change any middleware internal from v0 to v1.

I also created a bunch of documentation notes + a changelog files to make sure that it's clear for the user that this version comes with a breaking change.

PR #288, opened 2019-02-05
Allow use() to receive multiple middlewares

### LINKED ISSUE #277: Support passing more than 1 config at a time into use()
Alongside the current functionality of being able to chain `.use()` I find myself wanting to be able to pass more than 1 config into use at once. The use-case for this is when I have middleware "groups" set up elsewhere, and would rather import that onto handlers, than have to remember everything in the group every time I set up a new handler, for example:
```javascript
const middy = require('middy');
const middlewareGroup = [middleware1(), middleware2(), middleware3()];

exports.handler = middy((event, context, callback) => callback(null, 'foo'))
    .use(...middlewareGroup);
```

alexdebrie:
Closes #277 

This allows users to pass an array of middlewares to `use()`:

```javascript
const middy = require("@middy/core");
const middleware1 = require("sample-middleware1");
const middleware2 = require("sample-middleware2");
const middleware3 = require("sample-middleware3");
const middlewares = [middleware1(), middleware2(), middleware3()]

const originalHandler = (event, context, callback) => {
  /* your business logic */
};

const handler = middy(originalHandler);

handler.use(middlewares)

module.exports = { handler };
```

It should be backwards-compatible. 

All of the logic for the existing `use` function went into a new `applyMiddleware` method. The `use` function now is pretty small -- if the input is an array, it calls `applyMiddleware` with each element in the array. If the input is an object, it calls `applyMiddleware` with the object. Otherwise, it throws an error.

Open to suggestions / changes!


lmammino:
Not sure why tests in node 6 fails. I was even thinking to drop Node 6 in v1.0.0, but I shouldn't do it as part of this PR. I will investigate tomorrow

lmammino:
Finally managed to fix the issue. Some dependency was messed up (which reflected only in node6 where the package-lock is not used). Ready to merge

alexdebrie:
Awesome! Thanks for cleaning up the types and tracking down this issue, @lmammino 🙏 .

lmammino:
@alexdebrie, thanks to you for submitting this awesome PR :)

PR #290, opened 2019-02-07
Add default export to ssm middleware

dbartholomae:
This seems to be an oversight - the other middlewares have it, and ssm not having it breaks TypeScript imports

### DIFF
--- a/packages/secrets-manager/index.d.ts
+++ b/packages/secrets-manager/index.d.ts
@@ -10,3 +10,5 @@ interface ISecretsManagerOptions {
 }
 
 declare function secretsManager(opts?: ISecretsManagerOptions): middy.IMiddyMiddlewareObject;
+
+export default secretsManager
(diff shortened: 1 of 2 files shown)


theburningmonk:
@dbartholomae you're right! it's an oversight, it was my bad. While you're here, do you mind popping a default export into the secrets-manager package as well?

dbartholomae:
Added :)

packages/secrets-manager/index.d.ts
```diff
@@ -1,5 +1,6 @@
 import { SecretsManager } from 'aws-sdk'
 import middy from '../core'
+import ssm from '../ssm'
```
theburningmonk: is this required? the two should be separate, right?

dbartholomae: Thanks for catching that! My IDE snuck that in ^^'
I'll remove it and double check the rest.

dbartholomae: I've removed the line

lmammino:
@theburningmonk are you ok with the latest changes?

dbartholomae:
Accidentially added another commit, will remove it asap

PR #302, opened 2019-02-20
Add types for httpSecurityHeaders

crespowang:
I noticed the types for **httpSecurityHeaders** is added in 1.0.0-alpha but not master, so just do it to make it work for now.


lmammino:
Thanks, seems great at a first glance. I'll review this better and get it merged in the next few hours :)

Issue #169, opened 2018-04-25
HttpResponseSerializer middleware


lmammino:
I believe it would be awesome to have a `HttpResponseSerializer` middleware that would be able to use different serialization mechanisms based on the current content negotiation

e.g. if the user prefers JSON, the response *object* is automatically serialized to JSON, if the user prefers XML, then the response object is automatically converted to XML.

The middleware should be able to handle even error paths and serialize errors accordingly.

noetix:
How would status codes work? Assuming you don't want 200 for every non-error response.

MaxVinogradov:
Hi! I'm interested in this - what do you think about https://www.npmjs.com/package/fast-json-stringify ?  
Fastify framework use this with success.

noetix:
@MaxVinogradov

How you transform data is probably the easier part to this issue.

If you want to use `fast-json-stringify` you could modify the middleware I wrote:

```
export default () => ({
  after: (handler, next) => {
    handler.response = {
      statusCode: 200,
      headers: {
        'Content-Type': 'application/json'
      },
      body: JSON.stringify(handler.response) // change this
    }

    next()
  }
})
```

The bigger issues are the fact that you're always returning 200 and always JSON.

Some customisation should be added for status codes -- those that want 201 on POST success for example.

Data transformation should be pluggable (cause JSON, XML, PDF, CSV, etc, etc) and each would likely need its own configuration. Our use-case here is a PDF transformer that needs to point to a particular template to facilitate transformation.

noetix:
I have some middy work coming up and may have some time to contribute.

Here is what I was thinking for a response serializer. The below is the `@middy/http-content-negotiation` example refactored using the proposed.

```
import middy from '@middy/core'
import httpContentNegotiation from '@middy/http-content-negotiation'
import httpHeaderNormalizer from '@middy/http-header-normalizer'
import httpErrorHandler from '@middy/http-error-handler'
import httpResponseSerializer from '@middy/http-response-serializer'

const handler = middy((event, context, cb) => {
  let body

  switch (event.preferredLanguage) {
    case 'it-it':
      body = 'Ciao Mondo'
      break
    case 'fr-fr':
      body = 'Bonjour le monde'
      break
    default:
      body = 'Hello world'
  }

  return cb(null, {
    statusCode: 200,
    body
  })
})

handler
  .use(httpHeaderNormalizer())
  .use(httpContentNegotiation({
    parseCharsets: false,
    parseEncodings: false,
[...]

PR #304, opened 2019-02-25
Added http-response-serializer middleware

noetix:
Implements what was discussed in https://github.com/middyjs/middy/issues/169

One exception, it does not return `406` when a serializer can't be matched. I feel like `http-content-negotiation` should be blocking this on the incoming instead of failing on the outgoing -- wasting time generating a response.

Tests failing because I'm using `test.each` which requires a version bump on `jest`. You've done this on `master` already, just need to bump `1.0.0-alpha` branch.


lmammino:
Hello @noetix, sorry for the terrible delay,
I'll work on getting this one merged shortly. :)

lmammino:
Finished to fight the CI and other jest issues... version bumping and then it should be ready for merge :)

PR #306, opened 2019-02-27
caseless headers in CORS middleware

NShahri:
- http headers attributes are *case insensitive* and it can be different by browsers. 
- For example Chrome using *origin* and Firefox is using *Origin*. 
- I think we should not force developer to use another middleware before CORS middleware.

CHANGES:
- Using `caseless` package to `get` or `set` the header attributes.
- update unit tests
- if `headers.Origin` is provided and `options.origin` is equal to `*`, middleware returns `headers.Origin` as `origin` attribute. ( options.credentials is not important to the value of `headers.Origin`, please correct me if I am wrong )

Remaining Tasks:
- Document how default options `origin` and `origins` are working and which one has higher priority
- Document when handler returns any CORS attribute, then middleware won't update that attribute
- Document `Access-Control-Allow-Credentials` options

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "ajv": "^6.9.1",
     "ajv-i18n": "^3.4.0",
     "ajv-keywords": "^3.4.0",
+    "caseless": "^0.12.0",
     "content-type": "^1.0.4",
     "http-errors": "^1.7.1",
     "json-mask": "^0.3.8",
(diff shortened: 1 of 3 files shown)


lmammino:
Hello @NShahri, first of all, sorry for the delayed reply (busy times :( ).
Also, thanks for submitting this PR, but I am not fully sold on it.
I see the value on making the CORS middleware simpler to use (doens't require the normalize headers anymore), but this change will make the middleware inconsistent with all the other middlewares using HTTP headers. Also if we go down this path, we will make harder for developers to build their own http-headers-based middlewars.

Unless somebody else from the core team has strong opinions i am unfortunately 👎 on merging this.

CC: @dkatavic @theburningmonk @vladgolubev

PR #312, opened 2019-03-26
Prevent Cors middleware swallowing errors

### LINKED ISSUE #264: CORS handlers with error callbacks
If a handler tries to terminate lambda execution with error status for any reason, like trigering lambda retries, using 'cors' middleware might break it's implementation.

The following test fails. I know this is a definition issue, but I think it breaks lambda specs for error callback scenario:

```
const expect = require('chai').expect;
const middy = require('middy');
const {cors} = require('middy/middlewares');

describe('When a middy handler evokes callback passing an error', () => {
    const failerMiddyCorsHandler = middy((event, context, callback) => {
        callback('An error');
    }).use(cors());

    it('should result lambda error', () => {
        return failerMiddyCorsHandler({httpMethod: 'GET'}, {}, (error, result) => {
            expect(result).to.be.undefined;
            expect(error).to.equal('An error')
        })
    });
});
```

gsingh1:
Fixes #264 

Stumbled upon an issue this morning where we found that by using the `cors` middleware with `onError`, errors are not being passed here: 

https://github.com/middyjs/middy/blob/1d04aefd2a5cc8b16e3820bc0ca1d0654c3d3d92/src/middlewares/cors.js#L50

resulting in success callbacks and response of 200 empty body.

When used with other middlewares such as the `httpErrorHandler`, the non-http errors do not have a chance to be converted to an http response and are passed along and when using in addition with `cors`, the error is swallowed and the callback is called without an error resulting in a 200 response with an empty body.

This change allows any errors to be passed to the next middleware by `cors` if any and gives the user a chance to handle the error or if there is no further middleware in the chain, it will bubble up to the `terminate()` and callback correctly with an error here: 

https://github.com/middyjs/middy/blob/1d04aefd2a5cc8b16e3820bc0ca1d0654c3d3d92/src/middy.js#L152

instead of here:

https://github.com/middyjs/middy/blob/1d04aefd2a5cc8b16e3820bc0ca1d0654c3d3d92/src/middy.js#L155

The only problematic issue I see is the loss of cors headers if other middlewares do not handle the error after `cors`. Thoughts? The tests have also been updated to reflect this.

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -91,9 +91,29 @@ Sets headers in `after` and `onError` phases.
 - `credentials` (bool) (optional): if true, sets the `Access-Control-Allow-Origin` as request header `Origin`, if present (default `false`)
 
 NOTES:
+- If another middleware does not handle and swallow errors, then it will bubble all the way up 
+and terminate the Lambda invocation with an error. In this case API Gateway would return a default 502 response, and the CORS headers would be lost. To prevent this, you should use the `httpErrorHandler` middleware before the `cors` middleware like this:
 
-- When used with `httpErrorHandler` middleware, ensure to use the `httpErrorHandler` middleware before `cors` to avoid 
-the loss of CORS headers 
+```javascript
+const middy = require('middy')
+const { httpErrorHandler, cors } = require('middy/middlewares')
+
+const handler = middy((event, context, cb) => {
+  throw new createError.UnprocessableEntity()
+})
+
+handler.use(httpErrorHandler())
+       .use(cors())
+           
+// when Lambda runs the handler...
+handler({}, {}, (_, response) => {
+  expect(response.headers['Access-Control-Allow-Origin']).toEqual('*')
+  expect(response).toEqual({
+      statusCode: 422,
+      body: 'Unprocessable Entity'
+    })
+})
+```
 
 ### Sample usage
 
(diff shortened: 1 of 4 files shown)


vladgolubev:
> The only problematic issue I see is the loss of cors headers if other middlewares do not handle the error after cors.

I just try to remember to put the http error handler before `cors`. This can be added to the docs if it's not there already

```js
  .use(httpErrorHandler())
  .use(cors());
```

gsingh1:
@vladgolubev I'll add to the docs as part of this PR 👍

docs/middlewares.md
```diff
@@ -90,6 +90,11 @@ Sets headers in `after` and `onError` phases.
 - `headers` (string) (optional): value to put in Access-Control-Allow-Headers (default: `null`)
 - `credentials` (bool) (optional): if true, sets the `Access-Control-Allow-Origin` as request header `Origin`, if present (default `false`)
 
+NOTES:
+
+- When used with `httpErrorHandler` middleware, ensure to use the `httpErrorHandler` middleware before `cors` to avoid 
+the loss of CORS headers
```
theburningmonk: @gsingh1 you need to explain `the loss of CORS headers`, e.g. `if another middleware doesn't handle and swallow the error, then it'll bubble all the way to the top and terminate the Lambda invocation with an error. In this case API Gateway would return a default 502 response, and the CORS headers would be lost. To prevent this, you should use the httpErrorHandler middleware before the cors middleware like this:
<insert code example here>`

gsingh1: Updated explanation and added code sample

lmammino:
Thanks @gsingh1 for submitting this and to @theburningmonk and @vladgolubev for the review!

lmammino:
I suppose this has to be "forward"-ported to v1.0.0, right? @gsingh1 would you like to submit a PR for v.1.0.0 as well?

gsingh1:
No problem @lmammino, I'll submit later today 👍

PR #314, opened 2019-03-29
return promise from instance if no callback is present

shroomist:
Solves #198 
Did not touch that much here, just wrapped `instance()` logic in `new Promise` and return it if no callback is present.

Signed-off-by: Andrej Novikov <jazzaiman@gmail.com>

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-alpha.26"
+  "version": "1.0.0-alpha.27"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy-monorepo",
-  "version": "1.0.0-alpha.26",
+  "version": "1.0.0-alpha.27",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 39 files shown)


lmammino:
Thanks @shroomist, I'll version bump and get this one merged :)

vikasgarghb:
Looks like this do not work when running your function with `sam local invoke`. There is this callback function which is being set:
```
function() {
      if (!finished) {
        toWrap.apply(null, arguments);
        finished = true;
      }
    }
```

So, the condition for resolving the promise is never reached and the function just hangs.

vikasgarghb:
In order to get it working, I had to add this to my function: `context.callbackWaitsForEmptyEventLoop = false;`

lmammino:
Helllo @vikasgarghb, which version of middy are you using? Would you like to open a dedicated issue and maybe a PR to help us to address this problem?

vikasgarghb:
@lmammino I am on 1.0.0-beta. I will open a dedicated issue.

vikasgarghb:
Basically the documentation needs to be updated that you need to use `doNotWaitForEmptyEventLoop` middleware to make async/await handler work with middy.

vikasgarghb:
@lmammino https://github.com/middyjs/middy/issues/505

PR #315, opened 2019-03-29
1.0.0-alpha - Fix relative path that breaks TS compile

duro:
In the Secrets Manager middleware, there is a relative path in the index.d.ts file to middy core. While this probably works in a traditional npm/yarn project, when using yarn workspaces in a mono repo, it cannot be guaranteed that middy core will be installed right along side this plugin.

For example in my project I end up with the following structure:

|- /node_modules
|-- /@middy
|---- /core
|- /services
|-- /service-a
|---- /node_modules
|------ /@middy
|-------- /secrets-manager
|---- package.json
|- package.json

### DIFF
--- a/packages/secrets-manager/index.d.ts
+++ b/packages/secrets-manager/index.d.ts
@@ -1,5 +1,5 @@
 import { SecretsManager } from 'aws-sdk'
-import middy from '../core'
+import middy from '@middy/core'
 
 interface ISecretsManagerOptions {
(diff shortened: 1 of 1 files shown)


lmammino:
Hello @duro, thanks for your pr.
This seems to break the tests. 

Do we have to add middy core to the dependencies? do we need to do something similar for all the other core plugins?

PR #317, opened 2019-04-04
Multipart parser

TeddyHandleman:
**Description**: A middleware to parse incoming `multipart/form-data` requests. Implemented with BusBoy.

**Purpose**: Say a client requires a user to fill out a form. This might contain first name, last name, email, and it also may contain an avatar picture the user can select from their machine. An API Gateway endpoint may accept  this as a multipart/form-data request and convert it into a Binary/Base64 string. The lambda behind this endpoint needs some means of converting that string back to the multipart payload, and in turn convert it to JSON. Using BusBoy, we can parse out any mix of standard fields and files, and return it as a JSON body.

**Remaining Work**:
- I am completely new to Jest, and as such have been having trouble getting full coverage with the tests. I followed the model laid out by the tests for the JSON body parser, but the multipart middleware is asynchronous and this seems to give rise to some issues. All tests pass, but Jest considers certain lines to still be uncovered. 5/6 of the uncovered lines are arrow functions within the Busboy event listeners. Did a lot of research on this issue but couldn't find anything similar to this case.

**Potential Issues**
- Busboy requires that the `Content-Type` header be all lowercase. This might be alleviated using the normalizeHeaders middleware, however.

Lastly, just wanna say how much middy rocks! Has made my life so much easier :)


src/middlewares/multipartFormDataParser.js
```diff
@@ -0,0 +1,49 @@
+const BusBoy = require('busboy')
+const contentTypeLib = require('content-type')
+const createError = require('http-errors')
+
+module.exports = () => ({
+  before: async (handler, next) => {
```
lmammino: Since this middleware should be compatible with Node6 (still not deprecated in middy v0.x and won't probably even in middy 1.0.x). We should not use directly `async`/`await` in middlewares. If a user will try to use this middleware in Node 6 they would have a failure

src/middlewares/multipartFormDataParser.js
```diff
@@ -0,0 +1,49 @@
...
+    const { headers } = handler.event
+    if (!headers) {
+      return next()
+    }
+
+    const contentType = headers['content-type']
+    if (contentType) {
+      const { type } = contentTypeLib.parse(contentType)
+      if (type === 'multipart/form-data') {
```
lmammino: We should handle the case when the type is not what we expect and just call `next()`

I would prefer a reverse logic and do an early return as you already did in lines `8-10`. Something like:

```javascript
if (type !== 'multipart/form-data') {
  return next()
}
```

src/middlewares/multipartFormDataParser.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = () => ({
+  before: async (handler, next) => {
+    const { headers } = handler.event
+    if (!headers) {
+      return next()
+    }
+
+    const contentType = headers['content-type']
+    if (contentType) {
```
lmammino: Missed this one before...

We should something like below:

```javascript
if (typeof contentType === 'undefined') {
  return next()
}
```

src/middlewares/multipartFormDataParser.js
```diff
@@ -0,0 +1,49 @@
...
+  return new Promise((resolve, reject) => {
+    bb
+      .on('file', (fieldname, file, filename, encoding, mimetype) => {
+        let attachment = {
+          filename,
+          mimetype,
+          encoding
+        }
+
+        file.on('data', data => { attachment.content = data })
```
lmammino: Not sure about this one.

Does `BusyBoy` emit only one chunk when all the data is converted or could emit multiple times as soon as there's some new chunk of data available (as any common stream implementation)?

In the latter case, we should somehow concatenate the new chunks to the content as soon as they are emitted rather than just assigning.

Also, would `content` be a buffer or a string?

TeddyHandleman:
@lmammino happy to make the changes you requested, work schedules a little tight this week but will try to fit em in :)

lmammino:
No worries @TeddyHandleman, take your time and let me know if you need any help or if you have questions

src/middlewares/multipartFormDataParser.js
TeddyHandleman: Sounds good. Implemented.

src/middlewares/multipartFormDataParser.js
TeddyHandleman: Good point, removed.

src/middlewares/multipartFormDataParser.js
TeddyHandleman: I implemented this as a simple else clause, any particular reason to check the type? Seems we just need to know if contentType is false or not.

src/middlewares/multipartFormDataParser.js
TeddyHandleman: This was something I was also unsure of. I have this middleware implemented in my current project and I have only ever seen the data event fire once for a file. Changed this section to keep track of the chunks and their length, and on the end event use `Buffer.concat`

TeddyHandleman:
@lmammino ended up having time to fiddle with this. A couple things:

1. By typings do you mean the middleware.d.ts file? That seems to be the most relevant. This currently takes no config so there is not interface to define.

2. To get full coverage I had to test the promise directly i.e .`expect(parseMultipartData(event)).resolves.toEqual(...)`, `expect(...).rejects.toThrow(...)`. However, this means that I have to export the middleware as well as the function that actually parses the body. This is a slight change to the code style. I was initially using the [rewire](https://www.npmjs.com/package/rewire) package, but wanted to avoid adding a dependency just for this one case.

src/middlewares/multipartFormDataParser.js
lmammino: Thanks :)

src/middlewares/multipartFormDataParser.js
lmammino: Tnx

src/middlewares/multipartFormDataParser.js
lmammino: Not really. Kind of a default practice to be honest. In this case, it might actually make more sense to do a *falsy* check as we want to check if the content type is `null` or `false`...

src/middlewares/multipartFormDataParser.js
lmammino: I was finally able to have a look at `busboy` (which I am quite liking, by the way... 😛 ) and it seems that it is implemented as a `WritableStream`, which means that if you have a reasonably big input in a form (imagine a big picture > 16kb in most cases), you could see multiple chunks being emitted, so you have to concatenate them accordingly.

I'll try to review this better and see if the code needs to be adjusted accordingly

lmammino: I am also realizing that the way we are implementing this is "eager", meaning that we are loading all the content of the form in memory as busboy is parsing it.

This is generally not a big deal, and it's kind of an expected pattern in lambda (since we have to construct a single big object that represent all data for the current event).

Anyway this approach is generally an anti-pattern especially when managing big files, so I'll make sure to drop a note about this in the README, just to raise awareness with the users about the lambda trade-offs we are making here.

lmammino:
I did another round of review and it looks good.

I will do the version bump and the README changes as mentioned in the comments above and get this merged (hopefully today).

I was also thinking that it might make sense to expose `busboy` creation options to allow people to fine-tune the stream in case they need to achieve something more sophisticated/bespoke. This should be a very easy change, so I can take care of that :)

lmammino:
Did some more changes as per the latest commit:

> - renamed middleware to httpMultipartBodyParser  …
> - Added docs
> - Added support for specifying busboy options
> - Removed double export (exporting only middleware for consistency)

If CI pass, I'll version bump and get this merged.

@TeddyHandleman we should "forward"-port this middleware to the upcoming v1.0.0 (on the dedicated branch). Would you like to take care of that?

I will create an issue anyway to keep track of this

src/middlewares/multipartFormDataParser.js
dbartholomae: @lmammino I know I'm chiming in on an already merged PR, but I have actually been working on middy + busboy just last month, and maybe some of the things I learned might be of interest. Currently there are some strict limits on file sizes for both lambdas ([max. 6 MB invocation payload](https://docs.aws.amazon.com/lambda/latest/dg/limits.html)) as well as API Gateway ([max. 10 MB payload size](https://docs.amazonaws.cn/en_us/apigateway/latest/developerguide/limits.html)). In the end we abandonend the approach and now use an Elastic Beanstalk server. The 6MB limit might be of interest for anyone trying to use lambda with form-data.

lmammino: Good point. This is what I ended up writing for the docs. Feel free to submit a PR if you feel it's worth a redaction or an addition :)

PR #319, opened 2019-04-08
Make ICorsOptions fields optional

niik:
None of these fields are required when using the `http-cors` module but the type doesn't reflect that and forces users consuming the API via TypeScript to either cast their type to any or fill in the default values manually.

Looking at other middleware options it appears as if they all follow the same pattern of marking most fields as optional.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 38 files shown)


lmammino:
Thans @niik, I'll try to get this one merged shortly

PR #326, opened 2019-04-26
Bug fix: Allow multiple environment variables to be set to the same SSM parameter via name

garyhollandxyz:
The SSM middleware currently has a bug where, when using names to access SSM parameters, the last of multiple names pointing at the same SSM parameter will overwrite the others.

E.g. with the following middleware options, only `KEY_NAME_2` will be set.
```javaScript
names: {
          KEY_NAME_1: '/dev/service_name/key_name',
          KEY_NAME_2: '/dev/service_name/key_name'
}
```

This PR fixes this bug and deduplicates the SSM parameters which are being accessed by name so that in the above scenario, `/dev/service_name/key_name` is fetched exactly once and then set onto both `KEY_NAME_1` and `KEY_NAME_2`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy",
-  "version": "0.25.0",
+  "version": "0.25.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


lmammino:
That's pretty sweet :)

I'll try to get this one merged shortly, meanwhile, would you like to submit the same fix for branch 1.0.x?

garyhollandxyz:
Sure thing :)

PR #332, opened 2019-04-30
Support reviver parameter in for http-json-body-parser

roni-frantchi:
## Description
This PR introduces a [`reviver`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Parameters) option for `http-json-body-parser` to pass through to `JSON.parse`.  

## Motivation
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse):
> An optional reviver function can be provided to perform a transformation on the resulting object before it is returned.

A `1.0-alpha` PR is also available in #331

### DIFF
--- a/middlewares.d.ts
+++ b/middlewares.d.ts
@@ -46,6 +46,10 @@ interface IHTTPPartialResponseOptions {
   filteringKeyName?: string;
 }
 
+interface IJsonBodyParserOptions {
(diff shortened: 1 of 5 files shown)


lmammino:
THanks @roni-frantchi for this one too :)

Issue #270, opened 2019-01-10
Jest tests for middleware swallow expect errors


iDVB:
I was authoring a custom middy middleware and I implemented Jest tests similar to how you implement them for the core middlewares. (no idea how otherwise to do it) However, many of the tests for core middleware do not **fail** gracefully...

When a test **fails** a few things happen that are odd:
- the `Received` value is always `undefined`
- the handler calls the callback multiple times

Core `cors` middleware for example:
![image](https://user-images.githubusercontent.com/189506/50945802-6ed13400-1464-11e9-9796-452f294718c9.png)
I just caused it to fail by removing the `*` from the expected `Access-Control-Allow-Origin` 

In order to fix the problem for my own tests which also worked for the core ones, I had to throw when there was an error. 
```javascript
if (err) throw err; 
```

Example using `cors` middleware:

```javascript
test('Access-Control-Allow-Origin header should default to "*"', () => {
    const handler = middy((event, context, cb) => {
      cb(null, {});
    });

    handler.use(cors());

    const event = {
      httpMethod: 'GET',
    };

    handler(event, {}, (err, response) => {
      if (err) throw err;
      expect(response).toEqual({
        headers: {
          'Access-Control-Allow-Origin': '',
        },
      });
    });
  });
```

I'm new to jest and new to middy. Is there a reason the core middy tests don't fail properly? Or am I just missing something?

iaindurham:
When testing async code using callbacks in jest you need to pass in the `done` method as an argument to the test, then call it once the tests have completed. If you don't do this then the code operation won't pause for the outcome of the test which can cause unexpected results depending on when the code gets run.

See the [Jest docs ](https://jestjs.io/docs/en/asynchronous#callbacks) for coding for callbacks.

One thing that the docs don't mention is that if you pass an argument to `done` in your test, the test will fail with the argument returned as an error message on the console.

So your above example will look like:

```js
  test('Access-Control-Allow-Origin header should default to "*"', (done) => {
    const handler = middy((event, context, cb) => {
      cb(null, {});
    });

    handler.use(cors());

    const event = {
      httpMethod: 'GET',
    };

    handler(event, {}, (err, response) => {
      if (err) return done(err);
      expect(response).toEqual({
        headers: {
          'Access-Control-Allow-Origin': '',
        },
      });
      done();
    });
  });
```

iDVB:
[...]

PR #335, opened 2019-05-06
fix: content type request bug in http serializer

noetix:
This resolves the bug reported https://github.com/middyjs/middy/issues/169#issuecomment-487373166.

I've also made the tests async aware in case of test failures as per https://github.com/middyjs/middy/issues/270

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-alpha.31"
+  "version": "1.0.0-alpha.32"
 }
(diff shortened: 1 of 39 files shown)


roni-frantchi:
Thanks addressing my issue @noetix .  I'll test this branch today with our solution and report back.
There's another issue we've encountered that's somewhat related, when there are no headers set, getting:
```
    TypeError: Cannot read property 'headers' of undefined
        at opts.serializers.map.s (/Users/roni/dev/env0/node_modules/@middy/http-response-serializer/index.js:45:67)
```

lmammino:
@roni-frantchi, I added some changes that should address that one too. About to merge and publish, let me know if this solves your issue, otherwise feel free to open a dedicated issue to track that :)

noetix:
Thanks @lmammino, had this on my to-do list to fix.

Issue #325, opened 2019-04-25
httpErrorHandler breaks if code is minimized / mangled


akleiber:
# Problem

httpErrorHandler compares error's constructor name - but the check fails if code is minimized

There already was an issue open but was closed by the autor: https://github.com/middyjs/middy/issues/271

The problem still exists.
If code is optimized, for example via webpack than the constructor.name value is changed and the check will fail. Therefor the middleware will do nothing for http errors.
See https://github.com/middyjs/middy/blob/0.25.0/src/middlewares/httpErrorHandler.js#L10

Webpack optimization infos: https://webpack.js.org/configuration/optimization/

## Solution 1

Do not check for the constructor to check if the error is an http error.

##  Solution 2

Disable code minification

### Disable minification

```
module.exports = {
  //...
  optimization: {
    minimize: false
  }
};
```

### Do not mange classnames

```
const TerserPlugin = require('terser-webpack-plugin');

module.exports = {
  optimization: {
    minimizer: [
      new TerserPlugin({
        cache: true,
        parallel: true,
        sourceMap: true, // Must be set to true if using source-maps in production
        terserOptions: {
          // https://github.com/webpack-contrib/terser-webpack-plugin#terseroptions
            keep_classnames: true
        }
      }),
    ],
  }
```

lmammino:
Not sure if this option was discussed already, but the following might be a solution:

```javascript
const { HttpError } = require('http-errors')

// ...

if (someError instanceof HttpError) {
  // ...
}
```

What do you think?

cc @dkatavic

lmammino:
From version `0.27.0` (and shortly also from `1.0.0-alpha.33`) any error object containing the fields `statusCode` and `message` will be considered valid to run the middleware.
This solves any minification issue, but also makes you more flexible to use other libraries for error management, or simply allows you not to depend on a library for http errors.

PR #337, opened 2019-05-10
fix: httpErrorHandler minification type checking

chrisandrews7:
If the code is mangled during minification process when using webpack, the check for a type of `HttpError` in the httpErrorHandler middleware will fail. As the private property on `super_.name` will not be `HttpError`, the mangling would change the name.

This is due to [Terser](https://github.com/webpack-contrib/terser-webpack-plugin) setting both `keep_classnames` and `keep_fnames` to false by default.
Either way it shouldn't really use a private property to check the type of error.

This will close https://github.com/middyjs/middy/issues/325

### DIFF
--- a/docs/middlewares.md
+++ b/docs/middlewares.md
@@ -270,11 +270,10 @@ module.exports = { handler }
 
 ## [httpErrorHandler](/src/middlewares/httpErrorHandler.js)
 
-Automatically handles uncaught errors that are created with
-[`http-errors`](https://npm.im/http-errors) and creates a proper HTTP response
-for them (using the message and the status code provided by the error object).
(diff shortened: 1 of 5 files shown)


chrisandrews7:
This cant be merged until both https://github.com/jshttp/http-errors/issues/56 and  https://github.com/npm/npm/issues/19770 have been resolved. I guess thats why it was done in the current way.

chrisandrews7:
Given the only thing used by the response is `statusCode`. Could do a check on the `handler.error.statusCode` property as the type identifier in the meantime? @lmammino @dkatavic

lmammino:
I actually added an extra check for `message` as well, added a dedicated test case and updated the docs.

I think I have enough time to forward port this to 1.0.0

PR #345, opened 2019-05-23
Update warmup.js to work alongside with doNotWaitForEmptyEventLoop

lbertenasco:
This is required when using both `warmup` & `doNotWaitForEmptyEventLoop`.

### DIFF
--- a/src/middlewares/warmup.js
+++ b/src/middlewares/warmup.js
@@ -10,6 +10,7 @@ module.exports = (config) => {
     before: (handler, next) => {
       if (options.isWarmingUp(handler.event)) {
         options.onWarmup(handler.event)
+        handler.context.callbackWaitsForEmptyEventLoop = false;
         return handler.callback(null, 'warmup')
       }
 
--- a/src/middlewares/warmup.js
+++ b/src/middlewares/warmup.js
@@ -1,7 +1,8 @@
 module.exports = (config) => {
   const defaults = {
     isWarmingUp: (event) => event.source === 'serverless-plugin-warmup',
-    onWarmup: (event) => console.log('Exiting early via warmup Middleware')
+    onWarmup: (event) => console.log('Exiting early via warmup Middleware'),
+    waitForEmptyEventLoop: null
   }
 
   const options = Object.assign({}, defaults, config)
@@ -10,7 +11,9 @@ module.exports = (config) => {
     before: (handler, next) => {
       if (options.isWarmingUp(handler.event)) {
         options.onWarmup(handler.event)
-        handler.context.callbackWaitsForEmptyEventLoop = false;
+        if (waitForEmptyEventLoop !== null) {
+          handler.context.callbackWaitsForEmptyEventLoop = Boolean(options.waitForEmptyEventLoop);
+        }
         return handler.callback(null, 'warmup')
       }
 
--- a/src/middlewares/warmup.js
+++ b/src/middlewares/warmup.js
@@ -11,8 +11,8 @@ module.exports = (config) => {
     before: (handler, next) => {
       if (options.isWarmingUp(handler.event)) {
         options.onWarmup(handler.event)
-        if (waitForEmptyEventLoop !== null) {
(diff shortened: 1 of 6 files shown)


lmammino:
That's an interesting one.
It makes sense, but I am curious to know in which case a warmup lambda can have a hold on the event loop?

Are you instantiating a database connection or something like that in your lambda?

lbertenasco:
Yes, indeed.
But maybe there is a way to avoid having to change the event loop.
This was a quick and easy fix I did locally and assumed it might be useful to have it on the official repo.

lmammino:
Sorry for taking some time to reply to this.
I was reflecting on this being a non obvious side effect.

One way we could make it a little bit more specific is maybe to expose an option on the middleware, maybe something like this:

```diff
module.exports = (config) => {
  const defaults = {
    isWarmingUp: (event) => event.source === 'serverless-plugin-warmup',
-   onWarmup: (event) => console.log('Exiting early via warmup Middleware')
+   onWarmup: (event) => console.log('Exiting early via warmup Middleware'),
+   waitForEmptyEventLoop: true
  }

  const options = Object.assign({}, defaults, config)

  return ({
    before: (handler, next) => {
      if (options.isWarmingUp(handler.event)) {
        options.onWarmup(handler.event)
-       handler.context.callbackWaitsForEmptyEventLoop = false;
+       handler.context.callbackWaitsForEmptyEventLoop = Boolean(options.waitForEmptyEventLoop);
        return handler.callback(null, 'warmup')
      }

      next()
    }
  })
}
```

I am also keeping here the default value to `true` so we don't have breaking changes.

What do you think?

lbertenasco:
It makes sense.

Not sure if it will be better to initialize it as `null` and only modify `callbackWaitsForEmptyEventLoop` if there is an assigned boolean value.
> This is backwards compatible and might also avoid future issues where the user has to manually set the property on the middleware.

The naming on `doNotWaitForEmptyEventLoop` is a bit confusing too.
I'd rather have `waitForEmptyEventLoop(false)`
But this is another topic 😅

lmammino:
> Not sure if it will be better to initialize it as null and only modify callbackWaitsForEmptyEventLoop if there is an assigned boolean value.

I like this idea!

> The naming on doNotWaitForEmptyEventLoop is a bit confusing too.
> I'd rather have waitForEmptyEventLoop(false)

Also this one! :)

Would you like to do the necessary changes to the PR? It would be great if you could include changes to the docs and tests.

src/middlewares/warmup.js
```diff
@@ -10,7 +11,9 @@ module.exports = (config) => {
     before: (handler, next) => {
       if (options.isWarmingUp(handler.event)) {
         options.onWarmup(handler.event)
-        handler.context.callbackWaitsForEmptyEventLoop = false;
+        if (waitForEmptyEventLoop !== null) {
```
lmammino: Shouldn't this be something like `options.waitForEmptyEventLoop`?

lmammino:
Hello @lbertenasco, thanks for adding the necessary changes. I left a comment there.

It would be awesome if you could also add some tests to cover this new behaviour?

Thanks

lbertenasco:
Hey @lmammino ! Yes, sorry about the delay 😅 

I'll push the new changes today, which include the `options. waitForEmptyEventLoop` fix.

About this:
 
> The naming on doNotWaitForEmptyEventLoop is a bit confusing too.
> I'd rather have waitForEmptyEventLoop(false)

It might be a good idea to create a separate PR for this. But I'm not sure how you manage breaking changes. (Maybe we could export an alias for `doNotWaitForEmptyEventLoop`)
WDYT?

lmammino:
> It might be a good idea to create a separate PR for this. But I'm not sure how you manage breaking changes. (Maybe we could export an alias for doNotWaitForEmptyEventLoop)
WDYT?

I don't think there's any problem with this. This is a new option in this middleware so it's not really a BC. I am happy to ship this as it is :) (which i'll do shortly under `v0.28.0`)

PR #359, opened 2019-06-27
Update documentation cache middleware

ElinksFr:
If your `setValue` function return a Promise that resolve with a value it will be passed to `next` triggering an error when there is none. 

We could update the code of `src/middlewares/cache.js` doing 
```javascript
options.setValue(currentCacheKey, handler.response)
    .then(() => next())
    .catch(err => next(err))
```

instead of

```javascript
options.setValue(currentCacheKey, handler.response)
    .then(next)
```

But I am not sure of the impacts it could have if some people relly on the current implementation


dkatavic:
Good catch @ElinksFr , the current implementation will leave the error from `setValue` method unhandled. Do you wanna proceed with the fix in the code?

lmammino:
I agree with @dkatavic, a fix in the code seems the right way to solve this. Would you like to update your PR to do that?

ElinksFr:
Sorry for the delay. I updated the PR based on your feedbacks and added 2 tests

Issue #373, opened 2019-08-03
[TypeScript] Documentation/Examples for writing custom middleware


lakshmi-kannan:
First, thanks for this library. You almost saved me from writing boilerplate code. I have a self inflicted problem - I decided on using typescript for my lambda functions and I have a use case to write custom middleware to take the idToken/JWT from Authorization header and then decode it to get the user name which will be used in my business logic. 

Unfortunately, there aren't any examples of middlewares written in pure typescript in the repo. I went through the definitions and also the github issues trying to whip up one but failing miserably. Part of the problem is I am very new to typescript. It would be awesome if you can add a non-trivial example of writing a custom middleware in pure typescript in the repo. I can use that to write mine and probably contribute back what I am doing upstream. Decoding JWT from Cognito idToken and getting the user name seems like something that'd be useful.

lmammino:
Hello @lakshmi-kannan!
Thanks for using Middy :)

Unfortunately, I am not experienced with TypeScript to be able to provide a set of good examples. In any case, even by using TypeScript, you should be able to use plain JavaScript in your TypeScript projects, so, I believe, the examples here should work out of the box.

Let me know if you make any progress and if eventually, you would like to open a documentation PR to add this kind of examples.

lakshmi-kannan:
@lmammino Thanks! I finally got a custom middleware written in Tyepscript and hooked it up to run before my handler and it works. Sample code below: 

```export function idTokenMiddleware(): MiddlewareObject<any, any> {
  return {
    before: async (
      handler: HandlerLambda<APIGatewayEvent, APIGatewayProxyResult>
    ): Promise<void> => {
      console.log('ID Token parse middleware');
      const headers: { [name: string]: string } = handler.event.headers || {};
      const authHeader: string = 'Authorization';

      if (authHeader in headers) {
        const bearerToken: string = headers.Authorization;
        const idTokenStr: string = parseBearerToken(bearerToken);
        if (idTokenStr) {
          const idToken: IdToken = await parseIdToken(idTokenStr);
          console.log(`id token - ${idToken}`);
          return;
        }
      }

      console.log('Done ID token parse middleware before.');
      return Promise.reject(undefined);
    }
  };
}
```

However, there are couple of issues

1. I don't know how to pass the output from this async middleware to the handler. Reading through middy.js, it doesn't look like a return value from a middleware is fed into the next one. Is the assumption that middlewares are "stateless"? How can I pass a context object across all middlewares? Any particular examples? 

[...]

PR #382, opened 2019-08-25
Expose types in core TypeScript typings

sheepsteak:
In v1.0.0-alpha it's hard to make your own middleware with TypeScript because none of the middleware types are exported.

I've made some changes to the `core` typings to expose the types to allow making middleware with TypeScript.

For example, you can now do this:

```typescript
import { IHandlerLambda, IMiddyMiddlewareObject } from "@middy/core";
import { APIGatewayEvent } from "aws-lambda";

const myMiddleware = (): IMiddyMiddlewareObject => ({
  before: (handler: IHandlerLambda<APIGatewayEvent>) => {
    console.log(handler.event.path);

    return Promise.resolve();
  },
});

export default myMiddleware;
```

This should open the way for #373 to be worked on now too.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -1,46 +1,42 @@
 import { Callback, Context, Handler, ProxyResult } from 'aws-lambda'
 
-type AsyncHandler = (
+export type AsyncHandler = (
   event: any,
   context: Context
 ) => Promise<ProxyResult | object>
 
-declare var middy: {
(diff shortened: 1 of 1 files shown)


sheepsteak:
Whilst doing #383 I noticed that you're using `middy.IMiddyMiddlewareObject` in middleware TypeScript typings. I've not really used namespaces before so I had no idea you could do that. I'm going to close this because it doesn't offer anything new.

PR #383, opened 2019-08-25
Fix relative imports

sheepsteak:
Build on #315 by fixing all relative imports of `@middy/core` in 1.0.0-alpha.

- [x] Add `@middy/core` to `devDependencies` of packages
- [x] Change all relative imports

Closes #315.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-alpha.41"
+  "version": "1.0.0-alpha.42"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy-monorepo",
-  "version": "1.0.0-alpha.40",
+  "version": "1.0.0-alpha.42",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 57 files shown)


lmammino:
Hello @sheepsteak! First of all, thanks for this PR.
Anyway, I have a couple of questions.

I agree that we need to use absolute imports for typings (i guess otherwise they might not work as expected in certain scenarios.

I also agree about @middy/core being a devDependency, since we only use that within tests.

I am not so sold on why we want to change relative imports for tests. I have a feeling that keep using relative imports makes evolving the codebase easier by allowing us to keep everything tested and in sync before publishing a new release.

Am I missing something?

sheepsteak:
> I am not so sold on why we want to change relative imports for tests. I have a feeling that keep using relative imports makes evolving the codebase easier by allowing us to keep everything tested and in sync before publishing a new release.

Sorry, I have no idea why I did this. I must have just seen `../../core` and thought it needed changing too. 😇

I've removed that commit and merged in latest 1.0.0-alpha.

lmammino:
Thanks @sheepsteak.

I have another doubt (being a typescript noob!).

In the typings file we do:

```typescript
```

Will that work since we have `@middy/core` as a dev dependency or do we need to move that into `dependencies`?

vladgolubev:
@lmammino it will work, since it is also in `peerDependencies`, so user won't get around w/o `@middy/core` installed

lmammino:
Thanks @vladgolubev.

@sheepsteak I'll try to get this merged this by today/tomorrow! :)

PR #391, opened 2019-09-05
Rewrite tests to use async/await (fix #270)

### LINKED ISSUE #270: Jest tests for middleware swallow expect errors

vladgolubev:


lmammino:
Great job @vladgolubev, There's one test failing. Any clue why?

lmammino:
Fixed tests. Releasing this shortly :)

PR #401, opened 2019-09-30
TS: Make it possible to extend context type with typescript

zaqqaz:
These changes needed to make it possible (and easier) to extend context in runtime with typescript support.

So now we can use it like this: 

```ts
// Defining our extended type of the context
export interface ContextWithExtraArgs extends Context {
    extra?: ExtraType;
}

const before: MiddlewareFunction<SQSEventWithLogger, Response, ContextWithRecords> = ({
    event,
    context,
}) => {
    context.extra = someExtraValue;
};

const withExtraArgs: MiddlewareObject<SQSEventWithLogger, Response, ContextWithRecords> = {
    before,
};

const originalHandler = async (event: SQSEvent, context: ContextWithExtraArgs): Promise<void> => {
    // now we can use extra args here and typescript will happy as well
    console.log(context.extra);
};

const handler = applyMiddleware(originalHandler).use(withExtraArgs);
```

Thanks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -24,28 +24,28 @@ declare const middy: <H extends AsyncHandler<C>, C extends Context = Context>(ha
 >;
 
 declare namespace middy {
-  interface Middy<T, R> extends Handler<T, R> {
-    use: <C extends MiddlewareObject<T, R>>(middleware: C) => Middy<T, R>;
-    before: (callbackFn: MiddlewareFunction<T, R>) => Middy<T, R>;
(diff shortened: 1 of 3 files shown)


zaqqaz:
@vladgolubev, hey 👋. 
As I see that you've approved these changes, can we get them merged?

lmammino:
hey @zaqqaz, I'll get them merged shortly

lmammino:
@zaqqaz, Could you also submit those changes to the 1.0.0 branch?

PR #402, opened 2019-09-30
Update types to reflect `aws lambda` types.

### LINKED ISSUE #398: handler.callback taking the wrong arguement from generic function type
Given the following type definitions in `aws lambda`

```ts
export type Handler<TEvent = any, TResult = any> = (
    event: TEvent,
    context: Context,
    callback: Callback<TResult>,
) => void | Promise<TResult>;
```

`callback` take the type from `TResult` however in `middy` we have the following type decleration

```ts
  interface HandlerLambda<T = any, V = {}> {
    event: T;
    context: Context;
    response: V;
    error: Error;
    callback: Callback<T>;
  }
}
```

callback seems to be taking the wrong type and unable to use `APIGatewayProxyResult`

amwill04:
Closes #398

Updates `HandlerLambda` types

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,12 +43,12 @@ declare namespace middy {
 
   type NextFunction = (error?: any) => void;
 
-  interface HandlerLambda<T = any, V = {}> {
+  interface HandlerLambda<T = any, V = any> {
     event: T;
     context: Context;
     response: V;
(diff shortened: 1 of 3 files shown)


lmammino:
Thanks a lot @amwill04

PR #404, opened 2019-10-02
bug: validator coercion

### LINKED ISSUE #403: Validator type coercion issue
When using the [validator middleware](https://www.npmjs.com/package/@middy/validator) and specifying a schema with a type string, if the value is a number it will be coerced into a string and pass validation.

Illustrated below. The validator will mark the event as valid, and coerce the `foo` variable into a string. Despite the value actually being a number.

```js
const schema = {
  required: ['foo'],
  properties: {
    foo: {
      type: 'string'
    }
  }
}

const handler = middy((event, context, cb) => {
  console.log(typeof event.foo) // string???
}).use(validator({
  inputSchema: schema
}))

handler({
  foo: 123
}, {}, () => {})   
```

I've written a [test](https://github.com/chrisandrews7/middy/blob/06f7e583873222c37df50972638f06f7706fd43a/src/middlewares/__tests__/validator.js#L66) that covers this.

The problem looks to be from [coerceTypes](https://github.com/chrisandrews7/middy/blob/06f7e583873222c37df50972638f06f7706fd43a/src/middlewares/validator.js#L11) config in [ajv](https://www.npmjs.com/package/ajv).
As after removing that the test above passes, as do all of the other tests.

chrisandrews7:
Closes #403.

Doesn't appear to break any other tests related to query string params?

### DIFF
--- a/src/middlewares/__tests__/validator.js
+++ b/src/middlewares/__tests__/validator.js
@@ -63,6 +63,34 @@ describe('📦  Middleware Validator', () => {
     })
   })
 
+  test('It should handle numbers as strings as a BadRequest', () => {
+    const handler = middy((event, context, cb) => {
+      cb(null, event.body) // propagates the body as a response
+    })
+
+    const schema = {
+      required: ['foo'],
+      properties: {
+        foo: {
+          type: 'string'
+        }
+      }
+    }
+
+    handler.use(validator({
+      inputSchema: schema
+    }))
+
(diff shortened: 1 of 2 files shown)


lmammino:
hello @chrisandrews7, thanks for this PR, but I agree with the points raised by @willfarrell in the related issue (#404). Do you agree that his suggestions can address your problem? If so, can we close this PR?

Thanks

PR #405, opened 2019-10-02
Database manager as a middleware

sdomagala:
Hey,
I felt that middy is lacking database connection out of the box, so I've created this plugin.

I'm not sure about adding `knex` to middy as a whole though, do you feel that we should provide this default client as it is now or user should specify what tool he would like to use and how he'd like to destroy it(if needed)?

Cheers


lmammino:
hey @sdomagala, this seems very good, thanks!

What about moving this PR to version 1.0.0-alpha? It might need some small rework, but I will lock the current master branch in the next few weeks when 1.0.0-beta will be out and it's probably not worth the effort to have it published in both releases.

What do you think?

sdomagala:
Hey @lmammino,
I knew there was a branch that I was supposed to replicate this feature to, but forgot what it was 😉 
I will replicate this behaviour to this alpha branch, although, with this new way of doing things, do you plan to keep nodejs6.10 supported with this new release?

Also, is there a support for TypeScript on this alpha branch? Just by quickly checking it out I saw some typescript definitions but no ts package yet.

lmammino:
I am thinking to close this one and leave this feature only in the 1.0.0 branch. What do you think @sdomagala ?

sdomagala:
What would be the reason to have it on alpha branch but not in the official release?
I don't mind personally leaving it in alpha branch as I have it as local plugin right now, it's up to you really.

lmammino:
Sorry i didn't realize you already moved this to JS. It was more to avoid you the effort to work on this other PR. I guess at this point it isn't much effort to get this merged. I'll try to do it soon ;)

onlyplo:
Just to note, now all lambdas will need to compile / download Knex upon `npm install` which is kinda heavy.. unless users swap to typescript this PR has just bloated everyone's lambda...

Database connections feel very use case specific. Perhaps we can spin this off to a third party middleware instead?

simontabor:
Personally I think all of the bundled middlewares should be extracted into independent packages to avoid this kind of bloat and keep the "core" of middy very small and focussed

sdomagala:
Thanks for your opinion @onlyplo, that was my first question when I was opening this PR - do we want to add knex as dependency. I don't mind removing it completely and forcing user to provide client on his own but it didn't feel "out-of-the-box" to me. I like @simontabor's idea though about having every middleware separate so then it can have its own dependencies

gsingh1:
Relates to https://github.com/middyjs/middy/issues/440

theburningmonk:
@lmammino can we move this into a separate package like we did with the puresec middleware before? This is causing folks a fair bit of grief in terms of slower cold start and issues with webpacking their functions.

We need to be more conscious of adding middlewares that bring in big dependencies like `knex`, at least until 1.0.0 is out. Right now the validator plugin (which not many people use) is already adding 50-70ms to everyone's cold start and let's not make things worse.

lmammino:
> Personally I think all of the bundled middlewares should be extracted into independent packages to avoid this kind of bloat and keep the "core" of middy very small and focussed

This is what we are working on for version 1.0.0 (currently in alpha, but quite usable already).

> can we move this into a separate package like we did with the puresec middleware before?

I think at this point this is a sensible approach. We already have the package published under the 1.0.0 release and that should work even with the 0.x core with the caveat that you use at least node 10.

The reason for the caveat is that, knex dropped support for node 6-8 and there are vulnerabilities in the version currently used in 0.x.

Again, I agree we should move this package out, which basically means:

- Updating the readme to specify that the package has been removed from the core
- Add an explanation on how to get the new package from 1.0.0-alpha and the caveat that it works only with node10+
- remove the package from the codebase

I won't have much time to do the changes in the next few days, unfortunately, so I'd appreciate any help with this

gsingh1:
@lmammino @theburningmonk - PR raised https://github.com/middyjs/middy/pull/450 for above issues and removal of this middleware from core on 0.x

PR #412, opened 2019-10-14
move @types/aws-lambda to a peerDependency

### LINKED ISSUE #409: 1.0 - Shipping @types/aws-lambda as a dep (and not peer-dep) is breaking our builds
After the latest update of our dependencies where both our own @types/aws-lambda version was updated to `8.10.33` and we upgraded to @middy/core `1.0.0-alpha.45`.

Now we are getting errors like so:

```
The inferred type of 'myMiddleware' cannot be named without a reference to '@middy/core/node_modules/@types/aws-lambda'. This is likely not portable. A type annotation is necessary.
```

I think you are better off shipping @types/aws-lambda as a peer-dep. When I remove @middys version of aws-lambda, my build starts working again.

duro:
I've moved the @types/aws-lambda package to peer dependencies, as some project may have a slightly different version of these types, and when both exist in the same project, it will result in TypeScript errors like:

```
The inferred type of 'myMiddleware' cannot be named without a reference to '@middy/core/node_modules/@types/aws-lambda'. This is likely not portable. A type annotation is necessary.
```

I have also updated the README to reflect this for TypeScript users

closes #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ To install middy you can use NPM:
 npm install --save @middy/core
 ```
 
+If you are using TypeScript, you will also want to make sure that you have installed the @types/aws-lambda peer-dependency
+
+```bash
+npm install --save-dev @types/aws-lambda
+```
+
 ## Quick example
 
 code is better than 10,000 words, so let's jump into an example.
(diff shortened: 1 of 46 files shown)


lmammino:
Thanks a lot @duro! I'll get this reviewed and merged shortly!

PR #413, opened 2019-10-16
[typings] Make middy.use accept array of middlewares as well

vladgolubev:


vladgolubev:
@KnupMan could you please take a look?

packages/core/index.d.ts
```diff
@@ -25,7 +25,7 @@ declare const middy: <H extends AsyncHandler<C>, C extends Context = Context>(ha
 
 declare namespace middy {
   interface Middy<T, R, C extends Context = Context> extends Handler<T, R> {
-    use: <M extends MiddlewareObject<T, R, C>>(middleware: M) => Middy<T, R, C>;
+    use: <M extends MiddlewareObject<T, R, C>>(input: M | M[]) => Middy<T, R, C>;
```
lmammino: Very good one! Should we just rename `input` to `middlewares`? I believe it's slightly more descriptive

lmammino:
Awesome work @vladgolubev ! Thanks!

PR #415, opened 2019-10-18
Fix href for "Using async/await" link in ToC

andrhamm:


lmammino:
Thanks a lot @andrhamm :)

PR #422, opened 2019-11-11
Context specified in middy constructor should be passed forward to the Middy interface

hsz:


lmammino:
Thanks @hsz, can you check if the same issue is present also in branch 1.0.0-alpha and if we need a PR there as well?
Thanks :)

PR #433, opened 2019-12-05
fix broken link href in readme

KOREAN139:
#using-async-await is not compatible herf link to `Using async/await`
to link properly, it should use #using-asyncawait


lmammino:
Hey @KOREAN139 Thanks a lot. I'll get this merged shortly!

Issue #355, opened 2019-06-23
Remove support for node 6


lmammino:

willfarrell:
Might as well drop NodeJS v8.x while we're at it.

lmammino:
Agreed. I would it only on v1 though. Also I would add Node 12 to the CI pipeline as we do this

willfarrell:
Do we need to update the `tsconfig.json` files to use something > `es2015`?

PR #441, opened 2019-12-09
ci: deprecate old runtimes

willfarrell:
Closes: https://github.com/middyjs/middy/issues/355

FYI I'm working on middy stuff all week. Will be on Gitter as well.


lmammino:
Awesome! Thanks for the quick turnout on this :)
I'll do the version bump in a sec (which should also trigger the CI... that for some reason is not running otherwise 😕)

Issue #436, opened 2019-12-07
Lazy loading aws-sdk performance anti-pattern


willfarrell:
I'm in the process of building https://github.com/middyjs/middy/issues/428. While refactoring it to follow some of the patterns from `ssm`, I noticed warmup/coldstart times spiked.

After adding timers everywhere I found that lazy loading the `aws-sdk` as seen in `ssm` https://github.com/middyjs/middy/blob/master/src/middlewares/ssm.js#L166 added up to 4.5sec duration to a warmup/coldstart. Moving the declaration to the global scope removed the 4.5sec.

I'm running NodeJS v12.x w/ 128MB Memory.

Anyone else running into this? Maybe AWS improved something under the hood removing the once benefit of needing to lazy load?

cc @lmammino

lmammino:
Wow that's interesting. I wonder if the Lambda runtime does something special to speed up things here. Maybe @theburningmonk knows something (or knows who to ask to get some clarity about this)...

willfarrell:
Created a bare-bones example to test this. https://github.com/willfarrell/middy-serverless-perf

Using: NodeJS v12.x in `ca-central-1`
Testing on:
- *With* or *Without* `aws-sdk` as a dependency
- Lambda is *serverless* or *VPC*
- Dependency is used at the *global*, *option*, or *function* level scope.

<img width="1224" alt="Screen Shot 2019-12-09 at 08 54 19" src="https://user-images.githubusercontent.com/1097020/70450830-a0ca8c00-1a61-11ea-9a16-4ea6b9e60ef5.png">

Results show that there is indeed a jump up in execution time when a dependency is declared inside `before`, `after`, `onError`. Declaring in at the `global` level is generally a little faster than the `option` level.

I'm going to keep it running all day to get more data points. Declaring at the global and option level seem to flipflop on which is faster, I expect them to average out to be the same over time.

willfarrell:
This is a 1 hour avg, also added *warm* vs *cold* to see how that looks too (just in case). It's pretty clear that attaching a dependency in `before`, `after`, `onError` has a penalty. I'm happy to update `ssm`, and any others that have this pattern,  if you like.

Ran every 16min
<img width="1224" alt="Screen Shot 2019-12-09 at 10 28 24" src="https://user-images.githubusercontent.com/1097020/70458179-cd38d500-1a6e-11ea-962c-c0d53e69ab37.png">

Ran every 14min
[...]

PR #442, opened 2019-12-09
Dependency Performance Improvements

willfarrell:
Closes: https://github.com/middyjs/middy/issues/436

### DIFF
--- a/packages/db-manager/__tests__/index.js
+++ b/packages/db-manager/__tests__/index.js
@@ -3,7 +3,7 @@ jest.mock('knex')
 
 const knex = require('knex')
 
-const middy = require('@middy/core')
+const middy = require('../../core')
 const dbManager = require('../')
 
 describe('💾  Database manager', () => {
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-alpha.56"
+  "version": "1.0.0-alpha.57"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "middy-monorepo",
-  "version": "1.0.0-alpha.56",
+  "version": "1.0.0-alpha.57",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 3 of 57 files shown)


vladgolubev:
What about importing only relevant clients from aws-sdk? 

See https://theburningmonk.com/2019/03/just-how-expensive-is-the-full-aws-sdk/

willfarrell:
@vladgolubev Great article, and suggestion.

I'm curious, there is a common hack to enable keep-alive (below, sometimes seen as a layer). Does adding this in lobotomize cold start performance for faster aws service re-connections. If we only include the relevant clients only, does it undo keepAlive? @theburningmonk maybe you've already tested this?

... So many dimensions to test. A project for the new year.

```js
const AWS = require('aws-sdk')
const https = require('https')

const tlsAgent = new https.Agent({
  keepAlive: true,
  maxSockets: 50,
  rejectUnauthorized: true
})
tlsAgent.setMaxListeners(0)

AWS.config.update({
  httpOptions: {
    agent: tlsAgent
  }
})

module.exports = AWS
```

vladgolubev:
@willfarrell see https://theburningmonk.com/2019/02/lambda-optimization-tip-enable-http-keep-alive/

> UPDATE 22/09/2019: big thanks to Joe Bowbeer for mentioning this in the comments. Since [AWS SDK v2.463.0](https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md#24630) you no longer need these couple of lines of code change any more. Instead, set the environment variable AWS_NODEJS_CONNECTION_REUSE_ENABLED to 1 to make the SDK reuse connections by default.

packages/db-manager/__tests__/index.js
```diff
@@ -3,7 +3,7 @@ jest.mock('knex')
 
 const knex = require('knex')
 
-const middy = require('@middy/core')
+const middy = require('../../core')
```
lmammino: any reason for changing this?

willfarrell: Just to make it consistent with all the other modules. This ensures they're all being tested with the same version of core as the rest.

Issue #414, opened 2019-10-17
promise support documentation


amwill04:
Hey

I dont know if I am completelty misusing this or not but given the following:

```javascript
const middy = require('middy');


let count = 0

const handler = async function() {
  console.log(count)
  count++
  return {
        body: '{}',
        headers: {
            'content-type': 'application/json',
        },
        statusCode: 0,
    };
}

const asyncFunction = function() {
  return new Promise(function(resolve) {
    setTimeout(function(){
      resolve()
    }, 5000)
  })
}

const validator = function() {
  return {
    before: async function(handler, next) {
      console.log('start');
      await asyncFunction();
      console.log('end');
      next()
    }
  }
}

exports.handler = middy(handler).use(validator());
```

the `count++` will be duplicated and provide different results:

```
START RequestId: 265b8c95-de83-4478-a821-e8ea7af078be Version: $LATEST
2019-10-17T15:55:56.289Z	265b8c95-de83-4478-a821-e8ea7af078be	INFO	start
2019-10-17T15:56:01.295Z	265b8c95-de83-4478-a821-e8ea7af078be	INFO	end
2019-10-17T15:56:01.296Z	265b8c95-de83-4478-a821-e8ea7af078be	INFO	0
2019-10-17T15:56:01.297Z	265b8c95-de83-4478-a821-e8ea7af078be	INFO	1
END RequestId: 265b8c95-de83-4478-a821-e8ea7af078be
REPORT RequestId: 265b8c95-de83-4478-a821-e8ea7af078be	Duration: 5068.85 ms	Billed Duration: 5100 ms	Memory Size: 128 MB	Max Memory Used: 24 MB	Init Duration: 121.97 ms	
```

However, if you remove the `next()` call in the middleware it works as expected.

Am I misusing this or is this an issue?

amwill04:
Possible due to this in the docs:

```
Promise

Middy allows you to return promises (or throw errors) from your handlers (instead of calling callback()) and middlewares (instead of calling next()).
```

Think this is a clear case to `RTFM` on my behalf. Can this be confirmed please?

amwill04:
Trying to use the example in the `docs`

```javascript
const asyncValidator = () => {
  before: async (handler) => {
    if (handler.event.body) {
      await asyncValidate(handler.event.body)

      return {foo: bar}
    }

    return
  }
}

handler.use(asyncValidator())
```

[...]

PR #448, opened 2019-12-11
Add in RDS Auth Token option to dbManager

### LINKED ISSUE #428: Middleware: RDS Auth Token
Recently built a middleware that automatically creates an IAM auth token to allow DB connection without a hardcoded password. Is this something we would like in the project? Or a separate module? Let me know, I can submit a PR with Docs.

Ways of setting the DB password, quick comparison:
- `process.env`: Password is in plain text
- `ssm`: Password encrypted, requires round trip to fetch, has to be rotated with an external process (additional cost)
- `secretManager`: Password encrypted, Rotates password, requires round trip to fetch, adds additional cost
- `authToken`: No password in the DB, based on IAM (15min access rotation), generated at the instance, no additional cost

Using the RDS AuthToken as you can see is more secure, free, and can be faster.

Docs:
- DB Setup: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html
- IAM Policy: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html
- RDS Signer: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDS/Signer.html
- Step by Step: https://aws.amazon.com/blogs/database/use-iam-authentication-to-connect-with-sql-workbenchj-to-amazon-aurora-mysql-or-amazon-rds-for-mysql/

```js
// example use
module.exports = (
  app,
  { dbUsername = 'iam_api', dbConcurrency = 0 }
) =>
  middy(app)
    .use(doNotWaitForEmptyEventLoop())
    .use(ssm({
      cache: true,
      setToContext: true,
      names: {
        rdsHostname: '/database/postgres/hostname',
        rdsDatabase: '/database/postgres/database',
        rdsPort: '/database/postgres/port'
      }
    }))
    .use(rdsAuthToken({
      username: dbUsername
    }))
```
Middleware:
```js
const AWS = require('aws-sdk')

const defaultNames = {
  hostname:'rdsHostname',
  port:'rdsPort',
  username:'rdsUsername',
  password:'rdsPassword'
}

let authToken = null
module.exports = (options = {}) => {
  const map = Object.assign(defaultNames, options.names)
  return {
    before: (handler, next) => {
      const chooseVar = (name) => {
        return handler.context[map[name]] || options[name] || process.env[name] || ''
      }

      handler.context[map['hostname']] = chooseVar('hostname')
      handler.context[map['username']] = chooseVar('username')
      handler.context[map['port']] = chooseVar('port')

      // warmup
      if (!authToken) {
        const signer = new AWS.RDS.Signer({
          region: chooseVar('region') || process.env.AWS_REGION,
          hostname: handler.context[map['hostname']],
          port: Number.parseInt(handler.context[map['port']]),
          username: handler.context[map['username']]
        })

        signer.getAuthToken({}, (err, token) => {
          if (err) {
            console.error(err)
            return
          }
          authToken = token
          handler.context[map['password']] = authToken
          next()
        })
      } else {
        handler.context[map['password']] = authToken
        next()
      }
    }
  }
}
```

willfarrell:
Closes #428

I did build it as a separate middleware, but the amount of excess code needed and managing the expiry didn't seem right.

Looking for some feedback/help:
- [x] `rdsSigner` option name, need to change?
- [x] A way to mock out `RDS.Signer`, currently failing due to missing creds
- [x] Should we add a connection retry in? Not at this time
- [x] One test having issue with `async`

cc @sdomagala

### DIFF
--- a/packages/db-manager/README.md
+++ b/packages/db-manager/README.md
@@ -61,7 +61,7 @@ npm install --save @middy/db-manager
 - `secretsPath` (optional): if for any reason you want to pass credentials using context, pass path to secrets laying in context object  - good example is combining this middleware with [ssm](#ssm)
 - `removeSecrets` (optional): By default is true. Works only in combination with `secretsPath`. Removes sensitive data from context once client is initialized.
 - `forceNewConnection` (optional): Creates new connection on every run and destroys it after. Database client needs to have `destroy` function in order to properly clean up connections.
-- `awsSdkOptions` (optional): Will use to create an IAM RDS Auth Token for the database connection using `[RDS.Signer](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDS/Signer.html)`. See AWs docs for required params, `region` is automatically pulled from the `hostname` unless overridden.
+- `rdsSigner` (optional): Will use to create an IAM RDS Auth Token for the database connection using `[RDS.Signer](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDS/Signer.html)`. See AWs docs for required params, `region` is automatically pulled from the `hostname` unless overridden.
 
 ## Sample usage
 
--- a/packages/db-manager/README.md
+++ b/packages/db-manager/README.md
@@ -128,7 +128,7 @@ handler.use(secretsManager({
     throwOnFailedCall: true
 }));
 handler.use(dbManager({
-  client: knex
+  client: knex,
   config: {
     client: 'pg',
     connection: {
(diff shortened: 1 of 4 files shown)


willfarrell:
@lmammino Updated the tests. They're all passing now. However, I've noticed that because I've added `async` to `before` in the middleware it's causing the tests to spit out extra "silent" errors from asynchronous triggering. I get the feeling `core` doesn't support `async`. Can you verify?

I can refactor to use promises, but will involve duplicating some code which I prefer not to do.

willfarrell:
After some testing, I've found that using `async/await` causes middy to trigger twice resulting in all the downstream middlewares also being triggered twice including the handler logic itself. I have it working without `async/await` locally but it's not as pretty looking. I'm going to push this change for v1. I've flagged this in my v2 wishlist.

willfarrell:
Just found https://github.com/middyjs/middy/issues/414, of course I can't call next and async return duh. Will test and re-update

willfarrell:
@lmammino was this closed deliberately?

lmammino:
Nope, sorry 😅

lmammino:
Apparently I can't re-open... can you, @willfarrell ?

lmammino:
Ok, I see now. I think because this was targeting `1.0.0-alpha` which was deleted. Can you retarget the new branch?

willfarrell:
Nope, I just created at new PR. #462

PR #451, opened 2019-12-19
[TypeScript] Ability to set your own context

sdomagala:
Hey @lmammino,
Right now even though some packages rely on context modification you can't change it and use TypeScript in the same time.
I have changed typings so now your custom context is visible in TypeScript, without this change your middleware functions need to have format `HandlerLambda<never, never, CustomContext>`.

Cheers

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,14 +4,14 @@ import {
   Handler
 } from 'aws-lambda';
 
-declare type EventType<T> =
-  T extends (event: infer EventArgType, context: Context, callback: Callback<any>) => void ? EventArgType :
-  T extends (event: infer EventArgType, context: Context) => Promise<any> ? EventArgType :
+declare type EventType<T, C> =
+  T extends (event: infer EventArgType, context: C, callback: Callback<any>) => void ? EventArgType :
+  T extends (event: infer EventArgType, context: C) => Promise<any> ? EventArgType :
   never;
 
-declare type HandlerReturnType<T> =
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -4,23 +4,23 @@ import {
   Handler
 } from 'aws-lambda';
 
-declare type EventType<T> =
-  T extends (event: infer EventArgType, context: Context, callback: Callback<any>) => void ? EventArgType :
-  T extends (event: infer EventArgType, context: Context) => Promise<any> ? EventArgType :
+declare type EventType<T, C> =
```
lmammino: From the height of my Typescript *noobness*: does it make sense here to say `C extends Context`?

And same in all the other instances below...

sdomagala: This type should be already inherited from function that started it, `const middy` in this case. In here this is just generic type that was defined before and is just passed along.
I've tested it on my code and if you don't define custom Context, the generic one is used

sdomagala:
Hey @lmammino, do you think we can have this merged or we're still waiting for other reviewers?

index.d.ts
vladgolubev: @sdomagala is right, line 21 already declares `C extends Context = Context`

lmammino:
Sorry @sdomagala, i did miss your clarification on my comment. Getting this merged shortly and thanks for the reminder :)

lmammino:
Also, do we have the same problem in v1.0.0?

PR #461, opened 2020-01-06
Fix syntax highlighting in Readme

pavelloz:


vladgolubev:
Just FYI to other maintainers, I merged this PR with a `[ci skip]` in the commit message so it won't trigger an unnecessary release flow in the CI

image

lmammino:
Thanks @vladgolubev! I didn't know you could do that :) that's great to know for future instances like this!

Thanks a lot also to @pavelloz for submitting the fix!

PR #462, opened 2020-01-06
Add in RDS Auth Token option to dbManager

willfarrell:


lmammino:
Sorry for the delay on reviewing this! Great work will. I will try to get this merged shortly :)

willfarrell:
@sdomagala Can you give this a quick review?

lmammino:
Ooops I hit merge already. I hope everything is ok @sdomagala, If not we will open a new PR to address any necessary change

sdomagala:
Hey guys, I'm back in Tuesday and I will take a look then :)

Cheers

packages/db-manager/index.js
```diff
@@ -33,18 +48,24 @@ module.exports = (opts) => {
       if (!config) {
         throw new Error('Config is required in dbManager')
       }
+
       if (!dbInstance || forceNewConnection) {
-        if (secretsPath) {
-          config.connection = Object.assign({}, config.connection || {}, handler.context[secretsPath])
+        const secrets = {}
+
+        if (options.rdsSigner && secretsPath) {
```
willfarrell: @sdomagala this is the section where I had a concern about a breaking change. In the original version it wasn't clear how you were doing the mapping. Is it safe to assume the password always set to `config.connection.password`? If so, I will update `secrets[secretsPath]` => `secrets['password']`. Otherwise, it feels like we need another param to map it to the right parameter. cc @lmammino

sdomagala: Hey @willfarrell,
This `secretsPath` can be whatever user decides it to be. If he want to store credentials (from example from parameter store or ssm) under `context['DB_CREDENTIALS']` then he has a freedom to do so and this path will be used
Also, under this path can be an object, not just plain password so `secrets['password']` wouldn't work as currently you can pass username, database host etc if you wish. I'm not sure how rds signer works in details, but I feel that your fix would break the contract  for the old versions as it only lets user pass password, nothing else.

Hope that makes sense 😄

willfarrell: `under this path can be an object, not just plain password`
That's what I thought the code was doing, I've never used Secrets Manager. Sounds like it returns an object, and Parameter Store (ssm) returns a string.

I'll update to cover these differences and add an example.

All RDS Signer does is create a token that you use as the password. That's it.

Issue #460, opened 2020-01-05
FunctionShield fails to run on nodejs v12


willfarrell:
CloudWatch Error
```
undefined	ERROR	Uncaught Exception 	
{
    "errorType": "Error",
    "errorMessage": "No native build was found for runtime=node abi=72 platform=linuxglibc arch=x64",
    "stack": [
        "Error: No native build was found for runtime=node abi=72 platform=linuxglibc arch=x64",
        "    at Function.load.path (/var/task/node_modules/node-gyp-build/index.js:59:9)",
        "    at load (/var/task/node_modules/node-gyp-build/index.js:19:30)",
        "    at Object.<anonymous> (/var/task/node_modules/@puresec/function-shield/dist/index.js:1:1078)",
        "    at c (/var/task/node_modules/@puresec/function-shield/dist/index.js:1:158)",
        "    at /var/task/node_modules/@puresec/function-shield/dist/index.js:1:962",
        "    at Object.<anonymous> (/var/task/node_modules/@puresec/function-shield/dist/index.js:1:971)",
        "    at Module._compile (internal/modules/cjs/loader.js:956:30)",
        "    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)",
        "    at Module.load (internal/modules/cjs/loader.js:812:32)",
        "    at Function.Module._load (internal/modules/cjs/loader.js:724:14)",
        "    at Module.require (internal/modules/cjs/loader.js:849:19)",
        "    at require (internal/modules/cjs/helpers.js:74:18)",
        "    at Object.<anonymous> (/var/task/node_modules/@middy/function-shield/index.js:1:24)",
        "    at Module._compile (internal/modules/cjs/loader.js:956:30)",
        "    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)",
        "    at Module.load (internal/modules/cjs/loader.js:812:32)"
    ]
}
```

Also noticed that redirect https://www.puresec.io/product to https://www.paloaltonetworks.com/prisma/containers-serverless now. `puresec` got bought out?

Their [npm](https://www.npmjs.com/package/@puresec/function-shield) package hasn't been updated in over 3m and has declining use.

willfarrell:
Did some digging, we need to update the version of our dep. I'll push a PR in a bit.

willfarrell:
Bumped the dep version up to v2, still failing.

```
2020-01-05T23:44:04.596Z	undefined	ERROR	Uncaught Exception 	
{
[...]

PR #464, opened 2020-01-06
FucntionShield: add warning and bump dep version

willfarrell:
Addresses #460

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0-beta"
+  "version": "1.0.0-beta.1"
 }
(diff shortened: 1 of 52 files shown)


willfarrell:
There are on ~250 download/week (https://www.npmjs.com/package/@middy/function-shield)

It might be easier to just keep it in till the number drops down to near zero then remove it from the future versions.

lmammino:
Yeah, I was thinking to keep it in the same npm package, but just move the repo outside the main one to start cleaning up. It shouldn't be too much work, I'll try to do it during the day.

This way we can avoid publishing new releases of this when there are new releases of middy and vice-versa.

What do you think?

willfarrell:
I feel putting it in its own repo would mean we have to maintain it longer by having to keep the repo in the org.

lmammino:
Fair enough. I will get this merged soon :)

Issue #467, opened 2020-01-24
Lambda SQS Batch Error Handler Middleware


brettstack:
Hi team, I created a middleware https://github.com/brettstack/civ6-play-by-cloud-turn-notifier/tree/master/packages/lambda-batch-sqs-error-cleaner-middleware to capture the pattern discussed here https://twitter.com/AWSbrett/status/1213542155873947648.

With the Lambda SQS Trigger, either all messages in the batch get deleted from the queue (when the Lambda resolves successfully) or none get deleted and remain on the queue to be processed again (when the Lambda throws an error).

In the case where at least 1 message wasn't successfully processed, this middleware handles deleting the successfully processed messages for you, then throws an error to return the remaining messages to the queue. If all messages are processed successfully, the middleware doesn't delete the messages itself, but simply resolves and lets the Lambda service handle deleting the messages (outside of your function invocation).

The middleware determines successfully processed messages and failures based on the handler returning a `Promise.allSettled` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled result.

```js
const recordPromises = Records.map(async (record, index) => { /* Custom message processing logic */ })
const settledRecords = await Promise.allSettled(recordPromises)
return settledRecords
```

I'd appreciate any feedback you have. Let me know if you're interested in adding this to the core middleware in this repo.

lmammino:
Hello @brettstack, this looks great! Thanks for taking the time to formalize this here.

I am supportive for having this integrate into the core set of middlewares (for middy 1.0.0) , but i'd be curious to know how common this use case can be. If you believe that there will be a significative user base for it, than it makes sense to do extra due diligence to have this package always in sync and tested together with the rest of the core middlewares.

If not, it would still be good to have this well structured, documented and available on NPM so that we can add it to the list of third-party middlewares.

@willfarrell, what do you think?

brettstack:
This is the only data I have on how common this: https://stackoverflow.com/questions/54596129/processing-aws-lambda-messages-in-batches/54611899#54611899, which isn't a very popular SO question.

Partially failed batches is something everyone using SQS Batch processing will/should have to deal with. However, since users are already advised to check if a particular message has already been processed  (since SQS is at-least-once delivery), this issue is sometimes masked (especially if the developer is manually deleting the message at that point).

willfarrell:
[...]

PR #471, opened 2020-01-29
feat: add sqs-batch middleware

brettstack:
Related issue: #467

TODO:
- [x] Resolve dangling underscore and parameterized function overrides discussion


willfarrell:
Looks good.

Are there any required settings that need to be set on the SQS for this to work as expected (ie `BatchSize`)? If so we should document.

brettstack:
Excellent question. Batch size is 10 by default for SQS.

brettstack:
Tests are failing in node 10 due to `Promise.allSettled` not being available. How do we skip for this version?

willfarrell:
We can’t skip that version till AWS does. Please add in a poly fill for now.

willfarrell:
Thanks for the great work. 

@lmammino  `sqs` or `sqs-batch`?

brettstack:
Added the polyfill to the tests. To be clear: the middleware itself doesn't use `allSettled`, but the user needs to return a `Promise.allSettled()` value from their Lambda handler. So anyone wanting to use this middleware in Node < 12 will need to add a polyfill. I'll mention this in the docs.

brettstack:
`sqs-partial-batch-failure` would be the most descriptive. Or even:

`sqs-partial-batch-failure-cleaner`
`sqs-partial-batch-failure-handler`
`sqs-graceful-partial-batch-failure`

packages/sqs-batch/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "aws-sdk": ">=2.221.1"
+  },
+  "devDependencies": {
+    "@middy/core": "^1.0.0-beta",
+    "@serverless/event-mocks": "^1.1.1",
+    "promise.allsettled": "^1.0.2"
+  },
+  "gitHead": "7a6c0fbb8ab71d6a2171e678697de9f237568431",
+  "dependencies": {
+    "jest": "^25.1.0"
```
vladgolubev: Is this supposed to be a dependency?

willfarrell:
@lmammino Was just thinking. Should we add a section to the middleware readmes that lets people know who to flag when they have a question or find an issue with it?

packages/sqs-batch/package.json
willfarrell: `devDependencies` makes more sense to me, I'm guessing it was copied from somewhere.

brettstack: Absolutely not. Good catch.

packages/sqs-batch/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@middy/sqs-batch",
+  "version": "1.0.0-beta",
+  "description": "SQs batch error handler middleware for the middy framework",
```
vladgolubev:
```suggestion
  "description": "SQS batch error handler middleware for the middy framework",
```

brettstack: Ugh. The shift key on my shitty MacBook butterfly keyboard is going out. I'll fix this evening.

lmammino:
Sorry for the delay on this! Busy days... I should have some time this afternoon to have a proper look and get back to you with some feedback!

lmammino:
> Thanks for the great work.
>
> @lmammino sqs or sqs-batch?

I like `sqs-batch` between the 2. But between those ones:

> Some other options:
>
> sqs-partial-batch-failure-handler ❌handler is ambiguous
> sqs-partial-batch-failure-cleaner ✅cleaner is descriptive
> sqs-graceful-partial-batch-failure 👈I like graceful
> sqs-partial-batch-failure 🩳shortest

I like `sqs-partial-batch-failure` :)

> @lmammino Was just thinking. Should we add a section to the middleware readmes that lets people know who to flag when they have a question or find an issue with it?

We could, let's maybe create a dedicate issue / PR for this. What do you think?

packages/sqs-batch/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+## Sample usage
+
+```javascript
+const middy = require('@middy/core')
+const sqsBatch = require('@middy/sqs-batch')
+
+const originalHandler = (event, context, cb) => {
+  const recordPromises = Records.map(async (record, index) => { /* Custom message processing logic */ })
```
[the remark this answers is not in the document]
brettstack: Good catch. I meant to have `event.Records` 🙃

packages/sqs-batch/index.js
```diff
@@ -0,0 +1,91 @@
+/* eslint no-underscore-dangle:0 */
```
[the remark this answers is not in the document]
brettstack: The non-dangled version is defined as a parameter on the exported function to allow customers to override default functionality. I can remove this if you prefer, or rename those with some other prefix or suffix.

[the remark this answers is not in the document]
lmammino: Thanks for being available on changing this. I think in other middleware we just call that `defaultX`. Maybe this kind of prefix would be a bit more explicit. What do you think?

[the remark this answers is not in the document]
brettstack: Done!

brettstack:
I think we're good to go

lmammino:
Fantastic, thanks for your patience @brettstack! I should be able to have a release in the next few hours!

PR #477, opened 2020-02-10
Fix: Http-response-serializer returns {} when body is falsey

roni-frantchi:
When response is falsey (boolean `false` or the number `0`), `http-response-serializer` thinks it is `null` or `undefined`, thus creating an empty object `{}`.  

Root cause is the check doesn't actually test for `null` or `undefined` properly.

### DIFF
--- a/packages/http-response-serializer/__tests__/index.js
+++ b/packages/http-response-serializer/__tests__/index.js
@@ -252,4 +252,26 @@ describe('📦  Middleware Http Response Serializer', () => {
       body: '{}'
     })
   })
+
+  test('It should return false when response body is falsey', async () => {
+    const handler = middy((event, context, cb) => {
+      cb(null, { statusCode: 200, body: false })
+    })
+
+    const event = {
+      headers: {
+        Accept: 'text/plain'
+      }
+    }
+    handler.use(httpResponseSerializer(standardConfiguration))
+    const response = await invoke(handler, event)
+
+    expect(response).toEqual({
+      statusCode: 200,
+      headers: {
+        'Content-Type': 'text/plain'
+      },
+      body: false
+    })
+  })
 })
--- a/packages/http-response-serializer/__tests__/index.js
+++ b/packages/http-response-serializer/__tests__/index.js
@@ -273,24 +273,4 @@ describe('📦  Middleware Http Response Serializer', () => {
       body: false
     })
   })
-  test('It should return false when response body is falsey', async () => {
-    const handler = middy((event, context, cb) => {
-      cb(null, false)
-    })
-
-    const event = {
-      headers: {
-        Accept: 'text/plain'
-      }
-    }
-    handler.use(httpResponseSerializer(standardConfiguration))
-    const response = await invoke(handler, event)
-
-    expect(response).toEqual({
-      headers: {
-        'Content-Type': 'text/plain'
-      },
-      body: false
-    })
-  })
 })
--- a/packages/http-response-serializer/__tests__/index.js
+++ b/packages/http-response-serializer/__tests__/index.js
@@ -255,7 +255,7 @@ describe('📦  Middleware Http Response Serializer', () => {
 
   test('It should return false when response body is falsey', async () => {
     const handler = middy((event, context, cb) => {
-      cb(null, { statusCode: 200, body: false })
+      cb(null, false)
     })
 
     const event = {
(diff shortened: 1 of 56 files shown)


lmammino:
Awesome contribution @roni-frantchi! I'll try to get this merged and released shortly!

packages/http-response-serializer/index.js
```diff
@@ -43,7 +43,7 @@ const middleware = (opts, handler, next) => {
     if (!test) { return false }
 
     // if the response is null or undefined, normalizes it back to an object
-    handler.response = handler.response || {}
+    handler.response = handler.response != null ? handler.response : {}
```
lmammino: Shouldn't this be `!==` or am I missing some edge case here?

lmammino:
Actually I left a comment

packages/http-response-serializer/index.js
roni-frantchi: @lmammino `!=` is safer as `handler.response !== null` will not catch cases where `handler.response` is `undefined`

lmammino: Shall we be a bit more explicit then and define this as follows?

```javascript
handler.response = (handler.response !== null && typeof handler.response !== 'undefined')  ? handler.response : {}
```

What do you think?

roni-frantchi: I think it's pretty [common practice](https://stackoverflow.com/questions/2647867/how-to-determine-if-variable-is-undefined-or-null) and I find the alternative a little too verbose, but your call :)

roni-frantchi: See here @lmammino  https://github.com/middyjs/middy/pull/477/commits/c15631bf58de417fb2f81de3125d918cfea181b3

lmammino: Awesome, thanks for your patience. I'll version bump and merge in the next few hours!

packages/http-response-serializer/__tests__/index.js
```diff
@@ -252,4 +252,26 @@ describe('📦  Middleware Http Response Serializer', () => {
       body: '{}'
     })
   })
+
+  test('It should return false when response body is falsey', async () => {
+    const handler = middy((event, context, cb) => {
+      cb(null, { statusCode: 200, body: false })
```
dkatavic: If I got it right, this test case is not testing introduced changes in the detection of the type of response.  Response should be 
```
cb(null, false)
// or
cb(null, 0)
```
to test your change?

lmammino:
> Just note for future readers that this will be a potential breaking change.

@dkatavic, good point. What would you recommend here? I was thinking to add a `CHANGELOG.md` file in the folder to document the new behaviour.

dkatavic:
@lmammino `CHANGELOG.md` for breaking changes sounds great for me.

Question regarding the behavior of this middleware, in generic case of the response being any primitive data type (including `0` or `false`), code will throw an error on the next line when we try to set property `.headers` on `response`. Is this desired behavior?

packages/http-response-serializer/__tests__/index.js
roni-frantchi: This middleware will serialize the `body` - the issue occurred when `body` is `false`y

dkatavic: Maybe I got this wrong, but this test would pass with a previous implementation too, right?

roni-frantchi: I'll have a look @dkatavic

roni-frantchi:
@lmammino @dkatavic see the latest changes please.

lmammino:
I think this last approach makes more sense. @dkatavic if you are happy as well, I'll proceed with getting this merged and released.

lmammino:
hey @roni-frantchi, it seems I can't push changes to this branch, can you check if you have the permission disabled on your side?

roni-frantchi:
@lmammino I've given you write access to my fork

lmammino:
Thanks @roni-frantchi, I still have problems to push. Can you apply this patch yourself?

[patch.diff.zip](https://github.com/middyjs/middy/files/4215245/patch.diff.zip)

lmammino:
wait @roni-frantchi, i did manage to push. I accepted the invitation on the email.

for future reference, there should be a lighter way to give access only to the branch that is referenced by the pr when you submite the PR (and somewhere in the PR's settings)

squidfunk:
Just a note on an error into which I stumbled: The response is initialized with `null`. The JSON request serializer checks for `typeof handler.response === "object"`, which evaluates to `true`, as `null` is an `object`. This will lead to an error in the next statement when trying to set the `headers` on the `response`:

```
(node:7381) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'headers' of null
```

willfarrell:
@squidfunk Would you like to put together a PR to fix this?

squidfunk:
@willfarrell I'm not sure what you intended with setting `response` to `null`, so my take on this would be setting it to `undefined`. After all, `undefined` is safer than `null` for the reasons stated, as it is also "nullish", but not considered an object.

I will prepare an issue with a minimal reproducible case, so we can take it from there.

PR #491, opened 2020-03-06
[feature] sqs json body parser

thematrimix:
It occured to me that we have a http json body parser. Why not have a sqs json body parser since the Record.body will come in as a string. The only tricky part is there is no content type that tell us if the body is a json string. However, this should be okay as I suspect most of the time the dev will know if he/she will be dealing with json. It can also be mitigated by a simple try/catch or the user providing there own parser. So while this say it is a json body parser it could be used to parser whatever needed say xml by passing a proper safeParse.


packages/sqs-json-body-parser/.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "node": "6.10"
```
willfarrell: update to node 10

thematrimix: While this has merrit I have reservation of updating until core is updated.

thematrimix: Is there any plan to make the change across the board to target node 10?

Issue #485, opened 2020-02-24
Error that gets handled by middleware is still passed on to next moddleware


jacsamell:
When you have a chain of middlewares that implement onError, when the first middleware handles the error and calls next() without passing in the error as a parameter the next one should not be invoked according to the documentation.

However all onErrors in the chain are called regardless of whether the error has been passed into the next function.

philprime:
I noticed the same issue

lmammino:
Hello @jacsamell and @philprime, thanks a lot for reporting this.

This seems indeed a bug on how the error propagation is handled when using promises (or async-await).

Unfortunately, I currently have very little time to investigate this, but if you have some time to invest on this, [this part of the code]https://github.com/middyjs/middy/blob/1.0.0-beta/packages/core/index.js#L102-L139() might be a good starting point.

Sorry if I can't be more helpful at this time.

philprime:
Hi @lmammino, I just took a look at the code and if I change line 113 to the following, it does not execute the remaining error middlewares anymore:

https://github.com/middyjs/middy/blob/a64218d2199d580b70253c0c78d7541e31a09ee6/packages/core/index.js#L113

```javascript
if (nextMiddleware && !instance.__handledError) {
```

Now I expected it to call the remaining `onAfter` middlewares outside of the one that handled the error, but I couldn't find anything to that topic.

If I extend the sample code of your documentation by another `onAfter` handler in `middleware2`, should that be called?

```javascript
middleware1 = {
  onError: (handler) => {
    handler.response = { error: handler.error };
    return Promise.resolve();
    // Resolves to a falsy value
  }
}
middleware2 = {
  onError: (handler) => {
    return Promise.resolve(handler.error)
  },
  onAfter: (handler) => {
    // Should this be called?
  }
}
handler.use(middleware1).use(middleware2);
```

lmammino:
Thanks a lot for spending some time on this issue @philprime! It might be worth to revisit the test suite and maybe submit a PR if you think that change is enough to stop the bug and does not have side effects on other behaviours.

Regarding your question, `middleware2`'s `onAfter` should not be invoked by design. The idea is that once you are on the `onError` track you derailed outside the normal request/response flow and only `onError` handlers will be invoked (in order) giving them a chance to handle the error as needed.

If you want to have an additional step of error handling you should implement a `onError` also in `middleware2` and keep propagating the error from `middleware1` by doing `return Promise.reject(handler.error)`.

philprime:
[...]

PR #497, opened 2020-03-28
Changed error handling logic to stop after error has been handled

philprime:
Changed run error middlewares so that it conforms to the behaviour stated in the documentation

```javascript
middleware1 = {
  onError: (handler) => {
    handler.response = { error: handler.error };
    return Promise.resolve();
    // Resolves to a falsy value
  }
}
middleware2 = {
  onError: (handler) => {
    return Promise.resolve(handler.error)
  }
}
handler.use(middleware1).use(middleware2);
```

`middleware2` should not be called.

# Relevant issues:

- #485

### DIFF
--- a/src/__tests__/middy.js
+++ b/src/__tests__/middy.js
@@ -276,7 +276,7 @@ describe('🛵  Middy test suite', () => {
     })
   })
 
-  test('If theres an error and one error middleware handles the error, the next error middlewares is executed', (endTest) => {
+  test('If theres is an error and one error middleware handles the error, the next error middlewares should not be executed', (endTest) => {
     const expectedResponse = { message: 'error handled' }
 
     const onErrorMiddleware1 = jest.fn((handler, next) => {
@@ -297,7 +297,7 @@ describe('🛵  Middy test suite', () => {
     handler({}, {}, (err, response) => {
       expect(err).toBeNull()
       expect(onErrorMiddleware1).toBeCalled()
-      expect(onErrorMiddleware2).toBeCalled()
+      expect(onErrorMiddleware2).not.toHaveBeenCalled()
       expect(response).toBe(expectedResponse)
 
       endTest()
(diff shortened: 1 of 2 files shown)


lmammino:
Hello @philprime!
Thanks a lot for taking the time to address this and apologies again for the delay with my review.

At first glance, this looks good and more consistent with what is documented [here](https://github.com/middyjs/middy/tree/1.0.0-beta#handling-errors).

I am ok for merging this, but I prefer to do it after #504 so that I can release it together with few other pending PRs.

Meanwhile, I'd love to get @willfarrell or @dkatavic review as well here :)

src/__tests__/middy.js
```diff
@@ -276,7 +276,7 @@ describe('🛵  Middy test suite', () => {
     })
   })
 
-  test('If theres an error and one error middleware handles the error, the next error middlewares is executed', (endTest) => {
+  test('If theres is an error and one error middleware handles the error, the next error middlewares should not be executed', (endTest) => {
```
dkatavic, line 279: Should there be `there` instead of `theres`?

dkatavic: This was merged while inspecting the code, ignore it

lmammino:
We should port this one to 1.0.0-beta, realized only now this was targeting master :(

src/__tests__/middy.js
lmammino: I did mess up with this one a little. I will do another commit to properly release this on 0.x and also fix this typo. So totally valuable comment :) Thanks

lmammino: (i will also fix in the port to 1.0)

lmammino:
This fix will be released shortly in 0.35.0 and in 1.0.0-beta.11

lmammino:
I am revisiting this in the context of v1 (which includes a http-response-serializer that also works with the default HTTP error handler).

This PR breaks the original intended design (and therefore compatibility with some middlewares).

As per the docs:

> When a middleware handles the error and creates a response, **the execution is still propagated to all the other error middlewares** and they have a chance to update or replace the response as needed. At the end of the error middlewares sequence, the response is returned to the user.

This is needed for instance in the case of the HTTP error handler which will format the error in a correct response for the lambda HTTP integration. If this is combined with the response serializer it must be possible to convert this response to XML or other formats.

I will have to revert this on master

PR #502, opened 2020-04-13
Add failing tests for serializer search

jarrodldavis:
I am opening up this pull request early as a draft to get feedback on the issue and tests I've added before working on a fix.

I noticed that `@middy/http-response-serializer` doesn't appear to find serializers accurately when the first mime-type doesn't match.

For example:
- if the default is `application/json` and the `accept` header's value is `*/*`
- regardless of default, an `application/xml` serializer is present, and the `accept` header's value is `text/x-dvi, application/xml`

Instead of matching the serializer, the middleware simple acts as a pass-through and doesn't perform any serialization.

No.

No.

Any other comments?
-------------------
No.

Where has this been tested?
---------------------------
**Node.js Versions:** v12.16.2

**Middy Versions:** latest on branch `1.0.0-beta`

**AWS SDK Versions:** latest included in meta package (`^2.635.0`)

Todo list
---------

[x] Added tests
[ ] Updated relevant examples


lmammino:
Thanks for this one @jarrodldavis. Did you check if the new version of `accept` (`@hapi/accept`) solves this issue? If not, any clue what else can affect this?

jarrodldavis:
I didn't test with the updated `@hapi/accept`, but stepping through this module from its usage with a (private) project I'm working on, it appears the logic here is existing after testing the first type:

https://github.com/middyjs/middy/blob/63a5ae746b48c7a71dd82e41cacbb0e28fc87270/packages/http-response-serializer/index.js#L38-L44

From what I can tell, this is because `Array#find` will early exit out of the loop when it receives a truthy value, and an array is always truthy even when empty or it contains all falsy values.

lmammino:
Your analysis makes a lot of sense. Thanks for taking the time. Would you have some extra time to update this PR with a fix?

jarrodldavis:
Sure thing. It's probably going to be a day or two as I have some things I'm working on currently.

willfarrell:
@jarrodldavis Any update on this? Would you be able to rebase with the current `master` branch?

jarrodldavis:
I'm sorry but I'm really busy with work right now so I don't have time to get to this.

willfarrell:
All good, thanks for the update.

willfarrell:
These tests have been added in to v2-alpha and are passing. Thanks

PR #507, opened 2020-04-16
Ability to exclude paths in input-output-logger

roni-frantchi:
Adds the ability to exclude paths from input/output logger.  
Very useful to reduce log verbosity and scrub-off secrets or sensitive data.  

Where has this been tested?
---------------------------
**Node.js Versions:** v12.16.1

**Middy Versions:** 1.0.0-beta

**AWS SDK Versions:** 2.556.0

Todo list
---------

- [x] Added tests

### DIFF
--- a/packages/input-output-logger/README.md
+++ b/packages/input-output-logger/README.md
@@ -45,7 +45,7 @@ npm install --save @middy/input-output-logger
 ## Options
 
 `logger` property accept a function (default `console.log`)
-`exclude` property accept an array of paths to exclude from the log (default `[]`)
+`omitPaths` property accept an array of paths to omitPaths from the log (default `[]`)
 
 
 ## Sample usage
--- a/packages/input-output-logger/README.md
+++ b/packages/input-output-logger/README.md
@@ -45,6 +45,7 @@ npm install --save @middy/input-output-logger
 ## Options
 
 `logger` property accept a function (default `console.log`)
+`exclude` property accept an array of paths to exclude from the log (default `[]`)
 
 
 ## Sample usage
--- a/packages/input-output-logger/README.md
+++ b/packages/input-output-logger/README.md
@@ -45,7 +45,7 @@ npm install --save @middy/input-output-logger
 ## Options
 
 `logger` property accept a function (default `console.log`)
-`omitPaths` property accept an array of paths to omitPaths from the log (default `[]`)
+`omitPaths` property accepts an array of paths that will be used to remove particular fields from the logged objects. This could serve as a simple way to redact sensitive data from logs (default []).
 
 
 ## Sample usage
--- a/packages/input-output-logger/__tests__/index.js
+++ b/packages/input-output-logger/__tests__/index.js
@@ -18,19 +18,36 @@ describe('📦 Middleware Input Output Logger', () => {
     expect(logger).toHaveBeenCalledWith({ event: { foo: 'bar', fuu: 'baz' } })
     expect(logger).toHaveBeenCalledWith({ response: { message: 'hello world' } })
   })
-  test('It should omit paths', async () => {
-    const logger = jest.fn()
+  describe('omitPaths', () => {
+    test('It should omit paths', async () => {
+      const logger = jest.fn()
 
-    const handler = middy((event, context, cb) => {
-      cb(null, { message: 'hello world', bar: 'bi' })
+      const handler = middy((event, context, cb) => {
+        cb(null, { message: 'hello world', bar: 'bi' })
+      })
+
+      handler
+        .use(inputOutputLogger({ logger, omitPaths: ['event.foo', 'response.bar'] }))
+
+      await invoke(handler, { foo: 'bar', fuu: 'baz' })
+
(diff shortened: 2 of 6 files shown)


packages/input-output-logger/index.js
```diff
@@ -1,21 +1,29 @@
+const omit = require('lodash/omit')
+
 module.exports = (opts) => {
   const defaults = {
-    logger: data => console.log(JSON.stringify(data, null, 2))
+    logger: data => console.log(JSON.stringify(data, null, 2)),
+    exclude: []
```
lmammino, line 6: Maybe a little bit more explicit to call this something like `omitPaths`?

packages/input-output-logger/README.md
```diff
@@ -45,6 +45,7 @@ npm install --save @middy/input-output-logger
 ## Options
 
 `logger` property accept a function (default `console.log`)
+`exclude` property accept an array of paths to exclude from the log (default `[]`)
```
lmammino, line 48: I initially had a feeling that this would exclude the entire log message if an "exclude" path would match it.

Going through the code it seems that this works a bit more as a "field redaction" mechanism (e.g. to remove passwords from logs).

If that's the case, maybe we can be a bit more explicit here with something like this:

>  `exclude` property accepts an array of paths that will be used to remove particular fields from the logged objects. This could serve as a simple way to redact sensitive data from logs (default `[]`).

I was also suggesting that changing `exclude` to something more explicit can help to clarify the scope here. What do you think?

packages/input-output-logger/__tests__/index.js
```diff
@@ -18,4 +18,19 @@ describe('📦 Middleware Input Output Logger', () => {
     expect(logger).toHaveBeenCalledWith({ event: { foo: 'bar', fuu: 'baz' } })
     expect(logger).toHaveBeenCalledWith({ response: { message: 'hello world' } })
   })
+  test('It should exclude paths', async () => {
+    const logger = jest.fn()
+
+    const handler = middy((event, context, cb) => {
+      cb(null, { message: 'hello world', bar: 'bi' })
```
lmammino, line 25: what happens if `response` is not an object (e.g. a string). Should we add an extra test to check for these edge cases?

packages/input-output-logger/index.js
```diff
@@ -1,21 +1,29 @@
...
   }
 
-  const options = Object.assign({}, defaults, opts)
+  const { logger, exclude } = Object.assign({}, defaults, opts)
+
+  const omitAndLog = data => {
+    const message = omit(data, exclude)
```
lmammino, line 12: maybe for clarity let's rename this to `redactedMessage`

packages/input-output-logger/package.json
```diff
@@ -44,10 +44,12 @@
     "aws-sdk": ">=2.221.1"
   },
   "dependencies": {
-    "@types/node": "^10.0.8"
+    "lodash": "^4.17.15"
   },
   "devDependencies": {
     "@middy/core": "^1.0.0-beta.10",
+    "@types/node": "^10.0.8",
```
lmammino, line 51: good catch here!

packages/input-output-logger/index.js
roni-frantchi: Done [here](https://github.com/middyjs/middy/pull/507/commits/cd02db468a56d2184577eaeae40e5fadb0468508)

packages/input-output-logger/index.js
roni-frantchi: Done [here](https://github.com/middyjs/middy/pull/507/commits/cd02db468a56d2184577eaeae40e5fadb0468508)

packages/input-output-logger/README.md
roni-frantchi: Done [here](https://github.com/middyjs/middy/pull/507/commits/b213c25704ce52b3958a9ca8c23a389394952bdc)

packages/input-output-logger/__tests__/index.js
roni-frantchi: Done [here](https://github.com/middyjs/middy/pull/507/commits/d3b0fc04503b529ee1929ef6cb2e5bcf58948a7c)

packages/input-output-logger/index.js
lmammino: Tnx

packages/input-output-logger/__tests__/index.js
lmammino: Tnx

packages/input-output-logger/README.md
lmammino: Awesome, thanks!

packages/input-output-logger/index.js
lmammino: ;)

lmammino:
Releasing this in [`1.0.0-beta.11`](https://github.com/middyjs/middy/releases/tag/1.0.0-beta.11)

roni-frantchi:
Thanks @lmammino! ❤️

PR #509, opened 2020-04-20
Fix import middleware name

abdala:


lmammino:
Great catch, thanks :)

PR #539, opened 2020-06-26
docs(secrets-manager): fix Options

munierujp:
see https://github.com/middyjs/middy/blob/040ad16d0cbb8c41225a609fa8aeccd2a20d9436/packages/secrets-manager/index.js#L60-L63


willfarrell:
Good catch

PR #544, opened 2020-07-16
The `httpResponseSerializer` needs to find a serializer for a `type` by searching all the `types`

randytarampi:

And all the `opts.serializers` too.

[`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) short circuits on truthy return values of its callback, which means that we only would've ever traversed the array for the first element of `types` since `opts.serializers.map` just returns an array of booleans right?

Fix this typo and add a test to cover my use case of wanting to fall back to a default if no preferred serializer is found.

Nothing that I could find from my quick searching. I just needed this for myself right now.

N/A

Any other comments?
-------------------

None.

Where has this been tested?
---------------------------
**Node.js Versions:** 12

**Middy Versions:** 1

**AWS SDK Versions:** N/A

Todo list
---------

- [x] Added tests

### DIFF
--- a/packages/http-response-serializer/__tests__/index.js
+++ b/packages/http-response-serializer/__tests__/index.js
@@ -119,6 +119,32 @@ describe('📦  Middleware Http Response Serializer', () => {
     })
   })
 
+  test('It should use the default when no matching accept preferences are found', async () => {
+    const handler = middy((event, context, cb) => {
+      event.preferredContentType = 'text/java'
+
+      cb(null, createHttpResponse())
+    })
+
+    handler.use(httpResponseSerializer(standardConfiguration))
+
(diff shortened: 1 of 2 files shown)


randytarampi:
Hey @willfarrell, 

Got an idea of when this'll go out? I'm about to release something into production and don't really want to have to build & cut my own `@randytarampi/http-response-serializer` if the official fix is coming soon.

willfarrell:
Sorry about that. Been working around the clock these days. I'll make time this week to put a release together. expect a v1.1.0

randytarampi:
Thank you very much! 🙏

PR #552, opened 2020-08-15
Validator accepting plugins

### LINKED ISSUE #550: Ability to apply plugins to middy/validator
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

N/A

What happens?
-------------
In order to better customize middy/validator to show up custom errors, I need to apply this plugin: https://github.com/ajv-validator/ajv-errors

Ability apply plugins to validator ajv instance.

Any logs, error output, etc?
----------------------------
N/A

Any other comments?
-------------------
Thanks!

N/A

leog:

Implements the ability to apply passed plugins to validator

Closes #550 

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Node.js Versions:** 12.16.1

**Middy Versions:** 1.2.0

**AWS SDK Versions:** 2.656.0

Todo list
---------

[x] Added tests
[x] Updated relevant documentation
[x] Updated relevant examples

### DIFF
--- a/packages/validator/README.md
+++ b/packages/validator/README.md
@@ -41,7 +41,7 @@ response to the user.
 
 It can also be used in combination with [`httpcontentnegotiation`](#httpContentNegotiation) to load localised translations for the error messages (based on the currently requested language). This feature uses internally [`ajv-i18n`](http://npm.im/ajv-i18n) module, so reference to this module for options and more advanced use cases. By default the language used will be English (`en`), but you can redefine the default language by passing it in the `ajvOptions` options with the key `defaultLanguage` and specifying as value one of the [supported locales](https://www.npmjs.com/package/ajv-i18n#supported-locales).
 
-Also, this middleware accepts an array of plugins to be applied, such as `ajv-errors` to customize the internal `ajv` instance.
+Also, this middleware accepts an object with plugins to be applied to customize the internal `ajv` instance. Out-of-the-box, `ajv-i18n`, `ajv-errors` and `ajv-keywords` are being used.
 
 ## Install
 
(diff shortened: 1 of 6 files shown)


willfarrell:
Interesting approach, I was thinking something different. I like this better, cleaner. 

Thinking out loud here:  Maybe `avjKeywords` should be refactored into this pattern (have it as a default until v2)? Both `errors` and `keywords` can take options themselves. Seeing how `ajv-errors` is really the only plugin missing, what if we included it by default and let the user pass in an object of the plugins w/ options they want to use. Since we don't want to introduce a breaking change it would have to default to `{"keywords":{}}`.

What do you think?

leog:
Thanks for the feedback. And yes, makes sense. 

So to task it out if I understood correctly:
- [ ] Let the user pass plugins using the pattern `{"keywords":{}} enabling plugins options to be passed as well
- [ ] Use plugins implementation to include out-of-the-box `ajv-keywords` and `ajv-errors`

In case the user prefers to apply one of the out-of-the-box plugins for further customization I guess we should prioritize that, which we could as we would be defaulting the plugins similar to options 👍

leog:
Well, I think we might need to leave `ajv-errors` out after all, as it requires a possible breaking change due to `jsonPointers` ajv option having to be true, otherwise the plugin sets it automatically giving a warning. Thoughts?

BTW, got `i18n` into the new plugins system as well, PTAL.

leog:
Please let me know if this is OK @willfarrell.

renanwilliam:
This change broke a running project. I'm receiving this error:

- `Error: Cannot find module 'ajv-keywords'`

Is necessary any change or manual library import?

willfarrell:
`ajv-keywords` is included as a dep in the submodule. I'll do some additional testing today.

leog:
Let me know if I can help @willfarrell. It may be related to the dynamic require, can't see why at the moment

willfarrell:
@renanwilliam I was able to reproduce your code issue. Are you using `serverless` with `serverless-bundle`? `ajv-keywords` and `ajv-i18n` were moved to be dynamically required, which seems to be the cause. Debugging and testing out solutions now.

renanwilliam:
@willfarrell just additional information, I have added manually as a dependency but keeps throwing the error. Rolling back to ~1.2.0 solves the problem.

renanwilliam:
> @renanwilliam I was able to reproduce your code issue. Are you using `serverless` with `serverless-bundle`? `ajv-keywords` and `ajv-i18n` were moved to be dynamically required, which seems to be the cause. Debugging and testing out solutions now.

I'm using with `serverless-bundle`

willfarrell:
Did some testing, and was unsuccessful in finding a solution. Updating the bundle config should have done it (see below). 

serverless.yml
```
custom:
  bundle:
    sourcemaps: false
    caching: false
    linting: false
    forceInclude:
      - pg
      - ajv-keywords
      - ajv-i18n
      - ajv-errors
```
I'm not a webpack expert, so can't take it any furthur at this time. @renanwilliam if you can open an issue at https://github.com/AnomalyInnovations/serverless-bundle/issues/new, I'd be interested in what the solution is and how we can update our docs to prevent this issue for others. cc @AnomalyInnovations

What I can do is update the module to allow the ability to remove all default plugins by passing in `ajvOptions:{}`. Keep a look out for `v1.3.1`.

leog:
@willfarrell if you could share a reproducible project to test solutions I can give it a try, have a couple of ideas. Thanks @renanwilliam for reporting this and being patient.

willfarrell:
I was testing on a private repo. Any serverless project w/ serverless-bundle should invoke the issue.

ayushsharma82:
A 'plugin' array would have been more useful considering `ajv-keywords` got removed in v2 release. 

Any way to add the `ajv-keywords` plugin back in v2?

willfarrell:
@ayushsharma82 Please see my response on https://github.com/middyjs/middy/issues/698

Issue #558, opened 2020-08-29
Middy stores the request object on the global handler instance


thetrevdev:
What are the steps to reproduce this issue?
-------------------------------------------
This code shows the issue:
Calling a handler twice concurrently will clobber the request instance.

https://codesandbox.io/s/purple-star-n3ohc?file=/src/index.js

What happens?
-------------
The two requests clobber the instance object replacing both the context and responses.

What were you expecting to happen?
----------------------------------
They are scoped and resolve independently

Any logs, error output, etc?
----------------------------
See example above.

What versions of software are you using?
----------------------------------------
**Node.js Version:** 10/12

**Middy Version:** 1.2.0

**AWS SDK Version:** 2x

PR #559, opened 2020-08-29
Scope each request independently

thetrevdev:
Fixes -> https://codesandbox.io/s/purple-star-n3ohc?file=/src/index.js

This scopes the request object to an individual request.  This allows for calling a wrapped handler multiple times concurrently.  This is useful when trying to reuse individual handlers when batching.  Without this a handler may receive the wrong request or return the incorrect response when calling concurrently.

https://github.com/middyjs/middy/issues/558

Any other comments?
-------------------

This could be considered a breaking change if someone was mutating the handler object intending it to persist between requests.  
That being said there would be no reason to store your global state on the handler itself unless you thought it was scoped per request which this implements.

Where has this been tested?
---------------------------
**Node.js Versions:** 10/12

**Middy Versions:** 1.2.0

**AWS SDK Versions:** 2.713.0

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -64,7 +64,7 @@ const once = require('once')
  * @property {middlewareFunction} onError - the middleware function to attach as *error* middleware
  */
 
-const runMiddlewares = (middlewares, instance, done) => {
+const runMiddlewares = (middlewares, request, done) => {
   const stack = Array.from(middlewares)
   const runNext = (err) => {
     try {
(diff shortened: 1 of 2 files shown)


packages/core/index.js
```diff
@@ -149,44 +149,45 @@ const middy = (handler) => {
...
+          request.response = response
+          runMiddlewares(afterMiddlewares, request, (err) => {
             if (err) return errorHandler(err)
 
             terminate()
           })
         })
 
-        const handlerReturnValue = handler.call(instance, instance.event, context, (err, response) => {
+        const handlerReturnValue = handler.call(request, request.event, context, (err, response) => {
```
willfarrell, line 190: can you change `context` to `request.context`. There was another PR that modified this line.

PR #581, opened 2020-11-29
add sqs-json-body-parser to community middleware

Eomm:
I would like to add this first simple middleware to your list 👍


willfarrell:
How is this different from the existing `sqs` middleware? Missing feature?

Eomm:
Could you share a link?
I saw a sqs middleware per partial process elaboration 

In this case it just parse the body as a json using safe parser when needed

Eomm:
Why closing?

willfarrell:
Already committed to the readme :)

PR #582, opened 2020-12-08
Secretes Manager: Add ability to set string secret as environment variable

roni-frantchi:
Adds the ability to set string secret as environment variable.
It's often useful to have some secrets available on `process.env` in addition to the lambda's `context` instead of cascading it down a deeply nested function call stack.  

Todo list
---------

- [x] Added tests


willfarrell:
Thanks for creating this PR. I understand the use case and why it would be nice to have. However, I can't merge this, sorry. Setting secrets to the env is a big no-no. When an attacker finds a vulnerability, dumping the `process.env` is one of the easiest, and thus the first thing they will check, having secrets there will only make your life harder. I do appreciate you putting in the effort into this PR, however, I feel that security wins out over being developer-friendly in this case. Generally in these situations, refactoring can be done to achieve the best of both worlds. I'll leave this open for a bit to allow the community to weigh in.

Issue #585, opened 2020-12-16
Version 2


willfarrell:
See [/docs/CHAGNELOG.md](https://github.com/middyjs/middy/blob/release/2.x/docs/CHANGELOG.md) & [/docs/UPGRADE.md](https://github.com/middyjs/middy/blob/release/2.x/docs/UPGRADE.md) on `release/2.x` for what has changed.

## TODO / Ideas
- init commit: https://github.com/middyjs/middy/pull/587
- [x] security and performance best practice by default
- [x] Node v14, maybe v12, drop v10 (will use conditional chaining)
- [x] async/await core/middlewares (This will break existing middlewares) #356
- [x] AWS SDK v3 #584
  - [x] Update paginated requests to use generators
- [x] Unify middleware options 
  - [x] `context` manipulation
  - [x] FIPS 140-2 support - https://aws.amazon.com/compliance/fips/
  - [x] role assume (sts)
  - [x] caching
  - [x] enforse TLS connections https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/enforcing-tls.html
- [x] allow ssm to pull > 10 keys (if possible)
- [x] update avj to support draft-2019-09
- [x] Review and Update error handling
- [x] optimize dependency inclusion

If anyone has comments, suggestions, or must-have features. Please post them here.

## Documentation
  - [x] update options for each middleware
  - [x] detailed upgrade guide with why
  - [x] detailed changelog
  - [x] typescript section
  - [x] example showing chaining between middleware that use `internal`
  - [x] Examples showing middleware ordering (apigw, s3, sqs, sns, edge)
  - [x] Profiler documentation (includes clinicjs)
  - [x] bundling details - provide .yarnclean for middy and deps w/ ref to https://github.com/93v/starve
  - [x] serverless config recommendations
  - [x] lambda env for activating experimental flags - https://nodejs.org/api/cli.html#cli_node_options_options
  - [x] update logo to svg
- [x] AppoloClient common pattern https://github.com/middyjs/middy/issues/527
- [x] final proof read
- [x] update deprecation notice on older versions
- [x] updated code of conduct to v2

## Missing / Blocked
- [x] Node.js v14 support on AWS Lambda [Expected 2021-04-30](https://nodejs.org/en/about/releases/)
  - will transpile down to v12, not a blocker  
- [x] TypeScript Definitions
- [x] Profile running code  & micro optimmize
[...]

PR #587, opened 2020-12-29
Version 2-alpha

willfarrell:
https://github.com/middyjs/middy/issues/585

I will need assistance writing the TS definitions. Make a comment below if you'd like to help. I will be ready for help after this PR gets merged and we've given time to the community to comment on the changes.

Volunteers:
- @baileytincher


.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
```
baileytincher, line 22: Not that it's critical for tests, but since Lambda only supports 10.x and 12.x why make this change? https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html

willfarrell: For middy v2 I'm using esm and optional chaining. These are only enabled by default in node v14. middy v2 will not be released until AWS supports node v14, likely in Apr/May.

baileytincher: Why not use Babel to transpile to a Node 10 compatible version? https://github.com/middyjs/middy/blob/master/packages/core/.babelrc#L5

Looks like v1 core is still going all the way back to Node 6 which is probably worth updating regardless

baileytincher:
I see that you merged a draft of v2 🎉  If you're ready for PRs I can work on some of the TS issues throughout the coming week

.github/workflows/tests.yml
willfarrell: I try to avoid transpiler due to their configuration complexity and the increase maintenance they require. I have been reflecting more on this. During publish I will likely transpile a commonjs flavour to node v12 to allow for an easier upgrade to v2. I'll need to do more research and testing to best do this without complicating our testing suite.

AWS will likely deprecate node v10 in may this year.

willfarrell:
Go for it, check out the upgrade guide. 
- I would suggest starting with those with no option changes
- core should be fine to go
- ones with AWS services share a lot of the same options, but the fetch object is different. just scope for now, subject to change
- don't touch util, profiler, or rds-signer for now. subject to change

PR #592, opened 2021-01-03
Refactor core type generics naming

baileytincher:
Somewhat a WIP still. It would be good to come up with a way to test type definitions more thoroughly though. One option is to move some (or all) of the test files to TypeScript and that will naturally catch some potential breaking changes in future PRs.

I found the use of the `EventType` and `HandlerReturnType` definitions to be confusing as I was working on this. I think they can be simplified or removed entirely, but I'm going to mull over their purpose a bit. Would appreciate input on what they are meant to do as I did not find a need for such a construct in working on my own framework https://github.com/ShallotJS/shallot/blob/4781b6700e637d1e15dedf99d5202ddd799888b8/src/aws/index.ts#L44-L52

- Renames unclear generics in the type definitions of middy core. 
- Renames ambiguous references/imports.
- Replaces all uses of `any` with `unknown` https://stackoverflow.com/questions/51439843/unknown-vs-any

https://github.com/middyjs/middy/issues/585#issuecomment-752769767

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -1,57 +1,120 @@
 import {
-  Callback,
-  Context,
-  Handler
-} from 'aws-lambda'
+  Callback as AWSCallback,
+  Context as AWSContext,
+  Handler as AWSHandler,
+} from "aws-lambda";
 
-declare type EventType<T, C> =
-  T extends (event: infer EventArgType, context: C, callback: Callback<any>) => void ? EventArgType :
-  T extends (event: infer EventArgType, context: C) => Promise<any> ? EventArgType :
-  never;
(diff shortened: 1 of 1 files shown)


lmammino:
Great contribution @baileytincher! Thanks a lot.

How would you suggest to test these changes?

Also, be aware that the CI is currently failing the linting step.

willfarrell:
I think the linting error was from some earlier incomplete code I pushed. It was fixed last night. Re-pull, to resolve it.

@lmammino

willfarrell:
Closing in favour of #616

PR #612, opened 2021-02-02
Fix: onChange in v2

### LINKED ISSUE #611: v2.0.0-alpha4: error in SSM middleware
Since upgrading to v2.0.0-alpha4, a few of the students in my workshop have reported seeing this error during their exercises:

```
{
    "errorType": "SyntaxError",
    "errorMessage": "Unexpected token u in JSON at position 0",
    "stack": [
        "SyntaxError: Unexpected token u in JSON at position 0",
        "    at JSON.parse (<anonymous>)",
        "    at Object.onChange (/var/task/functions/get-restaurants.js:36:25)",
        "    at ssmMiddlewareBefore (/var/task/node_modules/@middy/ssm/index.js:157:175)",
        "    at runMiddlewares (/var/task/node_modules/@middy/core/index.js:207:86)",
        "    at middyPromise (/var/task/node_modules/@middy/core/index.js:100:15)",
        "    at Runtime.instance [as handler] (/var/task/node_modules/@middy/core/index.js:141:12)",
        "    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)"
    ]
}
```

The error is on a function whose handler looks like this:
```
module.exports.handler = wrap(async (event, context) => {
  const restaurants = await getRestaurants(process.env.defaultResults)
  const response = {
    statusCode: 200,
    body: JSON.stringify(restaurants)
  }

  return response
}).use(ssm({
  cache: true,
  cacheExpiryInMillis: 5 * 60 * 1000, // 5 mins
  names: {
    config: `/${serviceName}/${stage}/get-restaurants/config`
  },
  onChange: () => {
    const config = JSON.parse(process.env.config)
    process.env.defaultResults = config.defaultResults
  }  
}))
```

I haven't had the time to look into what's wrong with it exactly yet, but thought I'd give you a heads up that something's broken between v1.5.2 and v2.0.0-alpha4

willfarrell:
Fixes #611

### DIFF
--- a/packages/ssm/README.md
+++ b/packages/ssm/README.md
@@ -57,7 +57,7 @@ npm install --save @middy/ssm
 - `cacheExpiry` (number) (default `-1`): How long fetch data responses should be cached for. `-1`: cache forever, `0`: never cache, `n`: cache for n ms.
 - `setToEnv` (boolean) (default `false`): Store role tokens to `process.env`. **Storing secrets in `process.env` is considered security bad practice**
 - `setToContext` (boolean) (default `false`): Store role tokes to `handler.context`.
-- `onChange` (function(data)) (optional): Calls function when role tokens change after being initially set.
+- `onChange` (function(data)) (optional): Calls function when values change after being initially set.
 
 NOTES:
 - Lambda is required to have IAM permission for `ssm:GetParameters` and/or `ssm:GetParametersByPath` depending on what you're requesting.
--- a/packages/ssm/README.md
+++ b/packages/ssm/README.md
@@ -57,7 +57,6 @@ npm install --save @middy/ssm
 - `cacheExpiry` (number) (default `-1`): How long fetch data responses should be cached for. `-1`: cache forever, `0`: never cache, `n`: cache for n ms.
 - `setToEnv` (boolean) (default `false`): Store role tokens to `process.env`. **Storing secrets in `process.env` is considered security bad practice**
 - `setToContext` (boolean) (default `false`): Store role tokes to `handler.context`.
-- `onChange` (function(data)) (optional): Calls function when values change after being initially set.
 
 NOTES:
 - Lambda is required to have IAM permission for `ssm:GetParameters` and/or `ssm:GetParametersByPath` depending on what you're requesting.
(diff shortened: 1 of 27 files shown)


packages/ssm/README.md
```diff
@@ -57,7 +57,7 @@ npm install --save @middy/ssm
 - `cacheExpiry` (number) (default `-1`): How long fetch data responses should be cached for. `-1`: cache forever, `0`: never cache, `n`: cache for n ms.
 - `setToEnv` (boolean) (default `false`): Store role tokens to `process.env`. **Storing secrets in `process.env` is considered security bad practice**
 - `setToContext` (boolean) (default `false`): Store role tokes to `handler.context`.
-- `onChange` (function) (optional): Calls function when role tokens change after being initially set.
+- `onChange` (function(data)) (optional): Calls function when role tokens change after being initially set.
```
theburningmonk, line 60: `role token` doesn't make sense in this context

theburningmonk: maybe `parameter`?

willfarrell: Good catch, I also need to make another tweak to ensure it doesn't call on init.

willfarrell:
@lmammino @theburningmonk 
Just thinking on this. `onChange` isn't the best descriptor, `onFetch` might be better as the result may not have changed between fetches. Thoughts?

I currently have it implemented in this PR to pass in the data to onChange, but it feels like a hack for the use case in #611.

This is a better solution:
```js
middy(originalHandler)
  .use(ssm({
    fetchData:{
      key:'path'
    },
    setToContext: true
  }))
  .before((handler) => {
    process.env.defaults = handler.context.key.defaults
  })
```
For this specific use case I don't see a need for `onChange`.


> ... to chain middlewares that need to know when an upstream middleware requests a new value ...

Are there other known use cases that should be considered? Maybe we don't need an onChange hook right now?

lmammino:
What about removing `onChange` entirely and documenting this approach in the middleware?

willfarrell:
I agree, I think we can remove it entirely. This pattern is kinda already documented in the best practice section on the main readme. I'll update to remove the onChange, add an extra example to ssm, but leave in the hook to detect if it was pulled from cache for those building custom middleware that need to chain.

PR #613, opened 2021-02-05
Validator supporting ajv v7

leog:
Changed @middy/validator to support ajv v7.

Not aware of any issue with validator regarding latest ajv version.

N/A

Any other comments?
-------------------
Wasn't sure if a warning would be useful to let users know an older version of ajv was being used for the added plugins. Came up with the following code to be inserted before https://github.com/middyjs/middy/blob/master/packages/validator/index.js#L121.

```javascript
      pckJson = require(`ajv-${p}/package.json`)
      if(pckJson.dependencies && pckJson.dependencies.ajv && !semver.satisfies(semver.minVersion(pckJson.dependencies.ajv).version, "^7.0.0") ||  pckJson.devDependencies && pckJson.devDependencies.ajv && !semver.satisfies(semver.minVersion(pckJson.devDependencies.ajv).version, "^7.0.0")) {
        console.warn(`plugin ajv-${p} depends on a lower version of ajv v7`)
      }
```

That would mean `ajv-i18n` out-of-the-box plugin would trigger that warning.

Where has this been tested?
---------------------------
**Node.js Versions:** v12.18.0

**Middy Versions:** 1.5.2

**AWS SDK Versions:** N/A

Todo list
---------

[X] Feature/Fix fully implemented
[X] ~Added~ Updated tests
[ ] Updated relevant examples


willfarrell:
Thanks for this suggestion. However, the options passed in are very different between 6 and 7, thus if you don't override the default options you'll get an error. We have updated ajv to v7.x in middy@v2.x. Since its release is in the near term (1-2m), we won't introduce a change like this to v1.x. Sorry.

leog:
Hey @willfarrell. Thanks for the feedback and the insight. Looking forward to the v2 release then. Probably the `bsontype` plug-in example will need to change to another one like the `formats` as I implemented for this PR.

PR #617, opened 2021-02-20
Middleware JSDoc

willfarrell:
Types using JSDoc approach

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -1,4 +1,9 @@
 /**
+ * middy Namespace
+ * @namespace @middy/core
+ */
+
+/**
  * Middy factory function. Use it to wrap your existing handler to enable middlewares on it.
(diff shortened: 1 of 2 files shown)


lmammino:
Looks very good. Thanks for starting this.

One thing I would like to try is to use these JSDoc annotations to automatically generate TypeScript typings file.

This should be possible by running something like this:

```bash
npx tsc *.js --declaration --allowJs --emitDeclarationOnly --outDir .
```

A useful reference to see how much JSDoc the TypeScript compiler understands seems to be: [JSDoc Reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

Also, we could get some type checking (which could help us to validate the quality of the annotations) by adding:

```javascript
// @ts-check
```

At the top of every file.

Issue #618, opened 2021-02-21
New Middleware Proposal: CloudWatch custom metrics agent aws-embedded-metrics


fredericbarthelet:
This proposal includes the creation of a new middleware to hydrate `context.metrics` with a `MetricLogger` instance from [`aws-embedded-metrics`](https://github.com/awslabs/aws-embedded-metrics-node) and automatically flush after lambda execution.

PR #619, opened 2021-02-21
Add metrics middleware

fredericbarthelet:
Implements #618 

Where has this been tested?
---------------------------
**Node.js Versions:** 12

**Middy Versions:** 2.0.0-alpha6

**AWS SDK Versions:** No relevant

Todo list
---------

[ x ] Feature/Fix fully implemented
[ x ] Added tests
[ x ] Updated relevant documentation
[ x ] Updated relevant examples

### DIFF
--- a/packages/cloudwatch-metrics/package.json
+++ b/packages/cloudwatch-metrics/package.json
@@ -1,0 +1,56 @@
+{
+  "name": "@middy/cloudwatch-metrics",
+  "version": "2.0.0-alpha.6",
+  "description": "Embedded CloudWatch metrics middleware for the middy framework",
+  "type": "commonjs",
+  "engines": {
+    "node": ">=12"
+  },
+  "engineStrict": true,
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "index.js",
+  "types": "index.d.ts",
+  "files": [
+    "index.d.ts"
+  ],
+  "scripts": {
+    "test": "npm run test:typings && npm run test:unit",
+    "test:unit": "ava",
+    "test:typings": "typings-tester --config tsconfig.json index.d.ts"
+  },
+  "license": "MIT",
+  "keywords": [
+    "Lambda",
+    "Middleware",
+    "Serverless",
+    "Framework",
+    "AWS",
+    "AWS Lambda",
+    "Middy",
+    "HTTP",
+    "API",
+    "Metrics",
+    "Custom Metrics",
+    "Embedded Metrics",
+    "CloudWatch"
+  ],
+  "author": {
+    "name": "Middy contributors",
+    "url": "https://github.com/middyjs/middy/graphs/contributors"
+  },
+  "repository": {
+    "type": "git",
+    "url": "github:middyjs/middy",
+    "directory": "packages/cloudwatch-metrics"
+  },
+  "bugs": {
+    "url": "https://github.com/middyjs/middy/issues"
+  },
+  "homepage": "https://github.com/middyjs/middy#readme",
+  "dependencies": {
+    "aws-embedded-metrics": "2.0.4"
+  },
+  "gitHead": "7a6c0fbb8ab71d6a2171e678697de9f237568431"
+}
--- a/packages/cloudwatch-metrics/README.md
+++ b/packages/cloudwatch-metrics/README.md
@@ -1,0 +1,92 @@
+# Middy Cloudwatch metrics middleware
+
+<div align="center">
+  <img alt="Middy logo" src="https://raw.githubusercontent.com/middyjs/middy/master/img/middy-logo.png"/>
+</div>
+
+<div align="center">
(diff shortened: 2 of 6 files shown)


willfarrell:
I like it, I think this would be a great addition. Not 100% sold on the name yet. What do you think of `cloudwatch-metrics`? It feels long, but `metrics` feels too generic. @lmammino have any opinions on the name?

Would you be able to update this for merging into `release/2.x`? I'd like to avoid new additions to v1.x at this point.

Skimming through it's npm docs is `setProperty("RequestId", handler.path.to.RequestId)` required? Or is it set by default and this if for overriding?

lmammino:
I like `cloudwatch-metrics`. Before looking at the code, I was actually thinking that this middleware was providing generic hooks to send metrics to an arbitrary backend like statsd.

fredericbarthelet:
Thanks @willfarrell and @lmammino :)
I implemented your feedbacks :
- rebase on release/2.x 
- rename to `cloudwatch-metrics`

> Skimming through it's npm docs is setProperty("RequestId", handler.path.to.RequestId) required? Or is it set by default and this if for overriding?

`setProperty`, like `putMetric`, are part of the API available on the new `context.metrics` property. This method allows adding context specific property that you'd like to see on you Cloudwatch Insights console. There are no defaults for it, unlike namespace  and dimensions.

For exemple, in a lambda processing payment, you could use `setProperty("Total VAT", 89)` in case you'd like to have this metric available. Maybe my exemple (taken directly from the libs documentation) is not that much clear  ?

packages/cloudwatch-metrics/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "url": "https://github.com/middyjs/middy/issues"
+  },
+  "homepage": "https://github.com/middyjs/middy#readme",
+  "dependencies": {
+    "aws-embedded-metrics": "^2.0.4"
+  },
+  "devDependencies": {
+    "@middy/core": "^1.5.2",
+    "es6-promisify": "^6.0.2"
+  },
```
willfarrell, line 58: Can you remove devDependencies, they are not needed in v2

packages/cloudwatch-metrics/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "type": "git",
+    "url": "github:middyjs/middy",
+    "directory": "packages/cloudwatch-metrics"
+  },
+  "bugs": {
+    "url": "https://github.com/middyjs/middy/issues"
+  },
+  "homepage": "https://github.com/middyjs/middy#readme",
+  "dependencies": {
+    "aws-embedded-metrics": "^2.0.4"
```
willfarrell, line 53: Can we lock the version ` "aws-embedded-metrics": "2.0.4"`.

willfarrell (changes requested):

willfarrell (approved):

willfarrell:
@fredericbarthelet can you commit the `package-lock.json` as well please. It's required for the tests.

fredericbarthelet:
@willfarrell `package-lock.json` added :)
I noticed all package-lock uses `lockfileVersion: 1` except root package-lock, using v2 (from npm 7). Is this expected ? Should I target v1 or v2 for my lock file ?

https://github.com/middyjs/middy/blob/b75acc929df0f4f1bc75b99d6b9404b83b1e30af/package-lock.json#L4

willfarrell:
The lockfileVersion change is unexpected but good to know. My local must have finally updated causing the change. Can you update to use v1. The CI uses npm v6.

fredericbarthelet:
@willfarrell, I added a commit to change main `package-lock.json` file back to lockFileVersion 1 :)

willfarrell:
Thanks, looks like it's still failing on other lockfile v2's. I pushed a fix for these. One last rebase should do the trick, then we can merge in. Thank you for your patience on resolving this lockfile version issue. We have another PR that will properly bump all to lockfile v2 in the works.

fredericbarthelet:
@willfarrell I just rebased, everything's green now :)

willfarrell:
Great! I'm just waiting on another PR before minting a new version. Hope that's okay.

fredericbarthelet:
Thanks @willfarrell :) No problem waiting a bit more for v2 ;)

willfarrell:
@fredericbarthelet With AWS powertools released (https://github.com/awslabs/aws-lambda-powertools-typescript/tree/main/packages/metrics). Is their implementation better/easier to use than this middleware?

fredericbarthelet:
Hi @willfarrell, ultimately the powertool implementation is more complete and will be the prevailing implementation. However, in its current state, the powertool remains a beta-only, not to be used in production solution. I'd wait out of beta before deprecating the metrics middleware. Once they're out of beta, I'd be happy to update documentation with a migration path for the new solution.

willfarrell:
Perfect, I was thinking the same but wanted to make sure I got your feedback as you championed this middleware. I've added to the v4 list to revisit then, unless power tools hits production ready before v3 is released.

PR #624, opened 2021-03-06
Context authorizer from headers

MiguelNazMor:

Implements a new middy package to map headers values into requestContext.authorizer object.

This is motivated by the following issue:

While adding a proxy project in AWS Apigateway which points to a few different gateways (all of them with the same authorizer lambda) we got a problem when trying to fix the authorizer in the proxy level as the auhtorizer where fixing some values extracted from a jwt into the method.context.authorizer but it was not being passed to the proxied method. The solution seems to be map the method.context.auhtorizer.{value} to the integration.header.{value} and then use a middleware to re-map those values into the context.authorizer (to avoid changing all the code we were using to extract those values from the authorizer).

No

No

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Node.js Versions: 12.4

**Middy Versions: 1.5.2

**AWS SDK Versions: 2.796.0

Todo list
---------

[ Y] Feature/Fix fully implemented
[ Y] Added tests
[ Y] Updated relevant documentation
[ Y] Updated relevant examples


willfarrell:
Your current solution maps all of the headers to `context.authorizer.numberKey` which can be a security issue. I was unable to find AWS docs describing the limitations you're trying to solve here. Were you able to find any documentation on this since porting? It sounds like an AWS bug, a missing feature, or something that may be resolved by re-architecting how your AWS services communicate. Do you have a support ticket open yet?

If this is a documented AWS edge-case we'd be grateful to have this as part of the core middleware collection. If not the case, we'd be happy to have it as part of our list of community build middlewares.

Please note that middy v2 is almost ready for a beta release. We will not be adding new features to 1.x at this time.

Docs: 
- https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html
- https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-jwt-authorizer.html

PR #634, opened 2021-03-22
add: http-multipart-body-parser parse single keys with multiple files

huyennbl:

Single key with multiple file parser, but the key is not annotated with array:
Pseudo code for the request on client site:
```js
const formData = new FormData()
  fileList.forEach((file) => {
    formData.append('files', file)   // no [] here
  })
  
reqwest({
  url: 'http://localhost:3000/upload',
  method: 'post',
  data: formData,
  // others
}
```

No

Any other comments?
-------------------
Currently, if the request similar to what I described above, it would take only the last file.
If there's only single file per key, it is still an object, not an array to keep things, so fully compatible

Where has this been tested?
---------------------------
**Node.js Versions:**

**Middy Versions:**

**AWS SDK Versions:** …

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/http-multipart-body-parser/__tests__/index.js
+++ b/packages/http-multipart-body-parser/__tests__/index.js
@@ -184,4 +184,27 @@ describe('📦  Middleware Multipart Form Data Body Parser', () => {
     expect(Object.keys(response)).toContain('foo')
     expect(response.foo.length).toEqual(2)
   })
+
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks, I've also ported the change to v2.

PR #640, opened 2021-04-07
fix: correct typo in SSM package.

garnertb:
Infinite loop in @middy/ssm caused by typo recursion logic.

No.

fetchPath would infinitely iterate when there were > 10 parameters because the nextToken was undefined.  Below is a test case with 11 parameters.
```
//const fetchPath = (path, nextToken, values = {}) => {
fetchPath i:0, nextToken: undefined, values length: 0
fetchPath i:1, nextToken: undefined, values length: 10
fetchPath i:2, nextToken: undefined, values length: 10
fetchPath i:3, nextToken: undefined, values length: 10
...
```

After:
```
fetchPath i:0, nextToken: undefined, values length: 0
fetchPath i:1, nextToken: <token value>, values length: 10
done.
```

### DIFF
--- a/packages/ssm/index.js
+++ b/packages/ssm/index.js
@@ -106,7 +106,7 @@ const ssmMiddleware = (opts = {}) => {
             }
           })
         )
-        if (resp.NextToken) return fetchPath(path, resp.nextToken, values)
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for the PR and good catch! I'll put a release together this weekend.

Issue #641, opened 2021-04-08
Lighter and faster middleware


nponeccop:
This is a variation of performance issues reported in #639 

I have a prototype which has faster versions of:

- `@middy-validator` (that is #639)
- ~~`@middy/http-json-body-parser`~~ (already fixed by #643)
- `@middy/http-response-serializer`

It also validates that `@middy/core` and `@middy/http-error-handler` are already fast enough or at least that their performance impact is much less than of the three other middlewares.

The key achievement is that `serverless-offline` reports `Duration: 20 ms` while it's ~250 ms with current `@middy/*`

However the 12x+ achievement doesn't come for free: my versions perform less work than the upstream versions, and they have fewer features. They need to be polished still - e.g. the header issues are solved by hacks and don't work well. And of course we need an official benchmark to catch strange performance regressions in the future.

I guess a nearly zero overhead would be a very nice marketing feature for `@middy`. Of course we'd likely need full-fledged but slower versions too, for those needing extra features, or a way to customize the light middleware to plug in extra functionality. 

These are of course breaking changes, so it's for 3.x

willfarrell:
Looking forward to seeing them when they're ready.

willfarrell:
Pushed some optimizations for `@middy/http-response-serializer` to the main branch. https://github.com/middyjs/middy/commit/5278f7f47cb9fe34169a559a12ad4f3bb2fdf2d0

willfarrell:
Went digging into the flame graphs, Most of `@middy/http-response-serializer` time is spent inside `@hapi/accept`.

I'm going to close this issue. Now that we have tools in place to analyze the perf, there will be more improvements coming this year.

PR #643, opened 2021-04-08
@middy/http-json-body-parser perf improvement

willfarrell:
@nponeccop This should be much faster, let me know if this is what you had in mind.

Related to #641


nponeccop:
Here is my unscientific bench:

- `@middy/http-json-body-parser`: 30 32 34
- this PR: 26 30 30
- my [body-parser.js][1]: 15 19 16

Also note that the `content-type` parser allows everything permitted by https://tools.ietf.org/html/rfc7231#section-3.1.1.1 and then something more. I don't think we can ever fail on an RFC compliant input. However the degree to which we can handle non-complaint input is up to us. The tests of the `content-type` package seemingly allow more whitespace than the RFC.

Moreover my parser don't allow arbitrary application/foo+json content types. But if a user wants them, he can list them individually. It's very rare that you really need to support all json-related content types instead of a finite subset of them. Neither it is clear whether it's the regexp there that slows it down. It may be the http-errors package. So more research is needed.

[1]: https://github.com/nponeccop/serverless-openapi-templates/blob/ce5bf627d6557fb9fee3969bc486818c748a4f44/validation-middy/body-parser.js

willfarrell:
Yeah, it's the regex. I think supporting `application/foo+json` has value for those that use predefined standards. Let me play around with a few other ideas I have.

nponeccop:
My opinion is that the predefined standards also predefine the types. So it's only one or a few fixed standard-compliant content-types.

BTW `http-errors` also introduce a delay. I conditionally required it, so they only affect the "bad" path and the "good" path remains fast:

https://github.com/nponeccop/serverless-openapi-templates/commit/e820eb4869e2b03f14dcff156fc3eeeebf995565#diff-dd3eee162448fc7700f528048fe3d0d2b7b4b850685cdd3c716a295bf0f3beef

willfarrell:
After a bunch of jspref testing, I think this is a fast as it's going to get for now.

I did try passing in a static value and checking for an exact match but found it performed the same. Once we have a proper benchmarking tool in place we can revisit.

nponeccop:
> After a bunch of jsperf testing, I think this is a fast as it's going to get for now.

The current version is as fast as my best code. Congratulations!

> I did try passing in a static value and checking for an exact match but found it performed the same.

I think it's because the RegExp is still compiled. Also I wonder if there's measurable difference between // and '':

```js
const pattern = new RegExp('^application/(.+\\+)?json(;.*)?$')
```

BTW the PR has a typo: it's iMprovement.

nponeccop:
Formatting is wrong

willfarrell:
Good idea. The big gain was from only compiling the RegExp once.

```js
const slowest = new RegExp( '^application\/(.+\\+)?json(;.*)?$')
const fast = new RegExp( /^application\/(.+\\+)?json(;.*)?$/)
const fastest = /^application\/(.+\\+)?json(;.*)?$/
```

PR #665, opened 2021-04-26
Fix: Response Serializer does not replace entire object when falsey value is returned

### LINKED ISSUE #657: Regression: Response serializer no longer replaces response
Create a serializer that returns an object.  

What happens?
-------------
The serializer returned object is set on the body.  

The returned object should replace the entire response object as [the documentation states](https://github.com/middyjs/middy/tree/main/packages/http-response-serializer#serializer-functions):  

> _"...If an object is returned, the entire response object is replaced. This is useful if you want to manipulate headers or add additional attributes in the Lambda response."_

 (and as it was in v1) 

Any other comments?
-------------------
Seems to me like the behavior had changed since [this change](https://github.com/middyjs/middy/commit/610c64e47f85e210d942524d4665c264825d94ed#diff-d2f4c1e6809f1ffc51d2b64a7c53730ddaa4d2cef8193db00cf9c76f4711a8d0L42)

I think the only reason the [pre-existing unit test](https://github.com/middyjs/middy/blob/main/packages/http-response-serializer/__tests__/index.js#L208) did not break was because the response object itself [was mutated](https://github.com/middyjs/middy/blob/main/packages/http-response-serializer/__tests__/index.js#L217).  

**Node.js Version:** 14

**Middy Version:** 2.0.1

roni-frantchi:
In an attempt to address #657, `body` was checked to determined wether or not the entire response should be replaced (by @willfarrell).  
The problem with that fix is that any falsy value in `body` would not replace the entire response as well (think `false`, 0, or even `null` or `undefined` body).  

Resolves #657 

x

Any other comments?
-------------------
x

Where has this been tested?
---------------------------
**Node.js Versions:** 14

**Middy Versions:** 2.1.0

Todo list
---------

- [x] Added tests

### DIFF
--- a/packages/http-response-serializer/README.md
+++ b/packages/http-response-serializer/README.md
@@ -72,7 +72,7 @@ The function is passed the entire `response` object, and should return either a 
 
 If a string is returned, the `body` attribute of the response is updated.
 
-If an object is returned, the entire response object is replaced. This is useful if you want to manipulate headers or add additional attributes in the Lambda response.
+If an object with a `body` attribute is returned, the entire response object is replaced. This is useful if you want to manipulate headers or add additional attributes in the Lambda response.
 
 
 ## Content Type Negotiation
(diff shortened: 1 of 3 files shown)


willfarrell:
Good catch, and thanks for the quick PR.

PR #683, opened 2021-06-03
Custom serlializer for response body for http-error-handler

Amberlamps:
I want to throw errors in my code like this:

```javascript
throw createHttpError(StatusCodes.BAD_REQUEST, "Validation error", { errors: validation.errors });
```

In my logs I want `BadRequest: Validation error`

but in my api response I want:

```JSON
{
  "message": "Validation error",
  "errors": [{ ... }]
}
```

Unfortunately I can only achieve either of the two, but not both with the current implementation of http-error-handler, because the current implementation uses the `error.message` as the response body.

…

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Node.js Versions:** …

**Middy Versions:** …

**AWS SDK Versions:** …

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/http-error-handler/index.js
+++ b/packages/http-error-handler/index.js
@@ -33,7 +33,7 @@ const httpErrorHandlerMiddleware = (opts = {}) => {
     if (request.error?.expose) {
       request.response = normalizeHttpResponse(request.response)
       request.response.statusCode = request.error?.statusCode
-      request.response.body = request.error?.message
+      request.response.body = typeof options.serializer === "function" ? options.serializer(request.error) : request.error?.message
       request.response.headers['Content-Type'] =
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for submitting a PR. We encourage people to include their own error handler due to there not being a clear formatting standard everyone is using. We know `http-error-handler` will not work for all use cases but will catch anything that may not be caught.

Ideally, a middleware only does one thing. Currently, we don't have a way to handle what your PR addresses. This has been a common pain point for a few this year. We plan to address this in the next version which might be as early as this summer. The current idea is to reverse the order of the `onError` stack to allow the use of `http-response-serializer` to be run after `http-error-handler`.

PR #689, opened 2021-06-29
Update http-cors `credentials` documentation.

k-nut:
Updates documentation for the @middy/http-cors `credentials` option.

No.

Any other comments?
-------------------
I'm not sure about the original documentation. The `getOrigin` function also [does something](https://github.com/middyjs/middy/blob/main/packages/http-cors/index.js#L11) with `options.credentials` but I feel like the information that the `Acess-Control-Allow-Credentials` is also being set with this is certainly missing. I am happy to rephrase the line to also talk about the change to `Origin` if required.

Where has this been tested?
---------------------------
*Just a documetation change*

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/http-cors/README.md
+++ b/packages/http-cors/README.md
@@ -41,7 +41,7 @@ npm install --save @middy/http-cors
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for the PR!

PR #695, opened 2021-07-10
fix: Content-Type misspelling

rbasto1:
This fixes text responses with the correct mime type assigned for them (text/plain).


willfarrell:
Thanks for the PR!

PR #666, opened 2021-04-29
Typing Mutable Context

JimiPedros:
(Thanks for sending a pull request! Please make sure you click the link above to view the contribution guidelines, then fill out the blanks below.)

This is my first pull request to MiddyJS. At my workplace we use middleware that mutate the context of the request handler.  This pull request was inspired by the idea that we might be able to define the type of context which our handler expects based on the middleware used. I'm not 100% sure if this pull request makes sense but I am really keen to get some feedback thanks!

defining our mutable context
```
interface MutableContext extends Context {
  name: string
}
```

defining our mutable context request
```
type MutableContextRequest = middy.Request<APIGatewayProxyEvent, APIGatewayProxyResult, Error, MutableContext>
```

create a handler with context that we have defined (and expect) 
```
async function mutableContextDependantHandler (event: APIGatewayProxyEvent, context: MutableContext): Promise<APIGatewayProxyResult> {
  return {
    statusCode: 200,
    body: `Hello from ${context.name}`
  }
}

handler = middy(mutableContextDependantHandler)
```

request.context.name is type safe
```
let mutableContextMiddleware = {
  before: (request: MutableContextRequest) => {
    request.context.name = 'Foo'
  },
}

handler = handler.use(mutableContextMiddleware)
```

accessing name from context fails as property `name` does not exists on type MutableContext
```
async function handler (event: ..., context: MutableContext): Promise<APIGatewayProxyResult> {
  return {
    statusCode: 200,
    body: `Hello from ${context.value}` // Fails
  }
}

...

let mutableContextMiddleware = {
  before: (request: MutableContextRequest) => {
    request.context.value = 'Foo' // Fails
  },
}
```

What does this implement/fix? Explain your changes.
---------------------------------------------------
- This allows us to declare a type of context of our handler function expects.

Does this close any currently open issues?
------------------------------------------
- no, it will probably introduce issues.

Any relevant logs, error output, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
- This pull request depends on changes made to @types/aws-lambda
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52644

Where has this been tested?
---------------------------
**Node.js Versions:** …
- 14.15.4

**Middy Versions:** …
- 2.1.1

**AWS SDK Versions:** …
- @types/aws-lambda 8.10.75

Todo list
---------

[x] Feature/Fix fully implemented
[x] Added tests
[...]

Issue #589, opened 2020-12-30
TypeScript Support [Help Wanted]


willfarrell:
TypeScript support is made up 100% from community contributions. None of the core maintainers know or use TypeScript and thus are unable to offer support for it. We believe that TypeScript has value to some and are willing to merge in changes the community find valuable. 

If you would like to make Middy better, please consider submitting, reviewing, and/or commenting on PRs/issues. We generally like to give a week before merging a PR to give the community time to comment.

[TypeScript Issues](https://github.com/middyjs/middy/issues?q=is%3Aopen+is%3Aissue+label%3Atype-systems)

lmammino:
Any reason for moving type definitions outside every project and into `@types/x` ?

I personally prefer to keep them colocated just to make it easier to update them every time something changes.

Am I missing something important here?

willfarrell:
It was more of a question on what is TS best practice. See https://github.com/middyjs/middy/issues/591
Definitions are not going anywhere.

dbartholomae:
Best practice would be to have the type annotations right next to the code. If they are a separate file anyway, then maintaining them inside the repo in my experience works a bit better, as long as there are people who actively keep them updated and whose PRs are merged on time.
Concerning types in general, unfortunately, the setup with `.use` makes it hard to give really good type safety. [When we last discussed improved types for this repo](https://github.com/middyjs/middy/issues/300#issuecomment-578435413), I experimented around it a bit which lead to [lambda-middleware](https://github.com/dbartholomae/lambda-middleware). I'm still happy to help here as well, though this structural problem will limit what we can achieve.

willfarrell:
Thanks for commenting and bring attention to that last discussion. I've read a couple times today, will likely need another read. I'm wondering how others like `express` and other that also use the `.use` pattern have solved this TypeScript incompatibility. Would allowing `middy(handler, middlewares)` be all that is needed to allow better typing?

Happy to accept any help you can spare. I've updated the list at the top highlight deficiencies in v2. Just open a PR, I'm pretty active these days.

dbartholomae:
I'll take a look. I fear that there might be further complications, though, as TypeScript will need to know the order in which middlewares are applied, and this order only partially depends on the order in which the middleware is added.

willfarrell:
There are some lesser known hooks in core that might help.

1. `.use([...])` can take an array of middlewares
2. `middy(handler).__middlewares` returns the arrays of the `before`, `after`, `onError` in their execution orders.

[...]

PR #696, opened 2021-07-26
Typescript: Typed Context

mbarton:

Some middleware adds extra fields to the context. Typescript will complain if the fields are not declared optional, even if the middleware guarantees they are set before invoking the next step. With this pull request you can specify a custom Context type, including those fields.

This is not yet fully automated inference of the Context type by combining types from each middleware (see https://github.com/middyjs/middy/issues/589) but rather a first step towards it. You still have to update the type manually as you add/remove middleware.

The original PR was https://github.com/middyjs/middy/pull/666 by @JimiPedros. I have just rebased and fixed the automatic type tests.

First define your context type:

```typescript
interface CustomContext extends Context {
  name: string
}
```

and then use it in your handler:

```typescript
async function customHandler(event: APIGatewayProxyEvent, context: CustomContext): Promise<APIGatewayProxyResult> {
  return {
    statusCode: 200,
    body: `Hello from ${context.name}`
  }
}

handler = middy(customHandler)
```

You can also use the custom type when defining middleware:

```typescript
let customMiddleware = {
  before: (request: CustomContext) => {
    request.context.name = 'Foo'
  },
}

handler = handler.use(customMiddleware)
```

Accessing a field that does not exist on the context is a failure (in Typescript `strict` mode):

```typescript
async function handler (event: ..., context: CustomContext): Promise<APIGatewayProxyResult> {
  return {
    statusCode: 200,
    body: `Hello from ${context.value}` // Fails
  }
}

...

let customMiddleware = {
  before: (request: CustomContext) => {
    request.context.value = 'Foo' // Fails
  },
}
```

No. It may cause issues for existing Typescript users however I've verified it works on a @lego-shop and at worst we could back out the changes or fix forward.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -17,9 +17,9 @@ interface PluginObject {
   requestEnd?: PluginHookPromise
 }
 
-interface Request<TEvent = any, TResult = any, TErr = Error> {
+interface Request<TEvent = any, TResult = any, TErr = Error, TContext extends LambdaContext = LambdaContext> {
   event: TEvent
-  context: LambdaContext
+  context: TContext
   response: TResult | null
   error: TErr | null
   internal: {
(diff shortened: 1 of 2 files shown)


JimiPedros (commented):
Nice job @mbarton 🙂

willfarrell (approved):

willfarrell:
Looks good to me. @mbarton is this ready for merge?

mbarton:
Yep it's good to go!

willfarrell:
Awesome, thank you. I'll try release a new version this weekend.

Issue #594, opened 2021-01-03
New Middleware Proposal: S3


willfarrell:
Proposal for a new S3 middleware that could pull string or json from a file in S3 and store to internal.

See ssm for what it could look like: https://github.com/middyjs/middy/tree/main/packages/ssm

Please 👍🏻  this issue if it would have value to you. I'll build it if there > 10 👍🏻

If you can comment with your use case below, that would be helpful.

willfarrell:
@iDVB I was thinking of your middleware could benefit from this. Allow it to be simpler.

iDVB:
thanks @willfarrell this would be great. I think there would have to be some control on caching as well. 

Right now `middy-reroute` caches in memory and on CloudFront responses. And the cache time is configurable. 

I'd be interested in talking more or helping out with this.

willfarrell:
I was thinking it would look just like this, just with s3. https://github.com/middyjs/middy/tree/release/2.x/packages/ssm

iDVB:
Yup, that looks about right. Just a thought. This would only work with one of my S3 calls for my default config file. However, there are other "conditional" calls for 404.html etc that would likely still benefit from being baked into my middleware.

willfarrell:
How the aws middlewares are structured now support multiple requests based on a key mapping. There is no fetch only when needed, but there is resolve only when needed :). Let prefetch request everything you do need, resolve only what you need when you need it. There is an new `getInternal` util function that does all the magic in v2 :).

willfarrell:
Closing, been year without enough interest.

iDVB:
Sorry @willfarrell, been super busy on a number of things. This still sounds interesting to me. Is this still something that applys?

willfarrell:
Re-opening based on a comment from the powertools team (https://github.com/middyjs/middy/issues/948#issuecomment-1336476055).

willfarrell:
I've created PR #961 for this. Please feel free to comment or contribute types.

PR #716, opened 2021-08-29
S3 download manager

KillDozerX2:
Adds a new middleware for downloading s3 records to local automatically for lambdas connected to S3

None that I know

No, Just used this middleware and thought to contribute to source

Any other comments?
-------------------
There is a very good chance that this exists already but I couldn't find it and we use this in our organisation.
Also, I managed to somehow pull off Ava, but someone should look at my tests

Where has this been tested?
---------------------------
**Node.js Versions:**
v14.17.5

**Middy Versions:** …
v2.5.1
**AWS SDK Versions:** …
v2.964.0

Todo list
---------

[x] Added tests
[x] Updated relevant documentation
[x] Updated relevant examples

### DIFF
--- a/packages/s3-download-record/README.md
+++ b/packages/s3-download-record/README.md
@@ -42,7 +42,8 @@ npm install --save @middy/s3-download-record
 - `awsClientCapture` (function) (optional): Enable XRay by passing `captureAWSClient` from `aws-xray-sdk` in.
 - `httpsCapture` (function) (optional): Enable XRay by passing `captureHTTPsGlobal` from `aws-xray-sdk` in.
 - `disablePrefetch` (boolean) (default `false`): On cold start requests will trigger early if they can. Setting `awsClientAssumeRole` disables prefetch.
-
+- `directory` (string) (option): The directory where the file will be downloaded, defaults to `/tmp`
+- `prefixBucketName`: Defines whether or not to add the bucket name before the object key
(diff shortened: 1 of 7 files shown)


willfarrell:
Thanks for the PR, questions:

1. What is the use case for this? 
  - Depending on what that is I wonder is a change in architecture is necessary and something like `s3-object-response` might be useful.
  - There is also: https://github.com/middyjs/middy/issues/594 which would likely end up being `s3`
1. You're saving the file to disk? Why? Unless there is EFS attached, you only have 500MB, which is small enough to just keep in memory.
  - If you're working with large files, setting the request stream to the context would make more sense
  - If only dealing with small files, maybe it should auto parse (ie JSON) based on mime?
1. You have an option for prefixing with bucket? are you planning to have 1 lambda process events from multiple buckets?

Just a heads up, depending on the use case and how common it is, we may not included it in the core repository. However, it can be of course be added to our community middleware list :).

KillDozerX2:
Hey, thanks for the response, we use to resize and modify images using sharp and yes there is an EFS attached which is later used by a cloud watch timed function to train some model(no idea how that works). I'd be okay to move it to community list, let me check that out quickly.

willfarrell:
I have a lambda that does the same thing. Just do it in the handler if it's the only lambda that will use it.

PR #729, opened 2021-10-04
fix(ssm): Move devDep to Dep for exported TS def

### LINKED ISSUE #728: Cannot find module 'aws-xray-sdk'
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

1. `import ssm from '@middy/ssm'` in a TS project 

What happens?
-------------
building causes the following error....
```
node_modules/@middy/ssm/index.d.ts:2:34 - error TS2307: Cannot find module 'aws-xray-sdk' or its corresponding type declarations.
2 import { captureAWSClient } from 'aws-xray-sdk'
```

No errors. Definitely not relating to types for a dependency's dependency.

Other Thoughts
----------------
We are using aws-sdk v3 for our own needs aside from middy.... I don't see any peer or dep on the sdk in `@middy/core` or `@middy/ssm` surprisingly.... does this mean it just uses the default provided by AWS?

**Node.js Version:** v14.15.1
**Middy Version:** v2.5.1
**AWS SDK Version:** v3
    "@aws-sdk/client-firehose": "^3.34.0",
    "@aws-sdk/client-s3": "^3.34.0",

iDVB:
- Fixes #728

Fixes #728 
`aws-xray-sdk` needs to be a "Dependency" not a devdep.
@willfarrell Cheers! [Any deps of "exported" TS definitions should be set as "dependencies"](https://stackoverflow.com/a/46011417/1241688) since npm/yarn does not install devDeps-of-deps.

Fixes #728 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Node.js Versions:** …

**Middy Versions:** …

**AWS SDK Versions:** …

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples


iDVB:
@willfarrell I don't think this would effect anyone, not the TS devs or non-TS devs with regard to production files. That dependency is only used in the `*.d.ts` definition file would means it will not be used by non-TS developers. While TS developers only use definition files as part of the build process.

The only thing that would be very slightly larger (45KB) is the `node_modules` directory. But this is all just typical process to running a TS lib as far as I can tell.

willfarrell:
I've added a note to the README instructing to add missing devDependencies when needed. I agree that 45KB is small in this case. However, it would be inconsistent to do it for just one dependency, all would be needed which adds up significantly more (ie `aws-sdk`).

iDVB:
Sure, still not sure I understand what the concern is? 

1. It's likely to be far more common for people to be bundling their function code, which means no node_modules folder and all this stuff gets shaken out. 
2. As an example, Yarn v2-3 doesn't even have a node_modules folder, and would likely get shaken out of the production cache folder
3. It's common practise for JS lib owners to list deps or exported definitions as "dependencies".

willfarrell:
1. I would agree this is the case most of the time with javascript, but not for node.js on AWS lambda. At least in the code bases I've seem. Treeshaking isn't the default for popular tools for lambda yet.
2. We've opted to using the default package manager shipped with node. I hear this may change in future version of node.
3. I'll take your word that's the case in TypeScript first modules. I know aws-sdk v3 took that approach and got blasted by the community for adding 200MB dependencies just for definition files.

I'm trying to balance all use cases. I don't see this as a net positive at this time. I will keep this in mind as the ecosystem evolves.

Issue #738, opened 2021-10-21
Typescript type changes break promise calling pattern


markcarroll:
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

What are the steps to reproduce this issue?
-------------------------------------------
1. Upgrade to 2.5.2
2. Unit tests fail. For example:
```ts
  it("should fail 403 forbidden with missing parameters", async () => {
    const result = await handler({} as any, {} as any);
    expect(result.statusCode).toEqual(403);
  });
```
now gets:
> Expected 3 arguments, but got 2.

What happens?
-------------
All our unit tests that call the lambda handler with the AWS-supported Promise calling pattern fail because PR #704 now forces the type to require a callback parameter where previously it was not requires (as per AWS). 

What were you expecting to happen?
----------------------------------
Upgrade, nothing breaks. We should be able to call a handler with a `event` and `context` and get a Promise back. This does not require a callback parameter if we use lambda typings directly. 

This is the change that caused the problem. This worked previously:
```ts
declare type MiddlewareHandler<THandler extends LambdaHandler<any, any>, TContext> =
  THandler extends LambdaHandler<infer TEvent, infer TResult> // always true
    ? (event: TEvent, context: TContext) => Promise<TResult>
    : never
```
but when it was changed to:
```ts
type MiddyInputHandler<TEvent, TResult, TContext extends LambdaContext = LambdaContext> = (event: TEvent, context: TContext, callback: LambdaCallback<TResult>) => void | Promise<TResult>

declare type MiddlewareHandler<THandler extends LambdaHandler<any, any>, TContext extends LambdaContext = LambdaContext> =
  THandler extends LambdaHandler<infer TEvent, infer TResult> // always true
    ? MiddyInputHandler<TEvent, TResult, TContext>
    : never
```
it now requires a third `callback` parameter.  This means all the AWS samples that use Promises now come up with missing parameter errors. 

Easy solution would be to make the `callback` parameter optional like this:

```ts
type MiddyInputHandler<TEvent, TResult, TContext extends LambdaContext = LambdaContext> = (event: TEvent, context: TContext, callback?: LambdaCallback<TResult>) => void | Promise<TResult>
```

What versions of software are you using?
----------------------------------------
**Node.js Version:** …14

**Middy Version:** …2.5.2

**AWS SDK Version:** …2.x

willfarrell:
@markcarroll any interest in creating a PR for this?

[...]

PR #739, opened 2021-10-28
fix(types): optional callback

nazarvovk:
Implements the fix for issue #738


nazarvovk:
Closing this, because the approach I wanted to take here seems to be not possible due to a known limitation of typescript.

More on that [here](https://stackoverflow.com/a/69770240/10081669).

PR #749, opened 2021-11-15
feat/types/remove-aws-xray-sdk-deps

J4YF7O:
In several plugin definitions `aws-xray-sdk` was used. The problem is that we need to be able to continue using those plugins without being dependent on `aws-xray-sdk`.

In this PR I propose to remove this dependency, and to use the `Options` interface defined in `utils` to define the interfaces of each dependent plugins.
This allows a simpler management of types, in case of update of `aws-xray-sdk` (or other changes), we will have only one element to modify. The type of `awsClientCapture` in `utils.Options` (or other attribute types...)

There are some plugins that use only some attributes of the "Options" interface, for this case I use the typescript utility type : [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys)

[Typescript comment related](https://github.com/middyjs/middy/issues/589#issuecomment-968989688)

Where has this been tested?
---------------------------
- npm run test ✔️

### DIFF
--- a/packages/rds-signer/index.d.ts
+++ b/packages/rds-signer/index.d.ts
@@ -1,19 +1,8 @@
 import { RDS } from 'aws-sdk'
-import { captureAWSClient } from 'aws-xray-sdk'
 import middy from '@middy/core'
+import { Options as MiddyOptions } from '@middy/util'
 
-interface Options<Signer = RDS.Signer> {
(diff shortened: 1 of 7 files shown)


willfarrell:
Can someone else that uses TypeScript give this a review before I merge?

PR #773, opened 2022-01-06
Pass the request into requestStart so we can action on it

nicwise:

I need to do some work before the async contexts are setup, and the only way I can find to do it is in a plugin, in `requestStart`. Problem is, I dont have the **request** to get info from!

So, pass it in.

-----------------------------------------
Nope

Nope.

Any other comments?
------------------
Not really. Makes sense to have the request that is starting when starting a request, tho.

Where has this been tested?
---------------------------
**Node.js Versions:** 14.18

**Middy Versions:** latest (2.5?)

**AWS SDK Versions:** change doesn't affect it?

Todo list
---------

[ ] Added tests
[ ] Updated relevant examples

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -5,12 +5,13 @@ import {
 } from 'aws-lambda'
 
 declare type PluginHook = () => void
+declare type PluginHookWithRequest = (request: Request) => void
 declare type PluginHookWithMiddlewareName = (middlewareName: string) => void
 declare type PluginHookPromise = (request: Request) => Promise<unknown> | unknown
 
 interface PluginObject {
   beforePrefetch?: PluginHook
-  requestStart?: PluginHook
+  requestStart?: PluginHookWithRequest
   beforeMiddleware?: PluginHookWithMiddlewareName
   afterMiddleware?: PluginHookWithMiddlewareName
   beforeHandler?: PluginHook
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -5,7 +5,6 @@ const middy = (baseHandler = () => {}, plugin) => {
   const onErrorMiddlewares = []
 
   const instance = (event = {}, context = {}) => {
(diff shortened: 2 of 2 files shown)


willfarrell:
Thank for the PR. Could you share more details on your use case and what you're try to achieve? What `async contexts` are you referring to?

Have you tried the following?
```js
const handler = middy(baseHandler)
  .before((request) => {
    // change request here
  })
  .use(...)
  ...
```

Nothing is really "setup" between `requestStart` and the first `before` middleware. Everything is "setup" between `beforePrefetch` and the first `requestStart` (aka handler invocation scope).

nicwise:
Thanks @willfarrell I'd either missed or... something that `.before` call.

I'm using [Honeycomb's Beeline](https://github.com/honeycombio/beeline-nodejs) library for observability instrumentation, and the [initial trace has to be setup](https://github.com/honeycombio/beeline-nodejs/blob/main/docs/API.md#starttrace) before ANY async call is made, as it uses async context stuff, specifically the node `async_hooks`

```
Starts a new local trace and initializes the async context propagation machinery.
```

I'll try and use the `.before()` call - sounds like it'd work! Thanks!

nicwise:
thinking about it tho:

https://github.com/middyjs/middy/blob/main/packages/core/index.js#L76-L86

the before middlewares are run inside the async method, so its likely to not work. I'll try it out anyway tho. I'll also try and patch my node_modules version of middy to make sure the change I put in here will actually do what I want.

willfarrell:
Thanks for sharing, super helpful. Sounds like you'd need to wrap the resulting handler itself to achieve this ... maybe

```js
module.exports = wrapper(middy()...)
```

nicwise:
Fair enough. I'll try the wrapper option, which makes sense 🤦🏼

willfarrell:
Okay, I'm going to close this for now.

Issue #602, opened 2021-01-09
New Middleware Proposal: WebSockets Response


willfarrell:
Proposal to add in a middleware to manage websocket responses through API Gateway.

Please 👍🏻 this issue if it would have value to you.

If you can comment with your use case below, that would be helpful.

Ref:
- https://www.youtube.com/watch?v=ZuoZpf3JnY8
- https://ably.com/blog/introducing-the-websocket-handbook

KillDozerX2:
👍🏽
A few questions.

- What are some of the use-cases that you have in mind
- Do you want more help with it?

willfarrell:
At DataStream.org we process large dataset files. I was thinking for the next iteration of our app (whenever that happens) that we could use ws to push progress indicators. Browser push notifications may also work, but I haven't used them before, and may have extra costs. Lots of scoping to do still.

I'm always looking for help. If you have a more near term use case you're welcome start a PR.

KillDozerX2:
We had a simpler use case where we take the response body and use the connection ID from the event, stringify it and send it to the user

willfarrell:
Closing, see PR #787

PR #787, opened 2022-01-17
New Middleware: WebSocket Response

willfarrell:
new middleware #602

cc @KillDozerX2

### DIFF
--- a/packages/ws-response/index.js
+++ b/packages/ws-response/index.js
@@ -1,0 +1,62 @@
+import ApiGatewayManagementApi from 'aws-sdk/clients/apigatewaymanagementapi.js'// v2
+// import {ApiGatewayManagementApi} from '@aws-sdk/client-apigatewaymanagementapi' // v3
+
+import { canPrefetch, createClient, createPrefetchClient } from '@middy/util'
+
+const defaults = {
+  AwsClient: ApiGatewayManagementApi,
+  awsClientOptions: {}, // { endpoint }
+  awsClientAssumeRole: undefined,
+  awsClientCapture: undefined,
+  disablePrefetch: false
+}
+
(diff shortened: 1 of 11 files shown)


packages/ws-response/index.js
```diff
@@ -0,0 +1,62 @@
...
+      statusCode: 200
+    }
+  }
+
+  return {
+    after: wsResponderMiddlewareAfter
+  }
+}
+
+// TODO move to @middy/util?
```
lmammino, line 49: I would consider this specific to websockets (therefore relevant only for this middleware). Or are you anticipating users might want to use this even without the middleware?

lmammino (approved):

packages/ws-response/index.js
willfarrell: Because there is only one `ws` middleware, we don't need it in `util`. But can move if future use cases arise. I hope to be able to use this myself later in the year.

packages/ws-response/index.js
```diff
@@ -0,0 +1,62 @@
...
+      client = await createClient(options, request)
+    }
+
+    await client
+      .postToConnection(response)
+      .promise()
+
+    request.response = {
+      statusCode: 200
+    }
```
KillDozerX2, line 41: It would be a good idea if we could somehow attach the original response to request if some other middleware wants it.
For example, in our use case, after having sent a response, we invoke the next step in the step function where we need the response and therefore cannot use this since it will change the response after having sent it.

willfarrell: If you can create a PR to remove L39-41 and update any tests, that works for me.

KillDozerX2: Will do shortly

PR #789, opened 2022-01-21
Inaccurate return type for middy

nem035:
Even though we can set custom event and context types for our handler, AWS will still send (potentially) different data at runtime and we can't guarantee what they'll send.

Meaning that we cannot assume that the returned handler still accepts the custom event and context that we've given to the original `handler`.

```ts
import { Handler } from 'aws-lambda';

const handler: Handler<{ x: 5 }> = ({ x }) => ({ statusCode: 200, body: String(x + 1) })

const middlewarizedHandler = middy(handler);
//    ^^^^^^^^^^^^^^^^^^^^ this handler
// it can't expect an event that has a property x 
// and that that property has a value of 5
```

Right now the type of `middlewarizedHandler` would be `middy.MiddyfiedHandler<{ x: 5 }, any, Error, Context>` but it should probably be something like `middy.MiddyfiedHandler<any, any, Error, Context>`.

AWS can invoke the `middlewarizedHandler` without `x` but the types won't let that happen at dev time. 

```ts
middlewarizedHandler({}, {}, () => {});
// Argument of type '{}' is not assignable to parameter of type '{ x: 5; }'.
//  Property 'x' is missing in type '{}' but required in type '{ x: 5; }'.(2345)
//  input.tsx(4, 26): 'x' is declared here.
```

[Here's a quick TS playground to demonstrate the issue
](https://www.typescriptlang.org/play?ssl=9&ssc=1&pln=9&pc=39#code/JYWwDg9gTgLgBCYATJBPOAzKERwOQACiKqA9AMbQCmeA3AFCiSxwDecAEgIYB2SANlShwAvpmy48XAO4BnALT8uIAEZIudevUo9Z8ABa8BQgFycjgqAB52ADzMBWUQD44AXjhdZqHuTgAKO1EASndXQLg9LhgAV1kAYQgkKjMAJgAGdIAaOBUk1DMAZRgoYB4Ac39bOABqOABGUJFgrR09BGRjaS5SgC8qJG4+S3cOkn9DYaFghlJSOAW4AD0V1bX19bh6ObgYfWBZTBjfGGAIHjhyXjx4KlswKnJ4XjgqADcqHng96LhDQ64cDA2AesHQ1SMu0M32hQJBQhg6H+njgby4-BiVDgEAwcAc9GIXR6wH6gwsQkCIhyrCpAVCblcNOCQA)

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -69,7 +69,7 @@ declare type MiddlewareHandler<THandler extends LambdaHandler<any, any>, TContex
  * @param handler your original AWS Lambda function
  * @param plugin wraps around each middleware and handler to add custom lifecycle behaviours (e.g. to profile performance)
  */
-declare function middy<TEvent = any, TResult = any, TErr = Error, TContext extends LambdaContext = LambdaContext> (handler?: MiddlewareHandler<LambdaHandler<TEvent, TResult>, TContext>, plugin?: PluginObject): MiddyfiedHandler<TEvent, TResult, TErr, TContext>
+declare function middy<TEvent = any, TResult = any, TErr = Error, TContext extends LambdaContext = LambdaContext> (handler?: MiddlewareHandler<LambdaHandler<TEvent, TResult>, TContext>, plugin?: PluginObject): MiddyfiedHandler<any, TResult, TErr, any>
 
 declare namespace middy {
   export {


willfarrell:
Thanks for the PR! Remove the draft status when it's ready to go. Can you update it to be against the `release/3.x` branch please.

calvin-summer:
hi @nem035 . I think I can clear up what you're experiencing here, and we can close this PR since it would be a regression in type safety in any projects that are using the `TEvent` type parameter, including mine.

I don't think you actually want this result, but if your intention is to say that the `event` can be `any`, you are able to type that without any changes to middy. It's actually the default value for `TEvent` on both `Handler` from `aws-lambda` and `middy` from `@middy/core` in your example. See where `<TEvent = any ...` in the type definition excerpts below:
```
// (Handler from aws-lambda)
type Handler<TEvent = any, TResult = any> = (event: TEvent, context: Context, callback: Callback<TResult>) => void | Promise<TResult>
            ^^^^^^^^^^^^^
```
```
// (middy from @middy/core)
declare function middy<TEvent = any, TResult = any, TErr = Error, TContext extends LambdaContext = LambdaContext> (handler?: MiddlewareHandler<LambdaHandler<TEvent, TResult>, TContext>, plugin?: PluginObject): MiddyfiedHandler<TEvent, TResult, TErr, TContext>
                       ^^^^^^^^^^^^
```

So, in your code, you could specify `any` as the type parameter (i.e. `Handler<any>` or `middy<any>`) or you could allow the default to be applied (i.e. `Handler` or `middy`) to achieve an `any` type on the `event` parameter.

*However*, I think there might be some confusion here regarding aws types.

>Even though we can set custom event and context types for our handler, AWS will still send (potentially) different data at >runtime and we can't guarantee what they'll send.

aws provides many types in their `aws-lambda` package, including types for their `event`s. You can use these types as a contract for what properties aws will define on those parameters. So while you *can* declare that a `handler` will take an event with type `{ x: 5 }`, aws will not be sending anything like that. They will send something that matches the appropriate event type for your lambda. For example, if your lambda is behind an API gateway, you can use the type `APIGatewayEvent` as your `TEvent` type parameter (`Handler<APIGatewayEvent>` or `middy<APIGatewayEvent>`). 
Then, in your handler function, you can access properties on the `event` and have confidence that they are accurately defined and will exist at runtime.
I modified your playground example to demonstrate [here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBCYATJBPOAzKERwOQACiKqA9AMbQCmeA3AFCiSxwDecAEgIYB2SANlSgAaOAEEACgEkA4lxhUA7l1QBRAG5Ue8AL6ZsuPF0UBnALT8uIAEZIudevUo8T8ABa8BQgFydPgqAAeSVl5JRUNLRgAPjgAXjguE1QecjgqTW142PZ6OHy4Z1c4KCoARwBXKlcOKi4kIRN49MyYADo3OoaoEwYCwogXeFLK6pgAIQg0Zoyotusp1D6dRyL4Yi9lKGAALyokbj4A5o3UAAoPI6EASlogA).

I hope this helped :)

willfarrell:
@calvin-summer Thanks! Is this something that should be in the docs?

calvin-summer:
> @calvin-summer Thanks! Is this something that should be in the docs?

Yeah i think that could be very helpful for people getting started with typescript and middy :D

nem035:
> So while you can declare that a handler will take an event with type { x: 5 }, aws will not be sending anything like that. 

Unfortunately this isn't correct as the `aws-sdk` allows you to invoke any lambda with anything you want. This is actually where I noticed this discrepancy, I was working with custom events.

calvin-summer:
> I do agree with you that my exact change isn't necessarily the right solution and might not be necessary (at least not as is).
> 
> The reason I made this PR is to make a note of the situation and start a discussion.
> 
> I haven't had the time to fully familiarize myself with middy's codebase in order to propose something more robust.
> 
> As far as working with custom events, the other way to solve it is to always type it yourself as `any`, or a `Partial` or whatever is appropriate for your own system.
> 
> This might be the way to go moving forward but note that yes absolutely you can receive events that 1) aren't in the shape of any of the types in `aws-lambda` and 2) that those types aren't always following the exact contracts deployed in production, especially for new services.

oh okay, so invoking a lambda yourself with custom events. in that case, you should check out the top of `@types/aws-lambda/handler.d.ts` where they discuss [defining types for custom events](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4e3fcebbd592edc35a269c7b3bb93f24e84b9ebd/types/aws-lambda/handler.d.ts#L12)

in this case, it is up to you to ensure that your lambda is invoked with with the custom event type that you define for your handler
i.e.
```
interface CustomEvent { customField: string }

// handler definition
handler<CustomEvent>

// invoke call
invoke({..., Payload: JSON.stringify({ customField: 'foo' })})
```

PR #790, opened 2022-01-21
Support custom error message when throwing 422

nem035:
Just a suggestion for something I'd find useful. Feel free to close if not appropriate.

If it's something worth adding, I'll clean up the PR, add a test, etc.

### DIFF
--- a/packages/http-json-body-parser/index.d.ts
+++ b/packages/http-json-body-parser/index.d.ts
@@ -1,7 +1,8 @@
 import middy from '@middy/core'
 
 interface Options {
-  reviver?: (key: string, value: any) => any
+  reviver?: (key: string, value: any) => any,
+  rejectMsg?: string
 }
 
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for the PR. I'm going to have to say no because if we update to allow in one middleware to allow error message overriding then we'd have to update all middleware to also support. Some other middleware have multiple errors and the options could get complicated. However, in this case it might make sense to update the message to be `e.message` (ie `Unexpected token m in JSON at position 0`).

Can you share your use case on why you would like a custom message? Cleaner message or trying to catch this specific error to handle in a special way?

nem035:
Sure thing, makes sense. My use case was to use a message in line with some error messaging conventions.

willfarrell:
Can you share those conventions here? I'm working on v3 and open to improving error messages

PR #792, opened 2022-01-21
Fix: possible to omit paths on already stringified object from request/response

AuxB:
The fix allows parsing objects previously stringify in the response object onto `omitPaths` option input/output middleware. 
For this, a new function named `replaceStringifyObjectValue` was created, which try to parse a value if it is a string. This function is called in the replacer option of JSON.Stringify() just before building a logger message

No

The problem was spotted when we wanted to omit the values body of an HTTP endpoint but there already stringified so it is not possible to omit them. 
Please look at the added test, It is normally quite understandable, thanks!

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Node.js Versions:** v14.17.6

**Middy Versions:** v2.5.2

**AWS SDK Versions:** v2.897.0

Todo list
---------

[✅] Feature/Fix fully implemented
[✅] Added tests
[❌] Updated relevant documentation
[❌] Updated relevant examples

### DIFF
--- a/packages/input-output-logger/__tests__/index.js
+++ b/packages/input-output-logger/__tests__/index.js
@@ -56,6 +56,36 @@ test('It should omit paths', async (t) => {
 
   t.deepEqual(response, { message: 'hello world', bar: 'bi' })
 })
+
+test('It should omit paths and not stringify twice an existed stringify object', async (t) => {
+  const logger = sinon.spy()
+
+  const handler = middy((event, context) => {
+    return {
+      message: 'hello world',
+      bar: JSON.stringify({ fuu: 'bur', duu: 'bir' })
+    }
+  })
+
+  handler.use(
+    inputOutputLogger({ logger, omitPaths: ['event.foo', 'response.bar.fuu'] })
+  )
+
+  const response = await handler({ foo: 'bar', fuu: 'baz' })
+  t.log(logger)
+  t.true(logger.calledWith({ event: { fuu: 'baz' } }))
+  t.true(
+    logger.calledWith({
+      response: { message: 'hello world', bar: { duu: 'bir' } }
+    })
+  )
+
+  t.deepEqual(response, {
+    message: 'hello world',
+    bar: JSON.stringify({ fuu: 'bur', duu: 'bir' })
+  })
+})
+
 test('It should skip paths that do not exist', async (t) => {
   const logger = sinon.spy()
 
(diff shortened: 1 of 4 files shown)


willfarrell:
I like the idea, I'm worried about the possible performance hit this might introduce (I'm working on benchmarking test lately). For sure will have to be a breaking change. Can you share your use case to why this is needed?

1. Can you update the PR to be off of `feature/3.x`
2. `@middy/util` has `jsonSafeParse`, I'd prefer using that over including your own.

AuxB:
> Thanks for the PR!
> 
> I like the idea, I'm worried about the possible performance hit this might introduce (I'm working on benchmarking test lately). For sure will have to be a breaking change.
> 
> 1. Can you update the PR to be off of `feature/3.x`
> 2. `@middy/util` has `jsonSafeParse`, I'd prefer using that over including your own.

Thanks for your feedback !

For performance, I benchmark fastly with time command and there is nos difference but I make it onto the little object. We can imagine optionally putting the replacer on the JSON.stringify through `omitPaths` option.

By the way, I am eager to see your benchmark test !

I make the change as soon as possible

willfarrell:
Actually, you can already do this by setting `options.reviver`. Maybe we just need to update the docs with this example and  include in the tests? what do you think?

benchmarks are on the `release/3.x` branch. :) `npm run test:packages:benchmark`

AuxB:
I'm not sure to understand correctly. You want to use `options.reviver` to let more flexibility for the user on the parsing object?
My point was to say to avoid deep parsing if the user won't to.

Good job for benchmark test, nice feature !

By the way I update the branch :)

willfarrell:
Sorry, brain is a little scattered after this week. I was looking at `http-json-body-parser` that uses `reviver` with `JSON.parse()`. A similar pattern can be applied here, using `replacer` for `JSON.stringify()`.

```
const redactedMessage = omit(JSON.parse(JSON.stringify(message, options.replacer)), omitPaths)
```

You've got me thinking, could replacer be used to speed up the omit logic ...

willfarrell:
I went in and did an extensive refactor and included `replacer` while I was there (`release/3.x`). Let me know what you think. If you'd like to add an example use case to the README, that would be great.

AuxB:
Hello @willfarrell, 

Sorry for the late answer, I was little busy this last weeks.

Great news if you can integrate replace on the next release; when that will release ? Thanks !

willfarrell:
It should already be released as part of `3.0.0-alpha.x` if you'd like to test it out. I hope to have a beta out next month.

PR #808, opened 2022-02-16
SSM module / Fix object like values failing to set into the env

randomhash:
Hey! Would like to add option to not parse value ;)

An object like values are added to the environment as the string "[object Object]"
This solution is to add more options for the default behavior with the usage of the safeJsonParse.
The default behavior will not change

Nope

Test should cover it
-------------------

[ x ] Feature/Fix fully implemented
[ x ] Added tests
[ x ] Updated relevant documentation
[ x ] Updated relevant examples


willfarrell:
Thanks for putting this PR together. Unfortunately, `setToEnv` will be deprecated in v3. Is there a use case for this outside the use with `setToEnv`?

If you want to manually set to the env, and stringify, checkout: https://github.com/middyjs/middy/blob/release/3.x/docs/UPGRADE.md#notes

randomhash:
Hello! Thanks, no other use case
When is 3.0 planned for release?

randomhash:
@willfarrell why setToEnv is deprecated? it is safer in cases of proxy usage of events and in terms of lambda environments. I don't think this is a good decision in a long run
I can see login in common monolithic systems, but not for lambdas

PR #816, opened 2022-03-19
Support both cjs and esm in packages

willfarrell:
- tests esm on both nodejs v14 and v16, testing cjs not feasible
- transpiles cjs and esm es2020 (nodejs v14)

## Questions
- Should esm export as `mjs`? If so, leave the original js file in?


lmammino (approved):

lmammino:
I am not sure if there is any value in publishing also .mjs files since we are specifying `{"type":"module"}` in all our `package.json` files...

PS: the benchmarks seem to be broken

PR #844, opened 2022-05-23
fix: import DynamoDB from straight from aws-sdk

ifeltsweet:
Before this change if you had Webpack configured to not bundle `aws-sdk` via `externals` setting it would still bundle it because the import was done via `aws-sdk/clients/dynamodb.js` instead of `aws-sdk`. This is very unexpected behavior for most setups and would require someone figuring out why `aws-sdk` is getting bundled to spend many hours debugging.

### DIFF
--- a/packages/event-normalizer/README.md
+++ b/packages/event-normalizer/README.md
@@ -118,6 +118,34 @@ const lambdaHandler = (event, context) => {
 const handler = middy(lambdaHandler).use(eventNormalizer())
 ```
 
+## AWS SDK
+
+This middleware depends on `aws-sdk` package. Lambda runtime already includes `aws-sdk` by default and as such you normally don't need to package it in your function. If you are using Webpack and Serverless Framework to package your code you'd want to add `aws-sdk` to Webpack `externals` and to `forceExclude` of Serverless Framework Webpack configuration.
+
+1. Tell Webpack not to bundle `aws-sdk` into the output (e.g. handler.js):
+```
+# webpack.config.js
+
+var nodeExternals = require("webpack-node-externals");
+
+module.exports = {
+  externals: ["aws-sdk"],
+};
+```
+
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for opening a PR. Unfortunately a can’t approve this PR because would require every cold start to evaluate the entire sdk. I would be open to a PR that adds a note to the documentation letting others know about this gotcha.

ifeltsweet:
Thanks, I understand. I will update the PR with documentation changes to this package.

I must admit it is a little bit confusing why `aws-sdk` is listed as `devDependency` if the assumption is that `aws-sdk` is going to be packaged by users of this library by default (according to your comment):

```
Thanks for opening a PR. Unfortunately a can’t approve this PR because would require every cold start to evaluate the entire sdk. I would be open to a PR that adds a note to the documentation letting others know about this gotcha.
```

If it's listed as `devDependency` then I guess the automatic assumption should be that users of this library by default don't package or bundle `aws-sdk` and instead rely on the one provided by Lambda runtime.

PS. I noticed that other packages in this repository also follow the pattern of not importing straight from `aws-sdk`.

ifeltsweet:
@willfarrell I may have misunderstood. You can accept this PR if documentation is added about adding `aws-sdk` to Webpack `externals`? Or you would accept a different PR that only adds a note to documentation that notifies users to add `aws-sdk/clients/dynamodb.js` to Webpack `externals`?

BTW, I am not familiar how this plays out with packaging strategies other than Webpack.

ifeltsweet:
@willfarrell  I've added the notice to this PR. If you prefer to go the route of only updating the docs please let me know, however, I do feel that it's not an intuitive way for most setups.

Of course, I could be missing something and I do not know how most people package their functions.

PS: instructions are heavily inspired by https://docs.datadoghq.com/serverless/guide/serverless_tracing_and_webpack/.

willfarrell:
I copied these changes into the documentation. We will not be using `... from 'aws-sdk'` for cold start performance reasons.

PR #845, opened 2022-05-27
fix(http-error-handler): type error in 'logger' option

tzkz:
Fix error `Type 'boolean' is not assignable to type '(error: any) => void'.`

According to [readme](https://github.com/middyjs/middy/tree/main/packages/http-error-handler#options), "You can pass `false`" to the `logger` option


willfarrell (approved):

willfarrell:
Thanks for the PR! I'm away from home, will not be able to release for a week or so.

PR #851, opened 2022-06-07
fix: export route and method for middy http router

mihilmy:
Struggled to type the `ROUTES` array mainatined in a configuration file that gets imported for the httpRouter


willfarrell:
Thanks for the PR!

PR #854, opened 2022-06-11
fix(type): add `handler` in the `MiddyfiedHandler`

lucduong:
In the middy 3, we proposed the `.handler` function but not mentioned in the type for `TypeScript`. This change is adding to that.


willfarrell:
Thanks for the PR! Looks good.

PR #859, opened 2022-06-13
feat: improve s3 object response options

Puppo:
Improve s3 object response options


willfarrell:
Thanks for all the TS improvements. I can push a release at the end of the week, just in case there are any more you'd like to have included.

PR #860, opened 2022-06-13
docs: improvement

Puppo:
- Add filteringKeyName options into http-partial-response
- Fix namespace typo


willfarrell:
Awesome stuff!

PR #861, opened 2022-06-14
Fix TS Error in httpRouterHandler

vinceniko:
httpRouterHandler should return a MiddyfiedHandler to be type-checked correctly


vinceniko:
Same issues exist for the ws-router. i can file a separate pull request for those fixes

willfarrell:
Thanks for the PR!

PR #867, opened 2022-06-22
Fix TS Error in wsRouterHandler 

vinceniko:
Same as #861 but for wsRouterHandler


vinceniko:
Closed because this was fixed already. Ignore this pull request

PR #875, opened 2022-07-14
Change GitHub link in documentation from docusaurus to middy

AlessandroVol23:
The GitHub link in the footer wasn't changed and still pointed to the docusaurus repository instead of the middy repository.


willfarrell:
Thanks for the PR!

PR #878, opened 2022-07-23
add missing `optional chaining operators`

zvictor:
I get the error below when I invoke lambda functions without headers defined.

```haskell
ERROR	TypeError: Cannot read properties of undefined (reading 'accept')
job-crawler-container-dfe899e[0]     at null.httpResponseSerializerMiddlewareAfter (/node_modules/.pnpm/@middy+http-response-serializer@3.1.0/node_modules/@middy/http-response-serializer/index.js:21:89)
```

The error goes away when _optional chaining operators_ are added. Beware that _nullish coalescing operators_ are not a replacement of the _optional chaining operators_ and can be used together, as shown by MDN: 

* [Relationship with the optional chaining operator (?.)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator#relationship_with_the_optional_chaining_operator)

### DIFF
--- a/packages/http-response-serializer/index.js
+++ b/packages/http-response-serializer/index.js
@@ -21,7 +21,7 @@ const httpResponseSerializerMiddleware = (opts = {}) => {
       types = [request.event.requiredContentType]
     } else {
       const acceptHeader =
-        request.event.headers.Accept ?? request.event.headers.accept
+        request.event.headers?.Accept ?? request.event.headers?.accept
       types = [
         ...((acceptHeader && Accept.mediaTypes(acceptHeader)) ?? []),
         request.event.preferredContentType,


willfarrell:
Thanks for opening a PR. Can you share the AWS event that caused this error. All AWS HTTP events include headers to my knowledge, or at least I thought. Optional chaining when transpiled is bloated and slow which is why I only use it when necessary.

zvictor:
I am running [lambda inside docker](https://github.com/lambci/docker-lambda#running-lambda-functions).

This is the request:

```haskell
curl -d '{}' http://localhost:9001/2015-03-31/functions/myfunction/invocations
```

willfarrell:
`-d '{}'` is the issue, AWS would never send an event looking like that. `http-response-serializer` requires the accept header to know which mime type to return. If you're using docker to run tests, tests should have an expected event. Additionally I would recommend that you validate all inputs into your lambdas as good security practice. Having this in place would catch this issue earlier and with a better error message. I've updated to the docs to mention input validation.

zvictor:
> I've updated to the docs to mention input validation.

Could you be specific? I couldn't find any recent change to the docs.

I would like to learn and change my scripts to make them behave like a standard AWS call.

willfarrell:
Sorry, forgot to push commit. I added a comment to the bottom of https://middy.js.org/docs/middlewares/http-response-serializer#content-type-negotiation section. Should say: "When planning to use `Accept`, an external input, it is recommended to validate that it is an expected value.", in case the page is cached for you.

Checkout [`@middy/validator`](https://middy.js.org/docs/middlewares/validator) to be able to add header validation to your event requests.

Issue #769, opened 2022-01-05
Version 4


willfarrell:
PR: #908 

## Planned
- [x] Drop Node v14, Add support fro Node v18
  - [x] Update `engines` to `"node": ">=16"`
  - [x] Update github workflows `node-version`
  - [x] Update `swc` to use `es2021` (https://github.com/tsconfig/bases/tree/main/bases)
- [x] `event-normalizer`: review for any new events from [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-services.html) and [event list](https://github.com/michaelbrewer/aws-lambda-events)
- [x] `http-security-headers`: review against helmet, MDN, ASVS
- [x] AWS SDK v3 #584
  - [x] Update packages
  - [x] Update tests
  - [x] Update docs
  - [x] TypeScript
  - [x] Update to use async iterators for pagination in `ssm`, `secrets-manager`, etc (https://aws.amazon.com/blogs/developer/pagination-using-async-iterators-in-modular-aws-sdk-for-javascript/)
- [x] `http-urlencode-body-parser` throw 422 when unable to parse
- [x] `http-content-encoding` pull stream piping into new middleware `stream-transform-response`
- [x] `http-router` parse path params. no, can be done in v3
- [x] `http-router` REST and HTTP route paths should be the same (https://github.com/middyjs/middy/issues/873). No, see issue.
- [x] `validator` export `compileSchema` and require it be passed in to allow for better tree-shaking
- [x] Allow top level await within middy core when adding middleware `.use(await middleware())`
- [x] Update bundling docs
 
### >= 15
- [x] use replaceAll in `event-normalizer` [skip - slower than current]

#### remove polyfills
- [x] `core`
  - [x] Remove `AbortControler` polyfill
  - [x] remove `timers/promises` polyfill
- [x] `http-content-encoding`: remove `stream/promises` polyfill
- [x] `util` update tests to use `timers/promises` for delay

#### Error {cause} >= 16.9
- [x] `util` update `createError` to use `class HttpError extends Error`
- [x] `http-json-body-parser`: use proper `new Error(message, {cause})` notation
- [x] `http-multipart-body-parser`: use proper `new Error(message, {cause})` notation
- [x] `util`: use proper `new Error(message, {cause})` notation
- [x] `util`: use proper `super(message, {cause})` notation in `HttpError`
- [x] `validator`: use proper `new Error(message, {cause})` notation

## deprecate
[...]

PR #883, opened 2022-07-31
feat: self-composing UseFN

vicary:
In respond to https://github.com/middyjs/middy/issues/769#issuecomment-1196279277, this is the first attempt to make the event object self- and auto- composing through the chain of `.use()`.

Intended usage:
```ts
import type middy from "@middy/core";
import type { Context } from "aws-lambda";

const middyFoo = (): middy.MiddlewareObj<{ evtFoo: string }, any, Error, Context & { ctxFoo: string }> => ({
  before: ({ context, event }) => {
    context.ctxFoo = "foo";
    event.evtFoo = "foo";
  },
});

const middyBar = (): middy.MiddlewareObj<{ evtBar: string }, any, Error, Context & { ctxBar: string }> => ({
  before: ({ context, event }) => {
    context.ctxBar = "bar";
    event.evtBar = "bar";
  },
});


export const handler = middy()
  .use(middyFoo())
  .use(middyBar())
  .before(({ context: { ctxFoo, ctxBar }, event: { evtFoo, evtBar } }) => {
    // context and event inherits new properties from the .use() above
  })
  .handler(({ evtFoo, evtBar }, { ctxFoo, ctxBar }) => {
    // context and event inherits new properties from the .use() above
  });
```

### DIFF
--- a/packages/cloudwatch-metrics/index.d.ts
+++ b/packages/cloudwatch-metrics/index.d.ts
@@ -1,6 +1,6 @@
 import middy from '@middy/core'
 import { MetricsLogger } from 'aws-embedded-metrics'
-import type { Context } from 'aws-lambda'
+import type { Context as LambdaContext } from 'aws-lambda'
 export { MetricsLogger } from 'aws-embedded-metrics'
 
 interface Options {
@@ -8,13 +8,12 @@ interface Options {
   dimensions?: Array<Record<string, string>>
 }
 
+export type Context = LambdaContext & {
+  metrics: MetricsLogger
+}
+
 declare function cloudwatchMetrics(
   options?: Options
-): middy.MiddlewareObj<
-  unknown,
-  any,
-  Error,
-  Context & { metrics: MetricsLogger }
->
+): middy.MiddlewareObj<unknown, any, Error, Context>
 
 export default cloudwatchMetrics
--- a/packages/cloudwatch-metrics/index.d.ts
+++ b/packages/cloudwatch-metrics/index.d.ts
@@ -1,4 +1,6 @@
 import middy from '@middy/core'
+import { MetricsLogger } from 'aws-embedded-metrics'
+import type { Context } from 'aws-lambda'
 export { MetricsLogger } from 'aws-embedded-metrics'
 
 interface Options {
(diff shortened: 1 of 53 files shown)


vicary:
@willfarrell Should it be 3.0 or 4.0?

Also some old plugins add something to lambda context, while others add to events. I am kinda leaning towards extending the event object, so I intentionally left `TContext` untouched. Do you have a preference or encourage either side?

willfarrell:
v4 is on hold till we know what AWS is going to do for their next runtime update (v18). That could be as early as Sept, or a year from now. I'm open to minor changes for v3 for now. There can also be a separate PR for v4, to allow others to comment.

Changes to the event are the most common use in middlewares, however non-external inputs (ie secrets, db connections, etc) should live in the context.

I personally don't use TS, so it's up to the community to decide what is best for Middy. This PR looks safe fro v3.

vicary:
My team is currently adding everything to event, including `typeorm` and `fetch`, to make handlers testable.

But I want this PR to be generally align with your advertised usage, I'll update this PR again to have `TContext` also inherits through the chain of `.use()`.

vicary:
@willfarrell It's ready now.

Tried avoiding the necessity of `Context & ...` in usage, but that breaks the `extends LambdaContext` check. I chose to retain a more guaranteed typing.

Usage example in the description is updated to reflect the change.

vicary:
@willfarrell It's ready now.

Tried avoiding the necessity of `Context & ...` in usage, but that breaks the `extends LambdaContext` check. I chose to retain a more guaranteed typing.

Usage example in the description is updated to reflect the change.

vicary:
We may update the types in all packages, shall I do that here or in a new PR?

For example, `@middy/cloudwatch-metrics` with typings:

```ts
export const handler = middy()
  .use(cloudwatchMetrics())
  .handler(({}, { metrics }) => {
    // The `metrics` property and its methods will be discoverable by IDEs
  });
```

willfarrell:
Within this PR is fine.

vicary:
TypeScript has no way to perform string matching without introducing unused vars, I have to skip these errors.

https://github.com/middyjs/middy/blob/09937572e16d6fa5802e642439e972828a3e49aa/packages/ssm/index.d.ts#L10-L15

vicary:
Took another look at SSM and noticed a mistype, fixed it.

**Sidenote**

`ts-standard` (`npm run test:lint:ci`) is not suitable for typescript linting because there are broken rules, for example https://github.com/typescript-eslint/typescript-eslint/issues/1824.

I think we should use `eslint` directly so we can opt-in and out of particular rules, along with `eslint-plugin-prettier` so contributors can easily fix trivial format misalignments.

That's for another PR though, I'll stick with the broken indentations here to stay green.

willfarrell (approved):

willfarrell:
Looks good to me. I'm going to leave it open in case other would like to comment. I can merge and release next week. Thanks for all of the hard work putting this together.

PR #885, opened 2022-08-16
Update http-response-serializer default option to defaultContentType

harish-lifex:
In the new version of http-response-serializer middleware  the `default` option is renamed to `defaultContentType`


willfarrell:
Thanks for the PR!

PR #888, opened 2022-08-20
Improve typings for security-headers middleware

otahontas:
Hi! 

I started recently using security-headers with typescript and noticed that type definitions don't quite match the javascript implementation. Here are some fixes that allow typescript usage without using `// @ts-ignore`. 

Basically this allows typescript users to pass in false to skip setting header and adds some recently added headers.


willfarrell:
Thanks for the PR!

PR #893, opened 2022-08-26
Feature: http-error-handler to support error headers

### LINKED ISSUE #892: `http-error-handler` does not respect headers set for an HTTP Error
**Describe the bug**
`createError` function from the [http-errors](https://npm.im/http-errors) package allows setting HTTP headers to be sent to the client. `http-error-handler` however, ignores those.

**To Reproduce**
```ts
async function getTheThing() {
  throw createError(404, { message: 'Failed to get the thing', headers: { foo: 'bar' } });
}

async function theThingHandler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
  const theThing = await getTheThing();

  return {
    statusCode: StatusCodes.HTTP_STATUS_OK,
    body: JSON.stringify(theThing),
  };
}

const handler = middy(theThingHandler).use(httpErrorHandler());
```
**Current behaviour**

```JSON
{"Content-Type": "application/json"}
```

**Expected behaviour**
```JSON
{"Content-Type": "application/json", "foo": "bar"}
```

 - Node.js: 14.18.2
 - Middy: 3.2.2
 - AWS SDK: 2.1097.0

willfarrell:
@squ1b3r

Closes #892

### DIFF
--- a/packages/http-error-handler/__tests__/index.js
+++ b/packages/http-error-handler/__tests__/index.js
@@ -99,6 +99,29 @@ test('It should be possible to pass a custom logger function', async (t) => {
   t.true(logger.calledWith(expectedError))
 })
 
+test('It should be possible to pass in headers with error', async (t) => {
+  const handler = middy(() => {
+    const error = createError(422, 'Unprocessable Entity')
+    error.headers = {
+      Location: 'http://exmaple.org/500'
+    }
+    throw error
+  })
+
+  handler.use(httpErrorHandler({ logger: false }))
+
+  const response = await handler(null, context)
+
+  t.deepEqual(response, {
+    statusCode: 422,
+    body: 'Unprocessable Entity',
+    headers: {
+      'Content-Type': 'text/plain',
+      Location: 'http://exmaple.org/500'
+    }
+  })
+})
+
 test('It should create a response for HTTP errors created with a generic error', async (t) => {
   const handler = middy(() => {
     const err = new Error('A server error')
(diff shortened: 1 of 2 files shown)


packages/http-error-handler/__tests__/index.js
```diff
@@ -99,6 +99,29 @@ test('It should be possible to pass a custom logger function', async (t) => {
   t.true(logger.calledWith(expectedError))
 })
 
+test('It should be possible to pass in headers with error', async (t) => {
+  const handler = middy(() => {
+    const error = createError(422, 'Unprocessable Entity')
+    error.headers = {
```
squ1b3r, line 105: You can also pass headers to the `createError` like this:
```ts
createError(424, { message: 'Unprocessable Entity', headers: { Location: 'http://exmaple.org/500' } });
```
but this also works.

squ1b3r (approved):
LGTM! Thanks!

packages/http-error-handler/__tests__/index.js
willfarrell: The `createError` function here is a slimmed down version built into middy. It only has the functionality needed for the middlewares.

squ1b3r (approved):

squ1b3r:
@willfarrell once again, thank you for fixing this so quickly. This is going to save me a lot of time and pain. When do you think you can release this?

willfarrell:
I'll try for later today. Before Monday for sure.

PR #897, opened 2022-09-05
docs: replace typo `default` in sample with `defaultContentType`

lukebrowne:
There's a typo in the sample usage that trips up first time users of the `@middy/http-response-serializer`  so updated the top example to be consistent with the others.


willfarrell:
Thanks for the PR!

Issue #899, opened 2022-09-13
Http Router - multiple variables causes issues for the path in the regex


yannickrocks:
**Describe the bug**
In the Http-Router package, the following will not resolve correctly to create dyanmic routes

E.g. 
Imagine you have 2 function URL with the following paths:

- `/varA/{varA}/varB/{varB}`
- `/varA/{varA}/varC/{varC}`

The line of code 
`path = path.replace(regexpDynamicWildcards, '/?.*').replace(regexpDynamicParameters, '/.+')`
 will resolve both paths to `/^\/varA\/.+$/`

This is an issue as if you are trying to use the second path, it will use the first path's parameters which can cause issues further down the line.

**THE FIX**

change the following

*OLD*
```
const regexpDynamicParameters = /\/\{.+\}/g;
```

*NEW*
```
const regexpDynamicParameters = /\/\{[^\/]+}/g;
```

Regex helper: https://regex101.com/r/Xk1x6P/1

yannickrocks:
@willfarrell Working on it today. 

We also have found a way to populate the path parameters if that is something people would be interested in.

yannickrocks:
PR is here  @willfarrell 
https://github.com/middyjs/middy/pull/901

willfarrell:
> We also have found a way to populate the path parameters if that is something people would be interested in.

Can you open a new issue for this? 

When used with API Gateway the pathParameters are pre-parsed which is why it's not included currently. This also giving a perf boost compared to other routers. Are you using this with ALB or Function URLs? Would love to heard more about your use case.

yannickrocks:
HI @willfarrell 
Sure thing, We are using Function URLs to allow lambdas to call other lambdas internally. :D

PR #901, opened 2022-09-14
fix(http-router): change the dynamic regex to pick up all variables

yannickrocks:
resolves https://github.com/middyjs/middy/issues/899


willfarrell:
Thanks for the PR!

PR #912, opened 2022-10-03
validator: break out compileSchema

willfarrell:


willfarrell:
@lmammino Would love a quick review if you can.

Issue #909, opened 2022-09-28
Unable to use custom logger for @middy/sqs-partial-batch-failure


iaxentoi:
**Describe the bug**
The default logger of the sqs-partial-batch-failure middleware is not logging errors with the proper level for us.
We want to pass a custom logger for the middleware, but we get a type error.

**To Reproduce**
How to reproduce the behaviour:
Sample code 
```
  import requeueFailedMessages from '@middy/sqs-partial-batch-failure'
  export const handler = middleware(myHandler)
    .use(requeueFailedMessages({logger: console.error}))
```
The following error occurs:
`Argument of type '{ logger: { (...data: any[]): void; (message?: any, ...optionalParams: any[]): void; }; }' is not assignable to parameter of type 'Options<SQS>'.
  Object literal may only specify known properties, and 'logger' does not exist in type 'Options<SQS>'.`

**Expected behaviour**
We should be able to use a custom logger as specified in the docs: https://www.npmjs.com/package/@middy/http-error-handler#options

**Environment (please complete the following information):**
 - Node.js: [14.17.6]
 - Middy: [core: 3.1.0, sqs-partial-batch-failure: 3.1.0]
 - AWS SDK [2.1142.0]

**Additional context**

willfarrell:
It looks like the TypeScript definition is out of date. Would you like to submit a PR to resolve this?

david-preece:
Since we didn't see any progress on the issue and I am now facing it, I have taken the liberty of creating the PR myself.

PR #913, opened 2022-10-03
Update type definitions for Options in sqs-partial-batch-failure

david-preece:
Fixes https://github.com/middyjs/middy/issues/909 by updating the type definitions for `Options` in `packages/sqs-partial-batch-failure/index.d.ts`

```typescript
interface Options {
  logger?: (reason: any, record: any) => void
}
```


david-preece:
I have updated this PR with the updated documentation to reflect the updated code.

Issue #905, opened 2022-09-21
Lambda Handler type is not up-to-date


eranelbaz:
**Describe the bug**
with middy v3 and the new way to handle [timeouts using AbortController](https://middy.js.org/docs/writing-middlewares/timeouts)
the lambda handler function signature has been changed to accept `{ signal }`, but, middy accepts lambda handler that accepts `Callback` as the third argument to the function - https://github.com/middyjs/middy/blob/8f4a4fec7cdd827441ed33e2b20611f5518f2767/packages/core/index.d.ts#L77

**To Reproduce**
How to reproduce the behaviour:
1. Sample code - [middy example](https://middy.js.org/docs/writing-middlewares/timeouts) with the above middy definition to accept `LambdaHandler` where the third argument is `Callback<Result>`

**Expected behaviour**
Types should be up to date and specify the relevant parameters handler function should accept with the new abort controller behaviour

**Environment (please complete the following information):**
 - Node.js: 16.10.0
 - Middy: middy core 3.3.4
 - AWS SDK (`@types/aws-lambda`) 8.10.104

**Additional context**

eranelbaz:
Hey @willfarrell 
we'll open one in the near future

willfarrell:
Just an FYI, we're getting close to a first alpha for middy v4. Once that happens, only security patches will be applied to v3.

willfarrell:
Closing due to lack in interest in implementing. Please re-open if able to submit a PR.

PR #922, opened 2022-10-11
refactor(core/index.d.ts): add internal middy params as third lambda parameter

naorpeled:


naorpeled:
Hey @willfarrell,
I think we have two ways to resolve #905.

1) Move signal to `event` or `context` to not introduce any breaking type definition changes.
2) No longer allow to use `@types/aws-lambda` Handler type because it requires the third parameter to be of type callback and overriding it is problematic.

Wdyt? Is there something I'm missing, would love to hear your feedback 🙏

willfarrell:
> Move signal to event or context to not introduce any breaking type definition changes.
This is a no go, a) an unexpected pattern, b) is a breaking change itself

> No longer allow to use @types/aws-lambda Handler type because it requires the third parameter to be of type callback and overriding it is problematic.

Looking at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts could we not define our own Handler and import Context?

```
import {Context} from '@types/aws-lambda'
export type Handler<TEvent = any, TResult = any> = (
    event: TEvent,
    context: Context,
    abort: { signal: AbortControllerSignal},
) => void | Promise<TResult>;
```

naorpeled:
> > Move signal to event or context to not introduce any breaking type definition changes.
> > This is a no go, a) an unexpected pattern, b) is a breaking change itself
> 
> > No longer allow to use @types/aws-lambda Handler type because it requires the third parameter to be of type callback and overriding it is problematic.
> 
> Looking at [DefinitelyTyped/DefinitelyTyped@`master`/types/aws-lambda/handler.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts?rgh-link-date=2022-10-12T01%3A50%3A39Z) could we not define our own Handler and import Context?
> 
> ```
> import {Context} from '@types/aws-lambda'
> export type Handler<TEvent = any, TResult = any> = (
>     event: TEvent,
>     context: Context,
>     abort: { signal: AbortControllerSignal},
> ) => void | Promise<TResult>;
> ```

that's what I tried to do in this PR but it won't work because the third parameter conflicts with `@types/aws-lambda` Handler definition.

Callback's shape is 
```typescript
	 (error: any, result: T) => any
```

and our parameter's shape is:
```typescript
{
  signal: AbortSignal
}
```

willfarrell:
I don't know TypeScript, but this approach can't work? https://github.com/middyjs/middy/commit/7aec3f6cd39416e596b3a19f5c1e6be0d376aa22

Redefining `LambdaHander` and overriding `callback`. `callback` isn't used at all in middy

naorpeled:
> I don't know TypeScript, but this approach can't work? [`feature` (compare)](https://github.com/middyjs/middy/compare/feature/abortable-handler)
> 
> Redefining `LambdaHander` and overriding `callback`. `callback` isn't used at all in middy
> 
> If we remove callback, it'll have to be in v4 then

No, doesn't seem like there's a TypeScript workaround for it.

Issue #932, opened 2022-11-10
The default write format value if the content is not base64encoded causes images to upload with error


rodrigomf24:
**Describe the bug**
Using the `htmlMultipartBodyParser` for handling image uploads on a lambda properly parses the file, but if the file is an image the content has errors. I noticed that the default data type in this line https://github.com/middyjs/middy/blob/8318e22b7e0de530443a5b48fc54e152790b2c04/packages/http-multipart-body-parser/index.js#L95 is utf8, this is causing the image to have errors, if I swap it for binary it works. It would be great if this can be configurable.

**To Reproduce**
How to reproduce the behavior:

- create a serverless lambda with middy and inject the `htmlMultipartBodyParser` middleware
- in your lambda store the file to a temp folder and notice the image contains errors

**Expected behavior**
- the uploaded image file should be parsed without errors

**Environment (please complete the following information):**
 - Node.js: [e.g. 16]
 - Middy: [e.g. 3.0.0]
 - AWS SDK [e.g. 2.1000.0]

**Additional context**

willfarrell:
Thanks for reporting. It used to be `binary` until someone filed a bug and became a breaking change in v3. I'd be open to a PR to the v4 branch to add in an option to override `utf8`.

willfarrell:
New option added to the v4 branch. Big thanks to @salesh for adding in this functionality.

PR #934, opened 2022-11-12
feat: add optional charset option to multipart body parser

salesh:
Related to #932

This seemed like something I could contribute to, please let me know if I missed anything

### DIFF
--- a/packages/http-multipart-body-parser/__tests__/index.js
+++ b/packages/http-multipart-body-parser/__tests__/index.js
@@ -292,22 +292,3 @@ test('It should parse a field with multiple files successfully', async (t) => {
   t.true(Object.keys(response).includes('files'))
   t.is(response.files.length, 3)
 })
-
-test('It should parse an array from a multipart/form-data request en dash in binary when charset is passed', async (t) => {
-  const handler = middy((event, context) => {
-    return event.body // propagates the body as a response
-  })
-
-  handler.use(httpMultipartBodyParser({ charset: 'binary' }))
-
-  const event = {
-    headers: {
-      'Content-Type': 'multipart/form-data; boundary=TEST'
-    },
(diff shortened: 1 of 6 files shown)


packages/http-multipart-body-parser/__tests__/index.js
```diff
@@ -292,22 +292,3 @@ test('It should parse a field with multiple files successfully', async (t) => {
...
-    headers: {
-      'Content-Type': 'multipart/form-data; boundary=TEST'
-    },
-    body: '--TEST\r\nContent-Disposition: form-data; name=PartName\r\nContent-Type: application/json; charset=utf-8\r\n\r\n{"foo":"bar–"}\r\n--TEST--',
-    isBase64Encoded: false
-  }
-  const response = await handler(event, defaultContext)
-
-  t.deepEqual(response, { PartName: '{"foo":"bar–"}' })
-})
```
salesh: this test was completely wrong, I am trying to figure out more correct example - any help would be appreciated

willfarrell:
Thanks for the PR. A test when `binary` is passed in will be needed. See the v2 branch when binary was the default for a test that can be copied in. https://github.com/middyjs/middy/blob/2.x/packages/http-multipart-body-parser/__tests__/index.js

salesh:
> Thanks for the PR. A test when `binary` is passed in will be needed. See the v2 branch when binary was the default for a test that can be copied in. https://github.com/middyjs/middy/blob/2.x/packages/http-multipart-body-parser/__tests__/index.js

That test has an issue

```
  It should parse an array from a multipart/form-data request (binary)

  Rejected promise returned by test. Reason:

  UnprocessableEntityError (HttpError) {
    expose: true,
    status: 422,
    statusCode: 422,
    message: 'Invalid or malformed multipart/form-data was provided',
  }
```

how I understand it - an empty body is the root of the issue for that one

Issue #933, opened 2022-11-12
Expose `Request` Typescript interface from core types


saintybalboa:
**Is your feature request related to a problem? Please describe.**

I need to assign a type to function arguments of the middleware object (`MiddlewareObj`). Digging through the middy [codebase](https://github.com/middyjs/middy/blob/main/packages/core/index.d.ts#L24) I found the `Request` interface which seems to be what I need but it isn't exported so cannot be used.

**Describe the solution you'd like**

Export the [Request](https://github.com/middyjs/middy/blob/main/packages/core/index.d.ts#L24) interface so that consumers can assign the type to the middleware functions.

```ts
const middlewareObj = {
    before: (request: Request) => {
      ...
    }
}
```

**Describe alternatives you've considered**

If there's already a type available for this it might be worth adding this to the [docs](https://middy.js.org/docs/writing-middlewares/with-typescript/).

**Additional context**

N/A

willfarrell:
Would you be able to open a PR against the v4 branch to add this in?

saintybalboa:
@willfarrell here's a link to the [PR](https://github.com/middyjs/middy/pull/935). Let me know if there's anything I've missed / needs updating.

PR #935, opened 2022-11-14
Export the Request interface

saintybalboa:
**Issue:** https://github.com/middyjs/middy/issues/933

## Changes

Exports the [Request](https://github.com/middyjs/middy/blob/main/packages/core/index.d.ts#L24) interface so that consumers can assign the type to the argument in the middleware functions.

```ts
const middlewareObj = {
    before: (request: Request) => {
      ...
    }
}
```


willfarrell:
Thanks for contribution

PR #941, opened 2022-11-23
Allow handler setup to define additional properties on event

CorruptedHeart:
Useful if middleware adds properties to the event.

e.g.
A `authMiddleware` that adds `userId` from a JWT token to the event.

`handler<{userId: string}>(event => event.userId)` <-- userId will not complain in TS.


willfarrell:
Thanks for the PR. will be released soon.

PR #942, opened 2022-11-23
Fix alias name

estrehle:
For `responseSchema`, the readme incorrectly references `inputSchema` as alias. The correct alias is `outputSchema`.


willfarrell:
Thanks for the PR. `responseSchema` is an alias for `outputSchema`. `responseSchema` is deprecated in the next version, releasing in days.

PR #951, opened 2022-12-06
Update 05-bundling.md recommendation on bundling the aws-sdk

carlnordenfelt:
AWS themselves [recommend ](https://docs.aws.amazon.com/lambda/latest/operatorguide/sdks-functions.html)that you always bundle the `aws-sdk` with your project.
Using the build-in sdk should be limited to quick tests and is not suitable for production use.


willfarrell:
Thanks for the PR!

PR #957, opened 2022-12-08
fix: self-managed kafka event source name

nikovirtala:
According to Lambda [documentation](https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html) event source name for self-managed Apache Kafka is `SelfManagedKafka` *not* `aws:SelfManagedKafka`.

I found this by enabling `@middy/event-normalizer` middleware for my project with this event source type and noticed that the event normalizer does not Base64 decode the record value. Made a fork with fixed event source name and now it works as expected.


willfarrell:
Thanks for much for this! Kafka isn't a service I've used yet, so could never do a real test.

Issue #895, opened 2022-09-02
Http-Router Middleware: Will this support APIGatewayProxyHandlerV2?


yannickrocks:
**Is your feature request related to a problem? Please describe.**
I am wanting to use Aws functions URL which support V2 types. Will this middleware be upgraded ? 

The existing code is as follows:
```
import middy from '@middy/core'
import {
  APIGatewayProxyEvent,
  APIGatewayProxyResult,
  Handler as LambdaHandler
} from 'aws-lambda'

export type Method = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE' | 'OPTIONS' | 'ANY'

export interface Route {
  method: Method
  path: string
  handler: LambdaHandler<APIGatewayProxyEvent, APIGatewayProxyResult>
}

declare function httpRouterHandler (routes: Route[]): middy.MiddyfiedHandler

export default httpRouterHandler


```

codyfrisch:
Well the v2 isn't as much an upgrade as it is a variant for the API Gateway for HTTP API, vs the older REST API.

The bigger problem here is the terrible TypeScript support throughout Middy, which is enough reason I am abandoning using it. The example using typescript doesnt even work for creating a middleware, as its type doesnt match what is required for a .use statement.

willfarrell:
@yannickrocks `http-router` does support v2 (https://github.com/middyjs/middy/blob/main/packages/http-router/index.js#L88-L97), however, the type definitions may not as they are fully developed and maintained by the community. A PR to add support is most welcome.

I've replied to @codyfrisch comment over here https://github.com/middyjs/middy/issues/769#issuecomment-1235650192 . I'm happy to review a PR to improve TypeScript support across Middy.

vicary:
@yannickrocks The router is the one I intentionally left alone in my PR, because it requires multiple levels of type inference and I have no good way to defer an easy usage here. I think I'd like your input on how it's supposed to be used for both V1 and V2 users.

This is my best idea, you may improve from here. So, given the updated type definitions below:

```ts
import middy from '@middy/core'
import {
  APIGatewayProxyEvent,
  APIGatewayProxyEventV2,
  APIGatewayProxyResult,
  APIGatewayProxyResultV2,
  Handler as LambdaHandler
} from 'aws-lambda'

export type Method = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE' | 'OPTIONS' | 'ANY'

export interface Route<TEvent> {
  method: Method
[...]

PR #960, opened 2022-12-12
Feature: http-router - Expose Method as a string-enum.

carlnordenfelt:
Addresses [this comment](https://github.com/middyjs/middy/issues/895#issuecomment-1344252536) in #895.

Exposes the `Method` type as a String Enum instead of a union string.
Note that this is a breaking TS change as per the linked comment.

### DIFF
--- a/packages/http-router/index.d.ts
+++ b/packages/http-router/index.d.ts
@@ -7,14 +7,15 @@ import {
   Handler as LambdaHandler
 } from 'aws-lambda'
 
-export type Method =
-  | 'GET'
-  | 'POST'
-  | 'PUT'
-  | 'PATCH'
-  | 'DELETE'
-  | 'OPTIONS'
-  | 'ANY'
+export enum Method {
+  Get = 'GET',
+  Post = 'POST',
+  Put = 'PUT',
+  Patch = 'PATCH',
+  Delete = 'DELETE',
+  Options = 'OPTIONS',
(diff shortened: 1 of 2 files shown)


naorpeled:
LGTM 😎

willfarrell:
Thanks for the review

tamakisquare:
Changing from `type` to `enum` causes an issue. The `Method` enum can no longer be imported for use, because ambient declaration files (`d.ts` files) are not transpiled and thus not converted to JavaScript. @carlnordenfelt

naorpeled:
> Changing from `type` to `enum` causes an issue. The `Method` enum can no longer be imported for use, because ambient declaration files (`d.ts` files) are not transpiled and thus not converted to JavaScript. @carlnordenfelt

Good point and good catch,
@willfarrell let's revert this for now and write somewhere in the docs that in order to specify methods with TS you'd need to use `as const`

willfarrell:
If a change is required, please open a new PR. I'd like to hear from @carlnordenfelt before a change is merged on this.

naorpeled:
> If a change is required, please open a new PR. I'd like to hear from @carlnordenfelt before a change is merged on this.

I gave it some thought and have an idea on how to resolve this!
Will hopefully send the PR asap (tomorrow is a very intense day, I hope I'll be able to get it done though)

PR #968, opened 2022-12-16
make event.pathParameter a full object

tulsidas:
when creating pathParameters in a dynamic route, the resulting object is a `null prototype` object which causes problems for libraries that try to inspect pathParameters using prototype methods

this patch creates pathParameters as a "full fledged" object

### DIFF
--- a/packages/http-router/__tests__/index.js
+++ b/packages/http-router/__tests__/index.js
@@ -156,6 +156,22 @@ test('It should route to a dynamic route (/path/to) with `{proxy+}`', async (t) 
   t.true(response)
 })
 
+test('It should add pathParameters with object prototype', async (t) => {
+  const event = {
+    httpMethod: 'GET',
+    path: '/path/to'
+  }
+  const handler = httpRouter([
+    {
+      method: 'GET',
+      path: '/path/{proxy+}',
+      handler: () => true
+    }
+  ])
+  await handler(event, context)
+  t.truthy(event.pathParameters.__proto__)
+})
+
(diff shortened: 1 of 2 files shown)


willfarrell:
Can you add a test please to show when this happens and returns the new empty object you're proposing.

tulsidas:
@willfarrell added !

willfarrell:
Thanks! That's what I thought the use case was for, but wanted to make sure. Also this test will catch any regressions.

tulsidas:
Hi @willfarrell , one issue I'm having related to this is that the event may already have some `pathParameters` and the code is not updating them with the dynamic parts of the route, for example:

lambda is defined in `serverless.yml` as follows:

```
functions:
  index:
    handler: dist/index.handler
    events:
      - http:
          path: /{proxy+}
          method: any
```

And I have defined a dynamic route `/v1/users/{id}`

Then a request comes to `/v1/users/123`

But the AWS event already has `pathParameters` set, like:

```
{
  "resource": "/{proxy+}",
  "path": "/v1/users/123",
  "httpMethod": "GET",
  "pathParameters": {
    "proxy": "v1/users/123"
  },
  ...
```

So [this line](https://github.com/middyjs/middy/blob/main/packages/http-router/index.js#L46) won't set the required `{id: 123}` value in the `pathParameters` hence the lambda will fail 

Should this be fixed in `http-router`? (I can create a new PR for it) or I am doing something wrong with my configuration?

Thanks

PR #971, opened 2022-12-19
merge existing event.pathParameters with the ones inferred from the dynamic route

tulsidas:
Previous code only wrote `pathParameters` in the event for a dynamic route when existing `pathParameters` property was undefined, this PR fixed that and merges both existing `pathParameters` and the inferred from the dynamic route

### DIFF
--- a/packages/http-router/__tests__/index.js
+++ b/packages/http-router/__tests__/index.js
@@ -162,7 +162,7 @@ test('It should populate pathParameters to a dynamic route even if they already 
     httpMethod: 'GET',
     path: '/user/123',
     pathParameters: {
-      previous: '321',
+      previous: '321'
     }
   }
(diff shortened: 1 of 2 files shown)


willfarrell:
`http-router` was written for ALB & Function URL use cases that don't already have routing done for them by APIG.
Having an APIG for a single lambda doesn't make any sense now that Function URLs are a feature. Can you share a little more about your use case as to why you need this? The PR looks fine, just want to better understand why it's needed.

tulsidas:
we have a SPA with lambdas on the back-end, we are very close to the hard limit of 500 CF resources, and also the cold start of each lambda makes the page quite slow, so we are trying to bundle the endpoints in a few lambdas to avoid both problems

Issue #945, opened 2022-11-28
Discussion: How to handle `rawBody` deprecation?


adesso-os:
I was surprised to see that `rawBody` is deprecated. The [documentation describes](https://middy.js.org/docs/middlewares/http-json-body-parser) a use case that I currently have and for which I am using `rawBody` as suggested:

> If the body has been parsed as JSON, you can access the original body (e.g. for webhook signature validation) through the `request.event.rawBody`.

As this code path seems to be deprecated, I didn't see how to properly replace this with another approach.

On Gitter, it was suggested that I should just verify the payload before parsing it. However, that would require pretty drastic restructuring of how our lambdas are currently constructed. Understanding why the current use case is apparently problematic, might also help us in creating a better scaffold for ourselves for the future.

Given that our existing use case is covered in the documentation and is still part of the current version of the documentation, I assumed that a clear migration path would be offered for consumers of the existing `rawBody` approach. Ideally, I would just copy that suggested approach :)

adesso-os:
After giving this a bit more thought, I have built a middy middleware that acts before the body parser has done its work, validates the payload, parses the event and places the result on the `context` object so I can work with it in my function body later. So far, so good.

I was looking into this approach earlier already, but the types around `middy.MiddlewareFn` are very confusing for me still. Most of the time, I'm dealing with `any` right now and have to perform loads of validation.

This is especially annoying, because in my previous solution I was purely dealing with already validated and parsed input data.

- When I want to validate the payload, I need a key from an HTTP header.
- To safely access that header value, it must have been validated.
- To validate it, I need the validator middleware, which validates not only the headers, but the entire event.
- For that to pass, the event has to be parsed already. 
- For that to happen, I need the body-parser, which replaces the `event.body` (and, with future changes, the `event.rawBody` will be gone).

I feel like this forces me into running multiple validation passes to achieve something that was fairly simple with `rawBody` still being available.

willfarrell:
The main reason `rawBody` was removed was to be consistent with all of the other middlewares that modify the body/message/event. When dealing with dozens of different AWS event structures I felt it was would be better to not modify the event outside of the purpose of the middleware. In this case add `rawBody` when the purpose is to parse `body`. 

Having multiple validation isn't a bad thing. Starting with a simple one that validates the AWS event structure (a collection of JSON schema for all AWS events is something I've thought about putting together), then another validator for just the parsed data would allow separation of what is AWS and what is custom making maintenance easier.

[...]

PR #973, opened 2022-12-22
fix(http-json-body-parser): rawBody is missing

shyouhei:
http-json-body-parser's `index.d.ts` says it sets `rawBody`:

https://github.com/middyjs/middy/blob/9fe167c060ce01456ebf3067a0ac653938e5e1c2/packages/http-json-body-parser/index.d.ts#L14

...which is in practice missing.  I doubt if this is intentional.


willfarrell:
This was removed on purpose. See https://github.com/middyjs/middy/issues/945 for further discussion.
A PR to updated the types to reflect this change is welcome.

shyouhei:
Oh OK.  That's a sad news to me, but I respect your decision.  Thanks.

PR #979, opened 2023-01-04
enablePreflightReply feature added

logemann:
This feature (default -> true) handles all OPTION requests from a cors enabled browser automatically. Since a lot of people have problems understanding cors, i made it default to true. Only problem which might arise is if somebody really wants to handle other OPTION requests in their function. 

I have not found a way to distinguish a cors OPTION requests from potential business code based OPTION requests which are very rare anyway. Happy to discuss this scenario as well. 

Tell me what you think. Its a first try. Of course its tested in a real project of mine already.

### DIFF
--- a/packages/http-cors/index.js
+++ b/packages/http-cors/index.js
@@ -18,6 +18,7 @@ const getOrigin = (incomingOrigin, options = {}) => {
 const defaults = {
   getOrigin,
   credentials: undefined,
+  enablePreflightReply: true,
   headers: undefined,
   methods: undefined,
   origin: '*',
@@ -32,6 +33,14 @@ const defaults = {
 
 const httpCorsMiddleware = (opts = {}) => {
   const options = { ...defaults, ...opts }
+
+  const httpCorsMiddlewareBefore = async (request) => {
+    if (options.enablePreflightReply) {
+      if (request.event.httpMethod === "OPTIONS") {
+        return true;
+      }
+    }
+  }
 
   const httpCorsMiddlewareAfter = async (request) => {
     normalizeHttpResponse(request)
--- a/packages/http-cors/index.js
+++ b/packages/http-cors/index.js
@@ -1,154 +1,121 @@
-import { normalizeHttpResponse } from '@middy/util'
+import {normalizeHttpResponse} from '@middy/util';
 
 const getOrigin = (incomingOrigin, options = {}) => {
   if (options.origins.length > 0) {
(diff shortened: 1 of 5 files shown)


logemann:
i tested this against a very old version of middy core, which even wasnt called core back then but just "middy". I need to recheck things with current version.

packages/http-cors/index.js
```diff
@@ -22,123 +69,53 @@ const defaults = {
...
-    if (existingHeaders.includes('Access-Control-Allow-Credentials')) {
-      options.credentials =
-        headers['Access-Control-Allow-Credentials'] === 'true'
+  const options = {
+    ...defaults,
+    ...opts
+  };
+  const httpCorsMiddlewareBefore = async (request) => {
+    if (options.enablePreflightReply) {
+      if (request.event.httpMethod === "OPTIONS") {
```
willfarrell, line 87: You should use the snippet below to ensure it works for all http event versions.
```
const httpMethod = getVersionHttpMethod[request.event.version ?? '1.0']?.(
      request.event
    )
    ```

File: packages/http-cors/index.js [hunk 2]
```diff
@@ -22,123 +69,53 @@ const defaults = {
   methods: undefined,
   origin: '*',
   origins: [],
+  enablePreflightReply: true,
```
willfarrell, line 72: What do you think of `disableEarlyPreflightResponse` or `disableStaticOrigins` (name need more thought, always the hardest part) set to `true`, with it changing to `false` as the default in the next major release. We use the disable notation in other middleware and have it disabled for now so it's not a breaking change.

packages/http-cors/index.js
```diff
@@ -22,123 +69,53 @@ const defaults = {
...
   '1.0': (event) => event.httpMethod,
   '2.0': (event) => event.requestContext.http.method
-}
+};
+
+export default httpCorsMiddleware;
+
 
-export default httpCorsMiddleware
+//# sourceMappingURL=index.js.map
```
willfarrell, line 121: not needed

willfarrell (changes requested):

willfarrell:
Great start, can you also add in a unit test?

packages/http-cors/index.js
logemann: i like disableEarlyPreflightResponse ... can we go with that?

logemann:
quite busy this week. Will try to do the remaining changes hopefully this weekend.

willfarrell:
No worries

PR #985, opened 2023-01-10
Abort the handler timeout when the handler succeeds or fail

leon19:
This pull request addresses an issue where the handler timeout is not being correctly aborted when the handler either succeeds or fails.

This leads to the lambda handler and the timeout handler being both called instead of just one of them. This is a problem if the timeout handler produces side effects (for example, logging)

Let me know if you have any questions or concerns, regards!


willfarrell:
Thanks for the PR!

PR #989, opened 2023-01-17
feat(core): pass the `request` to the `timeoutEarlyResponse` handler

leon19:
This PR adds the `request` as an argument to the `timeoutEarlyResponse()` function. This change is necessary to ensure that the function can access information about the current request, such as headers or query parameters, when handling a timeout.

By passing the request, it can now use this information to provide a more appropriate response to the client in the event of a timeout.

Additionally, this change should not cause any issues with existing code as it only adds the request as an additional parameter.

### DIFF
--- a/packages/core/__tests__/index.js
+++ b/packages/core/__tests__/index.js
@@ -771,3 +771,27 @@ test('Should not invoke timeoutEarlyResponse on error', async (t) => {
 
   t.false(timeoutCalled)
 })
+
+test('Should pass the request to the timeoutEarlyResponse handler', async (t) => {
+  const event = {}
+  const context = {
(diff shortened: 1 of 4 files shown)


willfarrell:
Thanks for the PR. Do you have a sample `timeoutEarlyResponse` function you're planning to use with this changes. The original intent was to allow customization of errors to be thrown and not have much additional logic. This custom error then triggers the onError middleware stack which have access to the headers and query params.

leon19:
I didn't even though about that use case 😅. Throwing an error and letting the `onError` middleware handle it would totally fit my needs.

Thanks for your reply and feel free to reject the PR.

PR #992, opened 2023-01-18
Feature/disable before preflight response

### LINKED ISSUE #977: extend cors to also handle preflights
**Is your feature request related to a problem? Please describe.**
With the current state, even though somebody uses cors middleware, in his handler function he needs to handle the OPTION preflight request manually.

**Describe the solution you'd like**
Extend cors middleware to react on the OPTION request in "before" cycle so that the users handler function dont need to do it.

**Describe alternatives you've considered**
None

**Additional context**
Tell me what you think of it. Looks like an obvious feature in that middleware to me.

willfarrell:
Ref #979
Closes #977

### DIFF
--- a/packages/http-cors/index.d.ts
+++ b/packages/http-cors/index.d.ts
@@ -3,7 +3,7 @@ import middy from '@middy/core'
 export interface Options {
   getOrigin?: (incomingOrigin: string, options: Options) => string
   credentials?: boolean | string
-  enablePreflightReply: true,
+  disableBeforePreflightResponse: true
   headers?: string
   methods?: string
   origin?: string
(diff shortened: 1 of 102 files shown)


willfarrell:
@logemann can you take a look?

logemann:
sorry, that i have not got time at my hand to do it myself. Being a CTO of a startup... time is always scarce.  With "have a look", you mean checking out the branch and see it works as expected?

willfarrell:
Yeah, see if it meets your needs before I merge it in.

logemann:
looks good to me.

packages/http-cors/index.d.ts
```diff
@@ -3,6 +3,7 @@ import middy from '@middy/core'
 export interface Options {
   getOrigin?: (incomingOrigin: string, options: Options) => string
   credentials?: boolean | string
+  disableBeforePreflightResponse: true
```
robin-glimp, line 6: With `true` as the only valid input according to this interface, the latest update requires me to set this to true. It's not optional either as the documentation specifies. So shouldnt this be: `disableBeforePreflightResponse?: boolean`?

willfarrell: fix releasing shortly

Issue #990, opened 2023-01-17
Middy 4.1.0 causes Jest test runner to never exit. 


edgenard:
**Describe the bug**
Functions that are tested with Jest test running never exit. Jest just hangs. The test runs and exits fine if the function is not wrapped.

**To Reproduce**
1. Install Middy version 4.1.0
2. Wrap a function with middy
3. Run a test with Jest
**Expected behaviour**
I expected the Jest to exit. Without wrapping the function in middy the test runs and Jest exits.

**Environment (please complete the following information):**
 - Node.js:v18.12.1
 - Middy: 4.1.0
 - AWS-SDK: ^2.1295.0,
 - Jest:^29.3.1

**Additional context**

willfarrell:
I personally don't use Jest, so can't be much help here. I know other use Middy with Jest, so will leave open for a while to allow others to comment.

Including a sample test would be helpful for others to assist you.

stephenkilbourn:
i see similarly.  

```
module.exports.handler = middy(async (event, context) => {
  const restaurants = await getRestaurants(context.config.defaultResults)
  const response = {
    statusCode: 200,
    body: JSON.stringify(restaurants)
  }
  return response
}).use(ssm({
  cache: true,
  cacheExpiry: 1 * 60 * 1000, // 1 mins
  setToContext: true,
  fetchData: {
    config: `/${serviceName}/${stage}/get-restaurants/config`
  }
}))
```

test file 
```
  it(`Should return an array of 8 restaurants`, async () => {
    const res = await when.we_invoke_get_restaurants()

    expect(res.statusCode).toEqual(200)
    expect(res.body).toHaveLength(8)

    for (let restaurant of res.body) {
      expect(restaurant).toHaveProperty('name')
      expect(restaurant).toHaveProperty('image')
    }
  })
```

and output from jest --detectOpenHandles

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  Timeout

      26 |   }
      27 |   return response
    > 28 | }).use(ssm({
         |        ^
      29 |   cache: true,
      30 |   cacheExpiry: 1 * 60 * 1000, // 1 mins
      31 |   setToContext: true,

      at processCache (node_modules/@middy/util/index.cjs:130:43)
      at ssmMiddleware (node_modules/@middy/ssm/index.cjs:126:43)
      at Object.ssm (functions/get-restaurants.js:28:8)
      at require (tests/steps/when.js:11:19)
[...]

PR #996, opened 2023-01-28
Update `Testing` docs with importance of disabling cache

lmserrano:
This PR intends to improve the `Testing` documentation page by explaining the importance and the way to disable Middy's cache mechanism when running tests. Failing to do so may result in test processes ending not gracefully, which can cause issues namely in CI/CD environments.

This has been reported by @stephenkilbourn on [Middy 4.1.0 causes Jest test runner to never exit. #990](https://github.com/middyjs/middy/issues/990), who also posted the solution.

It took me a while to converge to this, since the suggestions currently existing in the documentation page related with lambda function handlers (which was also my case) do not resolve this issue, so I've thought of updating the documentation page and proposing these changes.


willfarrell:
Thanks for putting this together. Hopefully it's usefull to others using Jest.

Issue #998, opened 2023-01-30
Undefined error in aws-embedded-metrics when following cloudwatch-metrics docs


ohjann:
**Describe the bug**
The [cloudwatch-metrics docs](https://middy.js.org/docs/middlewares/cloudwatch-metrics/#options) mention:
> You can provide an empty array to record all metrics without dimensions.

when using the middleware. Unfortunately this causes an error with the `aws-embedded-metrics` package where [`aws-embedded-metrics-node/src/utils/Validator.ts`](https://github.com/awslabs/aws-embedded-metrics-node/blob/master/src/utils/Validator.ts#L35) expects an object but [what's passed by the middy middleware is undefined](https://github.com/middyjs/middy/blob/main/packages/cloudwatch-metrics/index.js#L17).

**To Reproduce**
How to reproduce the behaviour:
1. Sample code:
```
const middy = require('@middy/core')
const cloudwatchMetrics = require('@middy/cloudwatch-metrics')

const handler = middy((event, context) => {
  context.metrics.putMetric('ProcessingLatency', 100, 'Milliseconds')
  context.metrics.setProperty(
    'RequestId',
    '422b1569-16f6-4a03-b8f0-fe3fd9b100f8'
  )
})

handler.use(
  cloudwatchMetrics({
    namespace: 'myAppliction',
    dimensions: []
  })
)
```
2. Thrown error
```
"TypeError: Cannot convert undefined or null to object","    at Function.keys (<anonymous>)","    at validateDimensionSet (/Users/ohjann/Programming/my_package/node_modules/aws-embedded-metrics/lib/utils/Validator.js:39:16)
```

**Expected behaviour**
Either the docs need to be updated so reflect the fact that passing an empty array is not the correct way to configure this or the code needs to be updated to allow an empty array to work correctly.

**Environment (please complete the following information):**
 - Node.js: 16.16.0
 - Middy: 4.2.1
 - AWS SDK: 2.1066.0

**Additional context**
I was able to get it to work as expected by passing an array with an empty object like so:
```
handler.use(
  cloudwatchMetrics({
    namespace: 'myAppliction',
    dimensions: [{}]
  })
)
```

willfarrell:
Not setting `dimensions` in middy will get your required result, but we should update to having it better match the AWS docs. Would you like to put a PR together, update the default to `dimensions:[]` and the logic to reflect?

ohjann:
[...]

PR #1000, opened 2023-01-31
Fix setDimensions function arg and update docs

ohjann:
Fixes https://github.com/middyjs/middy/issues/998

[`aws-embedded-metrics` `setDimensions` function signature](https://github.com/awslabs/aws-embedded-metrics-node#metriclogger) is `Record<String, String> | Record<String, String>[]` whereas the `cloudwatch-metrics` middleware was expecting `Record<String, String>[]` and passing the dimensions args through via the spread operator, which if an empty array was passed (as per middy docs) would result in an undefined error.

This should be a backwards-compatible bug fix as the original middy `Record<String, String>[]` is a valid parameter for `setDimensions` even without the spread operator.

I've also updated the docs to reflect this change and removed the explicitly mentioned defaults as I couldn't find where in the `aws-embedded-metrics` where these defaults were described, and there in fact seems to be more defaults than just the three mentioned.

### DIFF
--- a/packages/cloudwatch-metrics/__tests__/index.js
+++ b/packages/cloudwatch-metrics/__tests__/index.js
@@ -75,7 +75,7 @@ test.serial(
 )
 
 test.serial(
-  'It should call metrics.setDimensions when option passed',
+  'It should call metrics.setDimensions when option passed using plain object',
   async (t) => {
     const handler = middy(() => {})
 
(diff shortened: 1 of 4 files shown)


willfarrell:
Thanks for putting this together. Once unit tests are passing I can merge and release.

packages/cloudwatch-metrics/__tests__/index.js
```diff
@@ -93,14 +93,12 @@ test.serial(
...
-              Agent: 'CloudWatchAgent',
-              Version: 2
-            }
-          ]
+          dimensions: {
+            Runtime: 'NodeJS',
+            Platform: 'ECS',
+            Agent: 'CloudWatchAgent',
+            Version: 2
+          }
```
willfarrell, line 101: don't we need a test for an array? If an array fails this would be considered a breaking change and would have to wait for v5.

ohjann: Just added a second test there for an array of objects

willfarrell:
Thanks for putting this together.

PR #1013, opened 2023-02-14
chore: update `ws-router` handler type

esfox:
Shouldn't the type of each handler of the `ws-router` be `APIGatewayProxyWebsocketHandlerV2` instead of `APIGatewayProxyHandlerV2`? For example, in `APIGatewayProxyHandlerV2`, there is no `event.requestContext.connectionId` or most of the properties listed here:
https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-mapping-template-reference.html

This includes updating the type of the `handler` property of the `Route` type in the declaration file.


willfarrell:
Thanks for the PR, I think you're right. Once typings are passing I'd be happy to merge and release.

esfox:
May have forgotten to update the test file `test-d.ts` though...

PR #1017, opened 2023-02-23
chore: remove @middy in plugin error messages

peakyDicers:
I don't think information about what libraries a developer is using should be exposed to the public.


willfarrell:
I was planning to do this in v5, thanks for the PR.

Issue #1018, opened 2023-02-23
httpRouterHandler is not a function


lxhunter:
**Describe the bug**
When i try to import and use '@middy/http-router', jest keeps telling me: 'TypeError: httpRouterHandler is not a function'

**To Reproduce**
How to reproduce the behaviour:

```javascript 
// Test
import { test } from '@jest/globals';

import {handler} from '../../functions/mock.mjs';

import { apiGatewayEvent } from './mock/apiGatewayEvent.mjs';

test('mock', () => {
  handler(apiGatewayEvent);
});

```

```javascript 
// mock.mjs
import middy from '@middy/core';
import * as httpRouterHandler from '@middy/http-router'
import inputOutputLogger from '@middy/input-output-logger';
import httpHeaderNormalizer from '@middy/http-header-normalizer';

const getHandler = middy()
  .handler((event, context) => ({
    statusCode: 200,
    body: '{"message": "orders"}',
  }));

const postHandler = middy()
  .handler((event, context) => ({
    statusCode: 200,
    body: '{"message": "returns"}',
  }));

const routes = [
  {
    method: 'GET',
    path: '/test/orders/{id}',
    handler: getHandler,
  },
  {
    method: 'POST',
    path: '/test/returns',
    handler: postHandler,
  },
];

export const handler = middy()
  .use(inputOutputLogger())
  .use(httpHeaderNormalizer())
  .handler(httpRouterHandler(routes));
```

throws:
```
    TypeError: httpRouterHandler is not a function

      59 |   .use(inputOutputLogger())
      60 |   .use(httpHeaderNormalizer())
    > 61 |   .handler(httpRouterHandler(routes));
         |            ^
      62 |
      63 |
      64 | // handler.use(inputOutputLogger());
```

**Expected behaviour**
Not to throw?!? :)

**Environment (please complete the following information):**
 - Node.js: 18
 - Middy: last release
 - AWS SDK: 2.1310.0

**Additional context**
This might be one of those errors where the error sits behind the keyboard, but i tried all i could and other middlewares are running fine.

Please help :)

salesh:
I jumped too fast to a conclusion with pr #1019 

Shouldn't you use 

```javascript
import  httpRouterHandler from '@middyjs/http-router'
```

lxhunter:
Hi!

I tried both, even gave it an unrelated name as it is the default export, still no success. Thank you for the quick response

salesh:
Strange thing since in my env (Node 18). with your example

[...]

PR #1019, opened 2023-02-23
fix: 1018 - correct naming for http router function

salesh:
#1018 fix documentation


salesh:
closing seems like I misunderstood #1018

salesh:
Although not sure why different naming in the documentation and https://github.com/middyjs/middy/blob/main/packages/http-router/index.js#L116 usually is the same :)

PR #1020, opened 2023-02-24
Clarify header normalization docs

microbit-matt-hillsdon:
The default is lowercase so prioritise that in the docs. Make it clear that the body parser middlewares are OK with either type of normalization.

Just a suggestion, seemed easiest as a PR. Feel free to edit or fix differently. When switching a project to middy I initially completely missed that the default behaviour was lowercase as the docs strongly imply canonical form unless you spot the default canonical false note further down the page.


willfarrell:
Thanks for the docs improvement

Issue #1009, opened 2023-02-10
Appconfig middleware throws error when trying to fetch configuration of type "feature flag"


mju-spyrosoft:
**Describe the bug**
Not sure if this is a bug or a feature request but when trying to fetch Appconfig configuration of type "Feature flag" middleware throws `Failed to resolve internal value` error.

**To Reproduce**
How to reproduce the behaviour:
1. Prepare configuration in appconfig of type feature flag 
![image](https://user-images.githubusercontent.com/85890042/218144023-9af90304-85d9-4e30-b7e8-3721eab0979a.png)

2. Sample code 
```ts
import middy from '@middy/core';
import appConfig from '@middy/appconfig';

async function baseHandler(event: any, context: any) {
  console.log(context.appconfig);
}

export const handler = middy(baseHandler).use(
  appConfig({
    fetchData: {
      config: {
        Application: 'xkhkg5h',
        ClientId: 'userSpecifiedId',
        Configuration: '2cqyhl',
        Environment: 'p5oyugh'
      }
    },
    setToContext: true
  })
);
```
3. Thrown error
```json
{
    "errorType": "Error",
    "errorMessage": "Failed to resolve internal values",
    "stack": [
        "Error: Failed to resolve internal values",
        "    at getInternal (/var/task/dist/index.js:26269:11)",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)",
        "    at async appConfigMiddlewareBefore (/var/task/dist/index.js:26409:20)",
        "    at async runMiddlewares (/var/task/dist/index.js:26124:17)",
        "    at async runRequest (/var/task/dist/index.js:26081:5)"
    ]
}
```

**Expected behaviour**
Middleware should fetch configuration of type "feature flag" without throwing error

**Environment (please complete the following information):**
- Node.js: [e.g. 18.13.0]
- Middy: [e.g. ^4.2.3]
- AWS SDK - "@aws-sdk/client-appconfigdata": "^3.267.0"

**Additional context**
I called `GetConfigurationCommand` used by appconfig middleware directly in index.js. Underlying error i got is:
```
{
    "errorType": "BadRequestException",
    "errorMessage": "Feature flag configurations must be accessed via AWS AppConfig Data's GetLatestConfiguration API.",
    "name": "BadRequestException",
    ...
}
```
[...]

PR #1029, opened 2023-03-13
fix(appconfig): #1009 rewrite to not use deprecated appconfig getConfiguration command

mju-spyrosoft:
#1009 

Rewrites parts of appconfig middleware to no longer use deprecated `GetConfigurationCommand` and use `GetLatestConfigurationCommand` instead. Basically uses new way of getting appconfig. 

- Calls start session configuration once per lambda initialization. See [why](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_GetLatestConfiguration.html#API_appconfigdata_GetLatestConfiguration_ResponseSyntax)
>This token should only be used once in your first call to GetLatestConfiguration. You must use the new token in the GetLatestConfiguration response (NextPollConfigurationToken) in each subsequent call to GetLatestConfiguration ......
- On subsequent calls will use `NextPollConfigurationToken` returned by previous api call
- NextConfigurationToken is stored internally in middleware
- Note that `fetchData` object keys have changed 
```
 key: {
  ApplicationIdentifier: '...',
  ConfigurationProfileIdentifier: '...',
  EnvironmentIdentifier: '...'
 }
```

@willfarrell I run into case where i'm not sure how it should be handled. 
When calling `GetLatestConfigurationCommand` it may return null value for config [if it hasn't change last poll ](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_GetLatestConfiguration.html#API_appconfigdata_GetLatestConfiguration_ResponseSyntax). Its problematic since middy already removed configuration from cache.

You can see this specific case in failing test `It should keep previous configuration value if getLatestConfiguration returns null configuration`

 How to handle this? It could be kept in internal middleware cache in the same way configurationTokens are in this PR but not sure if that's good apporach.

TODO

- [x] change  `@aws-sdk/client-appconfig` to  `@aws-sdk/client-appconfig-data` and update middleware logic
- [x] update skipped tests
- [x] store configuration token in middleware cache
- [x] uninstall `@aws-sdk/client-appconfig`
- [x] update index.d.ts, index.test.-d.ts
- [x] ~~Move `StartConfigurationSessionCommand` to cold start part of lambda~~
- [x] Add test for when mime type is not application/json
- [x] Add test for when StartConfigurationSessionCommand fails
- [x] Run it against real AWS account. 
  - [x] feature flag configuration
  - [x] freeform configuration - Appconfig hosted configuration
  - [x] freeform configuration - ssm
 - [x] add test for when fetching multiple paramters


willfarrell:
Thanks for putting this together.

> Calls start session configuration once per lambda initialization.

It would make sense to request this as part (`StartConfigurationSessionCommand`) during the cold start (before precache). This should allow you to combine `fetchLatestConfiguration` and `fetch` together and simplify the logic.

> Note that fetchData object keys have changed

Because the options will need changing, this will have to be a breaking change and wait for v5. I hope to have an alpha out in the early summer.

> When calling GetLatestConfigurationCommand it may return null value for config [if it hasn't change last poll ](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_GetLatestConfiguration.html#API_appconfigdata_GetLatestConfiguration_ResponseSyntax). Its problematic since middy already removed configuration from cache.

Storing the previous value as a fallback when the response is `null` is fine.

mju-spyrosoft:
> It would make sense to request this as part (StartConfigurationSessionCommand) during the cold start (before precache). This should allow you to combine fetchLatestConfiguration and fetch together and simplify the logic.

Sorry not sure if i get it. I can move `StartConfigurationSessionCommand` to be part of cold start but this would mean that appconfig client creation needs to be part of cold start as well, right? 

From what im seeing in other middlewares AWS client are initialized with `await createClient(options, request)` in `before` unless they are passed into middleware or are created by prefetch

```js
function startConfigurationSession() {
    for (const internalKey of Object.keys(options.fetchData)) {
      client
      .send(
        new StartConfigurationSessionCommand(options.fetchData[internalKey])
      )
      // put config tokens into token cache
  }

//Cold start stuff
  let prefetch, client
  if (canPrefetch(options)) {
    client = createPrefetchClient(options)
    prefetch = processCache(options, fetch)
  }

  if (!client) {
    client = await createClient(options, request) //No request here. 
  }
  startConfigurationSession()
```

willfarrell:
Ah yes, it can't be included always during the cold start because if X-Ray is attached it will break. Super annoying. Don't worry about it. If you can finish off the typings I can merge it in.

Thanks for you help on this.

mju-spyrosoft:
I tried to fix typings but I might be to big of typescript noob to understand this. It seems that In `index.test.d.ts` 

```js
const options = {
  AwsClient: AppConfigDataClient,
  ....
  // TODO this does not match. If i comment it out typescript does not give errors
  awsClientCapture: captureAWSv3Client,
} as const

....

//This gives typescript error 
// use with setToContext: true
expectType<middy.MiddlewareObj<unknown, any, Error, LambdaContext & Record<'config', any>>>(
  appConfig({ ...options, setToContext: true })
)
```
>TS2345: Argument of type ..... is not assignable to parameter of type ....  Types of property 'awsClientCapture' are incompatible.       Type '<T extends Client<any, any, any>>(client: T, manualSegment?: SegmentLike | undefined) => T' is not assignable to type '(service: AppConfigDataClient) => AppConfigDataClient'.

Not sure what to do about it.

mju-spyrosoft:
@willfarrell I think this should be for review. I need to read last time through code and see if workflows are passing

willfarrell:
Haven't forgotten about this, just been busy.

mju-spyrosoft:
No problem at all. I moved it for ready to review, updated todos in PR description.

naorpeled (approved):
LGTM!

willfarrell:
Didn't mean to close this, give me a few

PR #1039, opened 2023-04-08
Support streamifyResponse

### LINKED ISSUE #1038: Support Stream Repsonse
https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/

Examples
https://github.com/aws-samples/serverless-patterns/blob/main/lambda-streaming-large-sam/src/index.js
https://github.com/aws-samples/serverless-patterns/blob/main/lambda-streaming-ttfb-pipeline-sam/src/index.mjs

willfarrell:
Closes #1038

- [x] middy supports streamifyResponse with ReadableStream body
- [x] middy supports streamifyResponse without ReadableStream body (needed?)
- [x] Add new docs page
- [x] http-content-encoding support ReadableStream body
- [x] Update http-content-encoding docs
- [x] real lambda testing

## Test
Using Function URL, arm, nodejs18.x, esm

```js
export const handler = middy({ streamifyResponse: true })
  .handler(() => {
    return {
      statusCode: ***,
      headers: {
        'Content-Type': 'text/plain',
        'X-Custom-Header': 'Example-Custom-Header'
      },
      body: ***
    }
  })
```

### DIFF
--- a/packages/core/__tests__/index.js
+++ b/packages/core/__tests__/index.js
@@ -787,8 +787,35 @@ globalThis.awslambda = {
     }
   }
 }
+
+test('Should return with streamifyResponse:true using string', async (t) => {
+  const input = 'x'.repeat(1024 * 1024)
+  const handler = middy(
+    (event, context, { signal }) => {
+      return {
+        statusCode: 200,
+        headers: {
+          'Content-Type': 'plain/text'
+        },
+        body: input
+      }
+    },
+    {
+      streamifyResponse: true
+    }
+  )
+
+  let chunkResponse = ''
+  const responseStream = createWritableStream((chunk) => {
+    chunkResponse += chunk
+  })
+  const response = await handler(event, responseStream, context)
+  t.is(response, undefined)
+  t.is(chunkResponse, input)
+})
+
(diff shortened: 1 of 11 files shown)


willfarrell:
`@middy/core` takes care of all that complexity for you. You just need to return a ReadableStream.

`@datastream/core` is my got for managin streams. Take a look at: https://github.com/willfarrell/datastream/blob/main/packages/core/index.node.js#L141 for an example on how to create them.

abarone-btf:
sorry if the question is too noob :(, but how do I "write" to the stream constantly before returning it?. I need to write in various part of the code, not just at the end when returning.

// write to stream
// async - wait "long" call
// write again to the stream
// async - wait "long" call
// write again to the stream 
// return

@willfarrell

willfarrell:
All good bro, try something like this:

```javascript
function* iterator() {
  for (var i = 0; i <= 100; i++) {
      await new Promise(r => setTimeout(r, 20));
      yield `Hello ${i} \n`
    }		
}

return {
  statusCode: 200,
  body: Readable.from(iterator(), streamOptions);
}
```

willfarrell:
Glad I could help. 😊

PR #1056, opened 2023-05-22
Update 07-with-typescript.md

furysheep:
The type of the before middleware promise should be **APIGatewayProxyResult** not void, because it can return early.


willfarrell:
Thanks for the PR!

PR #1060, opened 2023-06-04
fix(types): declare HttpError as class

aztalbot:
Typescript was throwing an error when trying to work with the HttpError class (e.g. instantiating, using `instanceof`, or extending the class). I think it should declared as a class in types, since that's what it is.


willfarrell:
Thanks for the PR!

Issue #1063, opened 2023-06-09
[@middy/secrets-manager] Add conditional caching based on Rotation Window


KillDozerX2:
Using the secrets manager middleware feels amazing however a very common scenario with Secrets is the fact that they get rotated, and the current caching support for them only allows us to cache for a certain amount of time.

We would like to be able to specify a rotation window and the plugin will then set the cache expiry based on the time left from the last rotated timestamp.

Technically we could create another middleware that serves this use case, but secret rotation is fairly common.

willfarrell:
The [GetSecretValueCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-secrets-manager/classes/getsecretvaluecommand.html) doesn't return the details, but [DescribeSecretCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-secrets-manager/classes/describesecretcommand.html) does. I'll have to think on the best way to opt-in to it, and how to connect to the internal cache.

willfarrell:
Question: Should it handle changes in the rotation rules during the life cycle of a lambda?

KillDozerX2:
> Question: Should it handle changes in the rotation rules during the life cycle of a lambda?

I don't think that would be a good idea, but there are situations where the key could be rotated while the lambda is running. We should think about it.

KillDozerX2:
> The [GetSecretValueCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-secrets-manager/classes/getsecretvaluecommand.html) doesn't return the details, but [DescribeSecretCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-secrets-manager/classes/describesecretcommand.html) does. I'll have to think on the best way to opt-in to it, and how to connect to the internal cache.

And sometimes, there are not rotations details available and the secret rotation is done manually using some other process, does it make sense to use Last Updated Time in this situation?

KillDozerX2:
Hi @willfarrell I just noticed that the documentation site hasn't been updated.

willfarrell:
Thanks, publishing now

PR #1064, opened 2023-06-09
[secret-manager]: Add in support for secret rotation date

willfarrell:
@KillDozerX2 Can you give this a review and test? I don't personally use this service.

Adds in support for scheduled secret rotation.
Does not support manual secret rotation.

Closes: #1063

### DIFF
--- a/packages/appconfig/index.js
+++ b/packages/appconfig/index.js
@@ -20,6 +20,7 @@ const defaults = {
   fetchData: {},
   disablePrefetch: false,
   cacheKey: 'appconfig',
+  cacheKeyExpiry: {},
   cacheExpiry: -1,
   setToContext: false
 }
--- a/packages/dynamodb/index.js
+++ b/packages/dynamodb/index.js
@@ -18,6 +18,7 @@ const defaults = {
   fetchData: {},
   disablePrefetch: false,
   cacheKey: 'dynamodb',
+  cacheKeyExpiry: {},
   cacheExpiry: -1,
   setToContext: false
 }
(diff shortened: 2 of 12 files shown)


KillDozerX2:
Need some time for this.

willfarrell:
No worries, no rush

willfarrell:
@KillDozerX2 Just wanted to follow up on this.

KillDozerX2 (approved):

KillDozerX2 (changes requested):
Just one more request, sometimes the secrets are rotated by external process which results in the NextRotationDate being invalid. Can we provide an option for such a situation that would allow us to use `LastChangedDate` + `cacheExpiry`?

willfarrell:
> Can we provide an option for such a situation that would allow us to use LastChangedDate + cacheExpiry
Is that not equivalent to setting the rotation schedule to equal `cacheExpiry`? If you're managing the rotation schedule yourself, I feel that is be out of scope for the core middy middlewares at this time.

KillDozerX2:
> > Can we provide an option for such a situation that would allow us to use LastChangedDate + cacheExpiry
> 
> Is that not equivalent to setting the rotation schedule to equal `cacheExpiry`? If you're managing the rotation schedule yourself, I feel that is be out of scope for the core middy middlewares at this time.

Let's say we update a secret every 30 minutes ourself, we know that the secret has to be reset if cacheExpiry + LastUpdateTime is greater than 30 minutes. This is a real scenario and unfortunately we can't make everybody change the way they do things.

willfarrell:
Apologies on the delay. I've updated to better support when `cacheExpiry` is set, allowing the secret to refresh before the rotation window. If you can do another review, I'll get this merged and released.

KillDozerX2 (approved):
This is Good! Thank you ❤️

PR #1065, opened 2023-06-15
fix: updated dynamo unmarshall object reducer to mutate instead of spread

ykageyama-mondo:
The event normalizer middleware was doing a spread inside the reducer for dynamo db object unmarshalling. 

Replaced with mutation to reduce time complexity.


willfarrell:
Good catch, thanks!

Issue #1074, opened 2023-07-04
CommonJS undefined "default" in cloudwatch-metrics middleware - re introduced


jrbarnard:
**Describe the bug**

This bug - https://github.com/middyjs/middy/issues/988 has been re introduced as of v4.4.0

The fix put in place for the bug was to add a sed rule during the release. This targets the camelCase text:
`awsEmbeddedMetrics.default.createMetricsLogger` (see [here](https://github.com/middyjs/middy/blob/4.3.1/.github/workflows/release.yml#L40C25-L40C71))

Which as you can see in the code [here](https://www.npmjs.com/package/@middy/cloudwatch-metrics/v/4.3.1?activeTab=code) is present.

However, something changed between 4.3.1 and 4.4.0 and the generated .cjs file now outputs the following text: `const metrics = _awsembeddedmetrics.default.createMetricsLogger();` (see [here](https://www.npmjs.com/package/@middy/cloudwatch-metrics/v/4.4.0?activeTab=code)) so the sed rule does not pick it up and swap it out.

**To Reproduce**
How to reproduce the behaviour:
1. Attempt to use `@middy/cloudwatch-metrics` >= 4.4.0
```
import cloudwatchMetrics from '@middy/cloudwatch-metrics';
import middy from '@middy/core';

export default (handler) => {
 handler.use(cloudwatchMetrics({
   namespace: 'embedded-metrics',
   dimensions: [],
 }));
};
```
2. Thrown error
```
"TypeError: Cannot read properties of undefined (reading 'createMetricsLogger')","    at cloudwatchMetricsBefore (/path/node_modules/@middy/cloudwatch-metrics/index.cjs:23:53)
```

**Expected behaviour**

No error occurs

**Environment (please complete the following information):**
 - Node.js: 16
 - Middy: >= 4.4.0

willfarrell:
Thanks for reporting! Would you be able to put a PR together to fix this?

jrbarnard:
> Thanks for reporting! Would you be able to put a PR together to fix this?

@willfarrell Happy to, however I was unsure why 4.4.0 introduced the change to how the .cjs files are generated so I'm not sure if the scope of the issue is larger than this? I'll do some more digging but it might be the sed change in the release is the easiest way to work around it for now.

willfarrell:
My guess is a change in `swc` or some other dependency, I agree updating the `sed` is the best approach. We don't need to know the root cause for this one imo.

PR #1075, opened 2023-07-05
fix: use lowercased aws embedded metrics import for replacement

jrbarnard:
Fixes: https://github.com/middyjs/middy/issues/1074


willfarrell:
Awesome, thanks!

PR #1089, opened 2023-08-25
fix(typescript): ssm context is enriched correctly

### LINKED ISSUE #1084: TypeScript improvement: `getInternal` type safety
As recommended by @m-radzikowski in #1023 we should improve the `getInternal` function to provide some degree of type safety.

Here follows the original message with details of what's expected for this task.

---

If you do:

```ts
middy(handler)
    .use(
        ssm({
            fetchData: {
                sampleParam: '/mySampleParam',
            },
            setToContext: true,
        }),
    )
    .before(async request => {
        request.context.sampleParam // <--- the type is JsonValue
    })
```

In `before()`, the `request.context` is enriched with parameters from `fetchData`. This is done with this line: https://github.com/middyjs/middy/blob/9db002b439b36b24095a36d4ced30cd4bf304cd5/packages/ssm/index.d.ts#L17

And it's great! It conditionally modifies output `Context` type from `ssm()` middleware, adding new keys to it.

Theoretically, the same mechanism could be applied to `getInternal()`. The `MiddlewareObj` could have another generic type with internal keys. Initially it's empty, but next middleware could add to it. Pseudo code:

```ts
export interface MiddlewareObj<
  TEvent = unknown,
  TResult = any,
  TErr = Error,
  TContext extends LambdaContext = LambdaContext,
  TInternal = {}, // <----- new generic value
> {
  before?: MiddlewareFn<TEvent, TResult, TErr, TContext>
  after?: MiddlewareFn<TEvent, TResult, TErr, TContext>
  onError?: MiddlewareFn<TEvent, TResult, TErr, TContext>
}

export type Internal<TOptions extends Options> = Record<ExtractSingles<keyof TOptions['fetchData']>, JsonValue>

declare function ssm<TOptions extends Options> (
  options?: TOptions
): middy.MiddlewareObj<unknown, any, Error, Context<TOptions>, Internal<TOptions>>
```

Then in the `before()` the request has the type with possible key values that can be used in `getInternal()` call.

---

However, I see it's much easier to extract values right now from the context. But the type of ssm param values is `JsonValue` from type-fest. I suspect this is to support both String and StringList parameters. But is it possible to get anything else than `string`, `string[]`? If parameter is not found, an exception is thrown. So I think it can be only a string or a string array.

---

I also did not know you can do handler at the end. All examples I saw have the handler at top.

It's maybe less elegant than having the handler at the top of the file, but you get `Context` extended with param names 👍 

---

Another typing problem in middy is with `event-normalizer`. If you use `@types/aws-lambda` (and probably everyone doing TS is) you get nicely normalized event, but incompatible with types. Would it be possible to return the correct type based on the provided "raw" type from `@types/aws-lambda`? So if "raw" event type is `KinesisStreamEvent`, return similar type but with `data` being `JsonObject` (or `JsonValue`), etc.

_Originally posted by @m-radzikowski in https://github.com/middyjs/middy/issues/1023#issuecomment-1613479719_

lmammino:
Fixes #1084 

~~@m-radzikowski, I have a first draft for a potential solution to what you suggested in #1084. I am not sure I understood your point about JsonValue.~~

~~I'd love if you can give this a quick review and let me know if this is going in the right direction.~~

On a second look at what I originally did in this PR, I understood that what I did was only testing the current existing solution for `ssm`. 😓 

Now I think I understand that your suggestion is to generalise that approach to something that would work for all the other middleware that change the context. Is that correct?

With that being said, there's a bit of TS magic here that I am not entirely comfortable with... Can you, please, expand a bit on your pseudo-code maybe providing something more concrete that I can work with?

Thanks

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -50,7 +50,8 @@ export interface MiddlewareObj<
   TEvent = unknown,
   TResult = any,
   TErr = Error,
-  TContext extends LambdaContext = LambdaContext
+  TContext extends LambdaContext = LambdaContext,
+  TInternal = {}
 > {
   before?: MiddlewareFn<TEvent, TResult, TErr, TContext>
   after?: MiddlewareFn<TEvent, TResult, TErr, TContext>
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -43,7 +43,7 @@ declare type MiddlewareFn<
   TEvent = any,
   TResult = any,
   TErr = Error,
-  TContext extends LambdaContext = LambdaContext
+  TContext extends LambdaContext = LambdaContext,
 > = (request: Request<TEvent, TResult, TErr, TContext>) => any
 
 export interface MiddlewareObj<
@@ -51,7 +51,6 @@ export interface MiddlewareObj<
   TResult = any,
   TErr = Error,
   TContext extends LambdaContext = LambdaContext,
-  TInternal = {}
 > {
   before?: MiddlewareFn<TEvent, TResult, TErr, TContext>
   after?: MiddlewareFn<TEvent, TResult, TErr, TContext>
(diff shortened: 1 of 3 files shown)


m-radzikowski:
See my WIP PR with basic implementation: #1093

lmammino:
Closing in favour of #1093

PR #1094, opened 2023-08-29
fix: Update http-content-encoding.md

cjbt:
`httpContentEncoding`, in the sample code, was being imported but was not being referenced. Updated doc to reference unused `httpContentEncoding` import and removed `httpCompressMiddleware`


willfarrell:
Thanks for the PR!

PR #1100, opened 2023-09-09
feat(json-body-parser): allow specifying versioned APIGW events

### LINKED ISSUE #1085: TypeScript improvement: `@middy/http-json-body-parser` should support API Gateway V2 events
As discussed in #1023 and suggested by @apalchys:

---

List of issues my team see (I'm going to keep it up to date):

- `@middy/http-json-body-parser` defines event using legacy `APIGatewayEvent` which describes event payload format v1. No Payload format v2 support.

_Originally posted by @apalchys in https://github.com/middyjs/middy/issues/1023#issuecomment-1484228944_

naorpeled:
closes #1085

### DIFF
--- a/packages/http-json-body-parser/index.d.ts
+++ b/packages/http-json-body-parser/index.d.ts
@@ -7,13 +7,15 @@ interface Options {
   disableContentTypeError?: boolean
 }
 
-export type Event<APIGatewayVersionedEvent = APIGatewayEvent | APIGatewayProxyEventV2> = Omit<APIGatewayVersionedEvent, 'body'> & {
+type VersionedApiGatewayEvent = APIGatewayEvent | APIGatewayProxyEventV2
+
+export type Event<APIGatewayEventType extends VersionedApiGatewayEvent = VersionedApiGatewayEvent> = Omit<APIGatewayEventType, 'body'> & {
   /**
    * The body of the HTTP request.
    */
   body: JsonValue
 }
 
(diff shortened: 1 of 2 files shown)


packages/http-json-body-parser/index.d.ts
```diff
@@ -1,19 +1,21 @@
...
 
 interface Options {
   reviver?: (key: string, value: any) => any
   disableContentTypeError?: boolean
 }
 
-export type Event = Omit<APIGatewayEvent, 'body'> & {
+type VersionedApiGatewayEvent = APIGatewayEvent | APIGatewayProxyEventV2
+
+export type Event<APIGatewayEventType extends VersionedApiGatewayEvent = VersionedApiGatewayEvent> = Omit<APIGatewayEventType, 'body'> & {
```
naorpeled, line 12: BTW, I think that at some point we should rename this to ParsedEvent,
not super crucial, but it conveys its meaning more precisely.

@lbertenasco wdyt?

lmammino: I like the suggestion, can we start by doing a type alis and deprecating the old one? Thoughts @willfarrell ?

lmammino (approved):
This looks good to me! Thanks a million for taking the time @naorpeled !

@willfarrell , should we target this to the `main` branch or is this best to go together with `v5`?

I guess the choice probably depends on whether we'd like to rename `Event` to `ParsedEvent`...

willfarrell:
> should we target this to the main branch or is this best to go together with v5?

Let's put it all in v5, it will be soon enough.

Issue #1096, opened 2023-09-04
Middy handler always requires an async handler


minjie0501:
**Describe the bug**
The middy handler always expects the lambda handler to return `void` or `Promise<any>`. Now i have to make my handlers async even if they don't do any async tasks. I would like to use it with a synchronous handler.

**To Reproduce**
How to reproduce the behaviour:
Sample code:
```
import middy from "@middy/core";
import { APIGatewayProxyEvent, Context } from "aws-lambda";

const lambdaHandler = (event: APIGatewayProxyEvent, context: Context) => {
  return {
    statusCode: 200,
    body: JSON.stringify({
      message: "Hello world!",
      input: event,
    }),
  };
};

export const handler = middy(lambdaHandler);

```

Thrown error:
`Type '{ statusCode: number; body: string; }' is not assignable to type 'void | Promise<any>'.`

**Expected behaviour**
Middy should allow sync handlers

**Environment (please complete the following information):**
 - Node.js: 18
 - Middy: 4.6.1

willfarrell:
Sounds like the types need to be updated. Like to put a PR together?

cjbt:
Took a stab at this: https://github.com/middyjs/middy/pull/1102. Let me know if I'm missing anything.

minjie0501:
> Took a stab at this: #1102. Let me know if I'm missing anything.

Haven't really had time to take a look myself, thanks for the PR! 🙌

PR #1102, opened 2023-09-12
fix(core): allow sync handlers

cjbt:
closes https://github.com/middyjs/middy/issues/1096
- Extended Handler type from aws-lambda and created a new `LambdaHandler` type to include an additional return type
- updated `MiddyInputHandler` to take in additional non async return type.


packages/core/index.test-d.ts
```diff
@@ -196,3 +205,178 @@ expectType<middy.MiddyfiedHandler<unknown>>(streamifiedResponseHandler)
...
+    logGroupName: '',
+    logStreamName: '',
+    getRemainingTimeInMillis: (): number => 1,
+    done: () => { },
+    fail: (_) => { },
+    succeed: () => { }
+  }
+  // synced handler but could be void | Promise<TResult> | TResult
+  // so it needs to be awaited
+  return await syncedHandler(sampleEvent, sampleContext, () => {})
```
cjbt, line 318: Called it out in the comments, but calling it out again to say that this needs to be awaited just because it could be async.

cjbt:
Just to add, not sure why the test is failing on util since I don't have any changes to that package. Add insult to injury the same tests are passing for me locally.

packages/core/index.test-d.ts
```diff
@@ -3,10 +3,19 @@ import middy from '.'
...
   APIGatewayProxyResult,
-  Context,
-  Handler as LambdaHandler
+  Callback,
+  Context
 } from 'aws-lambda'
 
+// extended Handler type from aws-lambda
+// to include synced TResult
+type LambdaHandler<TEvent = any, TResult = any> = (
```
cjbt, line 12: I think an ideal solution is to somehow extend the return type of the original `Handler` type from aws. Looking to see if that's even possible.

packages/core/index.test-d.ts
```diff
@@ -4,9 +4,18 @@ import {
   APIGatewayProxyEvent,
   APIGatewayProxyResult,
   Context,
-  Handler as LambdaHandler
+  Handler as AWSLambdaHandler
 } from 'aws-lambda'
 
+type ModifyReturnType<T, NewReturn> = T extends (
```
cjbt, line 10: I think I landed on an implementation. This gives the ability to still depend on `AWSLambdaHandler` from aws and add an additional return type

willfarrell:
Thanks for this! I'll get it released shortly

PR #1105, opened 2023-09-17
fix: validator benchmark

aztalbot:
- small thing I noticed while working on #1104 , the benchmark wasn't testing a compiled schema getting passed in


willfarrell:
benchmarks aren't currently used, but will hopefully be a focus after v5 release. Thanks for this.

PR #1108, opened 2023-10-06
SSM: fix: some mixes of named/paths results in undefined named param values

WillGresham:
In the SSM  middleware when `fetchData` is a mix of names and paths we're seeing an issue where in some cases unless the parameters are ordered as `[...pathParams, ...namedParams]` we end up with named parameters being `undefined`

PR here for your review/discussion: it looks like the issue happens when the condition for batching causes a `continue` on all named items in a mixed list


willfarrell:
Thanks for the PR! I'll have this released out in a bit.

PR #1111, opened 2023-10-18
Pick 1.0 before vpc when both methods are present

joksnet:
This was found when using `http-event-normalizer` with `LocalLambda` that sends an event with both methods (see [here](https://github.com/vcfvct/ts-lambda-local-dev/blob/9d5f0eb22c3a1201cced58f33a5c4e22a35eca8c/src/local.lambda.ts#L70-L71)).


willfarrell:
Closing, root issue appears to have been addressed.

PR #1117, opened 2023-10-27
fix(http-error-handler): non http errors will always be handled 

qoomon:
http-error-handler currently does not handle non http errors if `fallbackMessage` is not set.

This PR will handled all non http errors as 'Internal Server Error' even if no fallback message is configured.
I think this is a more expected behaviour of an error handler.

### DIFF
--- a/packages/http-error-handler/index.js
+++ b/packages/http-error-handler/index.js
@@ -20,7 +20,7 @@ const httpErrorHandlerMiddleware = (opts = {}) => {
     }
 
     // Non-http error OR expose set to false
-    if (!request.error.expose || !request.error.statusCode)) {
+    if (!request.error.expose || !request.error.statusCode) {
       request.error = {
         statusCode: 500,
         message: options.fallbackMessage || 'Internal Server Error',
--- a/packages/http-error-handler/index.js
+++ b/packages/http-error-handler/index.js
@@ -20,13 +20,10 @@ const httpErrorHandlerMiddleware = (opts = {}) => {
     }
 
     // Non-http error OR expose set to false
-    if (
-      options.fallbackMessage &&
-      (!request.error.statusCode || !request.error.expose)
-    ) {
+    if (!request.error.expose || !request.error.statusCode)) {
       request.error = {
         statusCode: 500,
-        message: options.fallbackMessage,
+        message: options.fallbackMessage || 'Internal Server Error',
         expose: true
       }
     }


willfarrell:
Thanks for putting a PR together. This proposes a breaking change, so would need to go into v5 if approved. This logic is in place to allow handling http and internal errors in separate middlewares. Additionally, this is in place to not have the default message be in english.

Let me take some time to reflect on this. I agree that needing to always setting it for the most common use case can get annoying depending on how you structure your code.

qoomon:
I think in case of a http lambda, the http error handler is or should always be the last one. Every error handling can be done before, however if an unexpected error reaches the http error handler it should response with an 500 instead of just doing nothing.

willfarrell:
After some reflection and some more research, I think we can merge this into the v5 branch. I'll can update the tests, docs and change log. Thanks for putting this together.

grant-d:
We have a middy `APG` (Api Gateway) / `lambda` stack that has been working great for a couple of years. 
I upgraded the stack a couple of days ago, from `@middy 4 -> 5`
In doing so, our Custom Authorizer in APG does not handle `Unauthorized` properly anymore.

I realize that `5` is a breaking change, but I am not sure what the guidance is. Everything works fine in other lambdas, but APG has specific patterns and expectations for custom-authorizers, and that seems broken.

Please advise.

BEFORE (using `middy@4`):
```typescript
  // import @middy@4.0
  if { userIsNotAuthenticated } throw Error('Unauthenticated') // AWS prescribes this pattern with Custom Authorizers

  middy.use(httpErrorHandler()) // etc

  // APG CloudWatch logs (everything looks GOOD)
  Gateway response type: UNAUTHORIZED with status code: 401
  Gateway response body: 
  {
      "message": "Unauthorized"
  }

  // Postman response
  {
      "message": "Unauthorized"
  }
```

AFTER (upgrade to `middy@5`):
```typescript
  // import @middy@5.0 // This is the only change
  if { userIsNotAuthenticated } throw Error('Unauthenticated')
  // I also tried `throw createError.Unauthorized()` per docs, but similar result to below

  middy.use(httpErrorHandler())

  // APG CloudWatch logs (BAD)
  Authorizer result body before parsing:
  {
      "statusCode": 500,
      "headers": {} // Note the lack of a `message` property, which by default APG expects for 4XX responses/templates
  }
  Execution failed due to configuration error: Invalid JSON in response: Unrecognized field "statusCode" , not marked as ignorable

  // Postman response
  {
      "message": null
  }
```

AFTER (removed `httpErrorHandler`):
```typescript
  // import @middy@5.0
  if { userIsNotAuthenticated } throw Error('Unauthenticated')

  // middy.use(httpErrorHandler()) // Removed this handler

  // APG CloudWatch logs (kinda GOOD)
  // Gateway response type: UNAUTHORIZED with status code: 401 // Weird, this line is missing in logs now
  Gateway response body: 
  {
      "message": "Unauthorized"
  }

  // Postman response
  {
      "message": "Unauthorized"
  }
```

Issue #1116, opened 2023-10-26
Middyfied handler function type signature is incorrect


zzantozz:
**Describe the bug**

The type signature of the function returned by `middy(handler).use(middyJsonBodyParser())` is incorrect. A valid lambda function should accept two arguments of type `APIGatewayProxyEvent` and `Context`. The function returned by the above instead expects the event to be of type `Omit<APIGatewayProxyEvent, "body"> & { body: JsonValue; }"` in middy 4.6.5. The body has to be a string. Because `JsonValue` is a union including `string`, there are no compile-time failures, but the type is too broad. If you pass anything other than a string as the body, the json parser fails.

**To Reproduce**

This is a Jest test that intentionally fails to compile to show the type. I've omitted the rather lengthy code that creates the test context.

```typescript
import { APIGatewayProxyEvent, Context } from "aws-lambda";
import * as utils from "../mock-utils";
import middy from "@middy/core";
import middyJsonBodyParser from "@middy/http-json-body-parser";

test("middy type test", () => {
  const handler = (event: APIGatewayProxyEvent, _context: Context) => {
    console.log(event.body);
  };
  const wrapped = middy(handler).use(middyJsonBodyParser());
  wrapped({}, utils.sampleMiddifiedContext());
});
```

The full error:

```
test/unit/middy.test.ts:11:11 - error TS2345: Argument of type '{}' is not assignable to parameter of type 'Omit<APIGatewayProxyEvent, "body"> & { body: JsonValue; } & APIGatewayProxyEvent'.
  Type '{}' is missing the following properties from type 'Omit<APIGatewayProxyEvent, "body">': headers, multiValueHeaders, httpMethod, isBase64Encoded, and 7 more.

11   wrapped({}, utils.sampleMiddifiedContext());
```

**Expected behaviour**

I expect the type signature of the function returned by middy to be the same as a lambda function that hasn't been processed by middy.

**Environment:**

 - Node.js: 18.18.0
 - Middy: 4.6.5

willfarrell:
Is this issue addressed in https://github.com/middyjs/middy/pull/1119#issuecomment-1785281655 ?

zzantozz:
It doesn't appear to be fixed. With middy `5.0.0-alpha.2`, it still gives approximately the same error:
```
[...]

PR #1119, opened 2023-10-29
fix(http-json-body-parser): parameter typings

GentileFulvio:
An attemp to fix the following bug #1116 .
This is my first time submitting a PR in this repo so let me know if I interpreted anything wrong

### DIFF
--- a/packages/http-json-body-parser/index.d.ts
+++ b/packages/http-json-body-parser/index.d.ts
@@ -14,6 +14,6 @@ export type Event = Omit<APIGatewayEvent, 'body'> & {
   body: JsonValue
 }
 
-declare function jsonBodyParser (options?: Options): middy.MiddlewareObj<Event>
+declare function jsonBodyParser (options?: Options): middy.MiddlewareObj<APIGatewayEvent, Event>
 
 export default jsonBodyParser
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for putting the PR together. I think this change has already been made in our v5 branch (see below), does this meet your needs? We're just waiting on AWS to release the nodejs-20 runtime before releasing it. There is an alpha you can get now though.

```
import { APIGatewayEvent, APIGatewayProxyEventV2 } from 'aws-lambda'
import { JsonValue } from 'type-fest'

interface Options {
  reviver?: (key: string, value: any) => any
  disableContentTypeError?: boolean
}

type VersionedApiGatewayEvent = APIGatewayEvent | APIGatewayProxyEventV2

export type Event<APIGatewayEventType extends VersionedApiGatewayEvent = VersionedApiGatewayEvent> = Omit<APIGatewayEventType, 'body'> & {
  /**
   * The body of the HTTP request.
   */
  body: JsonValue
}

declare function jsonBodyParser<APIGatewayEventType extends VersionedApiGatewayEvent = VersionedApiGatewayEvent> (options?: Options): middy.MiddlewareObj<Event<APIGatewayEventType>>

export default jsonBodyParser
```

willfarrell:
Close because I believe this has already been done in v5. Please re-open if you think I'm mistaken.

PR #1120, opened 2023-10-30
http-json-body-parser - json a default content type

qoomon:


willfarrell:
Thanks for putting a PR together. Can you share your use case that requires this change. Looks like you want a way to allow invalid http request for some reason.

qoomon:
I'd like the server/lambda to assume content-type as json if content-type header is not set.

willfarrell:
I would suggest having the request send the proper headers. If that's not possible a custom normalizer middleware to clean up the invalid request can be added. Something like:

```
.before((request) => {
  if (!request.event.headers.['content-type'] && request.event.body.substring(0,1) === '{') {
    request.event.headers.['content-type'] = 'application/json'
  }
})
```

> Any HTTP/1.1 message containing an entity-body SHOULD include a Content-Type header field defining the media type of that body. If and only if the media type is not given by a Content-Type field, the recipient MAY attempt to guess the media type via inspection of its content and/or the name extension(s) of the URI used to identify the resource. If the media type remains unknown, the recipient SHOULD treat it as type "application/octet-stream".

http://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html

Reflecting on this a little more, I wonder if the `http-event-normalizer` should have an option to set a default `Content-Type` for those that don't want their endpoint to be as strict.

qoomon:
On one site I understand your points, on the other side it still feel a little odd because the opposite middleware `httpJsonBodySerializer` has a default `defaultContentType` option.

willfarrell:
`http-json-body-parser` interprets headers from other developers (out of your control). `http-response-serializer` interprets headers from the developer that made the response (in your control).

qoomon:
Because of  `http-json-body-parser ` interprets headers out of my control I want at least control how to interpret a content if no content type header is set.

willfarrell:
I just pushed a feature into `http-header-normalizer` to allow default headers. It will be in 4.7.0

PR #1122, opened 2023-11-01
fix: don't throw an error on undefined response or event

### LINKED ISSUE #1121: TypeError: Cannot read properties of undefined (reading 'Accept')
https://github.com/middyjs/middy/blob/f6c152404d7c714b365ad60839ef3ea6b6d554eb/packages/http-response-serializer/index.js#L12C26-L12C26

It appears that `normalizeHttpResponse()` in the event of undefined or an empty object creates new objects and returns them. This function is relying on side effects for the changes so if new objects are created they're not seen. The result is that if an undefined object is passed in you get an error. It should probably be more like

```
request = normalizeHttpResponse(request);
```

phawxby:
Closes #1121 

> It appears that normalizeHttpResponse() in the event of undefined or an empty object creates new objects and returns them. This function is relying on side effects for the changes so if new objects are created they're not seen. The result is that if an undefined object is passed in you get an error. It should probably be more like

### DIFF
--- a/packages/http-event-normalizer/index.js
+++ b/packages/http-event-normalizer/index.js
@@ -21,6 +21,7 @@ const httpEventNormalizerMiddleware = () => {
     // event.headers ??= {} // Will always have at least one header
     event.pathParameters ??= {}
     event.queryStringParameters ??= {}
+    event.headers ??= {}
   }
 
   return {
--- a/packages/http-response-serializer/index.js
+++ b/packages/http-response-serializer/index.js
@@ -9,7 +9,10 @@ const defaults = {
 const httpResponseSerializerMiddleware = (opts = {}) => {
   const { serializers, defaultContentType } = { ...defaults, ...opts }
   const httpResponseSerializerMiddlewareAfter = async (request) => {
-    normalizeHttpResponse(request)
+    request.response = normalizeHttpResponse(request)
+
+    request.event ??= {}
+    request.event.headers ??= {}
 
     // skip serialization when Content-Type or content-type is already set
     if (
--- a/packages/http-response-serializer/index.js
+++ b/packages/http-response-serializer/index.js
@@ -11,9 +11,6 @@ const httpResponseSerializerMiddleware = (opts = {}) => {
   const httpResponseSerializerMiddlewareAfter = async (request) => {
     request.response = normalizeHttpResponse(request)
 
-    request.event ??= {}
-    request.event.headers ??= {}
-
     // skip serialization when Content-Type or content-type is already set
     if (
       request.response.headers['Content-Type'] ??
@@ -29,7 +26,8 @@ const httpResponseSerializerMiddleware = (opts = {}) => {
       types = [request.event.requiredContentType]
     } else {
       const acceptHeader =
-        request.event.headers.Accept ?? request.event.headers.accept
+        request.event.headers &&
+        (request.event.headers.Accept ?? request.event.headers.accept)
       types = [
         ...((acceptHeader && Accept.mediaTypes(acceptHeader)) ?? []),
         request.event.preferredContentType,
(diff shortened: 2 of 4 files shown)


packages/http-response-serializer/index.js
```diff
@@ -9,7 +9,10 @@ const defaults = {
 const httpResponseSerializerMiddleware = (opts = {}) => {
   const { serializers, defaultContentType } = { ...defaults, ...opts }
   const httpResponseSerializerMiddlewareAfter = async (request) => {
-    normalizeHttpResponse(request)
+    request.response = normalizeHttpResponse(request)
+
+    request.event ??= {}
+    request.event.headers ??= {}
```
willfarrell, line 15: Event normalization should be in https://github.com/middyjs/middy/blob/main/packages/http-event-normalizer/index.js.
However,
1. How are you triggering a lambda without an event?
2. How are you triggering a lambda without headers in an HTTP event?

phawxby: Good question, I'm incrementally making my way through the cloudwatch logs to try and figure that out. I believe the inner code is throwing an error, returning something bad, then the response serialiser is trying to handle it but then throwing an error too, so I can't actually see the inner error right now. 

So while clearly I need to fix the inner code I feel like it makes sense to make the serialiser a bit more fault tolerant so the inner error can be seen.

```
ERROR	Invoke Error 	
{
    "errorType": "TypeError",
    "errorMessage": "Cannot read properties of undefined (reading 'Accept')",
    "stack": [
        "TypeError: Cannot read properties of undefined (reading 'Accept')",
        "    at httpResponseSerializerMiddlewareAfter (/var/task/handler.js:101142:50)",
        "    at runMiddlewares (/var/task/handler.js:97057:23)",
        "    at async runRequest (/var/task/handler.js:97035:7)"
    ]
}
```

I've moved the code as suggested and gone the route of using a null conditional instead.

phawxby: I found the issue this morning and it was relating to the way the data was being formatted when being passed into an integration test. It probably would've been easier to see had this outer error not occurred.

packages/http-response-serializer/index.js
```diff
@@ -9,7 +9,7 @@ const defaults = {
 const httpResponseSerializerMiddleware = (opts = {}) => {
   const { serializers, defaultContentType } = { ...defaults, ...opts }
   const httpResponseSerializerMiddlewareAfter = async (request) => {
-    normalizeHttpResponse(request)
+    request.response = normalizeHttpResponse(request)
```
willfarrell, line 12: I haven't reviewed this yet. Blocking off time at the end of the year is hard.

packages/http-event-normalizer/index.js
```diff
@@ -21,6 +21,7 @@ const httpEventNormalizerMiddleware = () => {
     // event.headers ??= {} // Will always have at least one header
```
willfarrell, line 21: @phawxby Did you see this comment?

I was asking `How are you triggering a lambda without headers in an HTTP event?` to better understand how you're able to create an HTTP event without any headers. As far as I'm aware, it's impossible. I'm really curious how you're able to trigger lambda through APIG/ALB/Function URL without headers.

Others have run into something similar which turned out be their unit tests being incomplete.

packages/http-response-serializer/index.js
willfarrell: After some refection, I opted to make a different change that had the same affect that was more consistent with other middlewares. See 4.6.6

packages/http-event-normalizer/index.js
willfarrell: Ah, that make so much sense now. Middlewares with `http-*` were built to be used with APIG/ALB/Function URL and the event they create. These middlewares are overkill for a lambda that gets called directly with Invoke.

I'm going to close this PR for now then.

PR #1124, opened 2023-11-08
feat: add `@iress/middy-http-path-router` to third-party middleware docs

RodneyMarsh:


willfarrell:
Thanks for building something for the middy community. Once you're ready, click `ready for review` and I'll merge it in.

Curious, what was missing from https://middy.js.org/docs/routers/http-router that inspired you to build your own? It looks really similar except for using `:var` vs `{var}` (APIG notation) within paths.

PR #1131, opened 2023-11-11
fix(middlewares/http-json-body-parser): narrow body type to string

### LINKED ISSUE #1116: Middyfied handler function type signature is incorrect

naorpeled:
closes #1116


willfarrell:
Thanks! Can you update to go against the v5 branch please.

naorpeled:
> Thanks! Can you update to go against the v5 branch please.

sure thing.

PR #1132, opened 2023-11-14
fix: don't throw errors in http-event-normalizer

robertbeal:

A PR to propose changing the behaviour of the http-event-normalizer (more as a discussion point that formal PR)

Currently, the http-event-normalizer middleware [will throw an error](https://github.com/middyjs/middy/blob/main/packages/http-event-normalizer/index.js#L8) if the HTTP version cannot be established. For example if HTTP Method is missing from the event.

- (discussion point) should the middleware care? Or care enough to throw an error which breaks the code flow. My argument is that it's designed to ensure objects are instantiated/normalised,  and not to police the HTTP validity of an event object.
- It differs in behaviour (inconsistency) from the http-header-normalizer which does not throw an error if the HTTP version cannot be figured out. This PR would make the behaviour more consistent.

Impact:

It is a breaking change as the behaviour of the middleware will be different

No

N/A

**Environment:**
 - Node.js: 18

**Any other comments?**

**Todo List:**
- [x] Amended tests
  - [x] Unit tests
  - [ ] Benchmark tests (if applicable)


robertbeal:
Just had a thought, I could amend the middleware to be configurable, as to whether it throws an error or not. That would maintain compatibility..

willfarrell:
Thanks for the PR and good point. It should be up to the validator to throw that error. We're just about to release v5, so your timing is perfect. If you can update it be against the `release/5.0` branch I can get it merged in before it releases (any day now, waiting for the new runtime to release).

packages/http-event-normalizer/index.js
```diff
@@ -3,13 +3,6 @@ const httpEventNormalizerMiddleware = () => {
     const { event } = request
 
     const version = pickVersion(event)
-    const isHttpEvent = isVersionHttpEvent[version]?.(event)
-    if (!isHttpEvent) {
-      throw new Error('[http-event-normalizer] Unknown http event format')
-    }
-    // VPC Lattice is an http event, however uses a different notation
-    // - query_string_parameters
-    // - is_base64_encoded
```
willfarrell, line 12: Can you leave this comment in please.

willfarrell (changes requested):

packages/http-event-normalizer/index.js
robertbeal: my bad, didn't intend to remove that. I'll add it back

robertbeal:
Thanks for the prompt response. I've rebased it off `release/5.0` (hope I got the upstream/fetch/rebase right).

robertbeal:
Will re-do with a cleaner PR git log

PR #1133, opened 2023-11-15
fix: don't throw error for non http event

robertbeal:

A PR to propose changing the behaviour of the http-event-normalizer (more as a discussion point that formal PR)

Currently, the http-event-normalizer middleware [will throw an error](https://github.com/middyjs/middy/blob/main/packages/http-event-normalizer/index.js#L8) if the HTTP version cannot be established. For example if HTTP Method is missing from the event.

- (discussion point) should the middleware care? Or care enough to throw an error which breaks the code flow. My argument is that it's designed to ensure objects are instantiated/normalised,  and not to police the HTTP validity of an event object.
- It differs in behaviour (inconsistency) from the http-header-normalizer which does not throw an error if the HTTP version cannot be figured out. This PR would make the behaviour more consistent.

Impact:

It is a breaking change as the behaviour of the middleware will be different

No

N/A

**Environment:**
 - Node.js: 18

**Any other comments?**

**Todo List:**
- [x] Amended tests
  - [x] Unit tests
  - [ ] Benchmark tests (if applicable)

### DIFF
--- a/packages/http-event-normalizer/__tests__/index.js
+++ b/packages/http-event-normalizer/__tests__/index.js
@@ -7,30 +7,22 @@ const context = {
   getRemainingTimeInMillis: () => 1000
 }
 
-test('It should throw error when invalid version', async (t) => {
+test('It should not throw error when invalid version', async (t) => {
   const event = {
     version: '3.0'
   }
 
   const handler = middy((event) => event).use(httpEventNormalizer())
-  try {
-    await handler(event, context)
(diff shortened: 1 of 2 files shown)


robertbeal:
@willfarrell FYI re-made the PR as I messed up rebasing. Should all be ok now, squashed into a single commit.

willfarrell:
Thanks for updating. v5 will release later today.

PR #1146, opened 2023-11-29
feat: Add name property to MiddlewareObj typings

daedal-knickerbockers:
Each middleware can be assigned an optional name that might then handed to a plugin in the `beforeMiddleware` or `afterMiddleware` hook. This is currently not possible when using TypeScript with the default `MiddlewareObj` interface, as it does not allow the `name` property.

This PR adds an optional `name` property to the `MiddlewareObj`.


willfarrell:
Thanks for the PR!

PR #1157, opened 2024-01-09
Check if body was passed in

vanerleo:
Check if body was passed in
useful when using catchall path


willfarrell:
Thanks for opening a PR.
You can already do this by setting `disableContentTypeError:true`. When the `Content-Type` is not `application/json` it will skip the middleware.

PR #1158, opened 2024-01-09
Add the rest of the parameters in the http-multipart-body-parser middleware

lopepardo:
I found that the last three parameters of the 'file' event in Busboy were not being passed correctly. The 'info' parameter was being destructured, which only pointed to the 'filename' attribute of the event.

Before the Change
Captura de pantalla 2024-01-09 135020

After the Change
Captura de pantalla 2024-01-09 135044


willfarrell:
Thanks for the PR! I'll get it released out shortly.

PR #1160, opened 2024-01-10
Do not throw error if event is not http (http-cors)

sharifhh:
for reference: https://github.com/middyjs/middy/pull/1133

we can call lambdas using invoke or any other way other than API gateway.

I don't think http-cors middleware should break the whole code if the event it is receiving is not a real http event, it will append to headers and that's it.

### DIFF
--- a/packages/http-cors/__tests__/index.js
+++ b/packages/http-cors/__tests__/index.js
@@ -819,15 +819,11 @@ test('it should set Vary header if present in config', async (t) => {
   })
 })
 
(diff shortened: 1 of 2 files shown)


OmarMuhtaseb:
Mmm, it makes sense. Since in AWS, you can invoke the lambda directly using `aws-sdk`. So, this error should not be fired in this case.

willfarrell:
Thanks for the PR. Yeah, I don't think we need this error anymore.

Issue #1164, opened 2024-01-30
index.d.ts still allows callback based handlers. 


everett1992:
**Describe the bug**
The third argument of MiddyInputHandler is documented as `callback: () => ...`, but in practice it is { signal: AbortSignal }`.

**To Reproduce**
```
    const wrapped = middy((e, c, callback) => {
      callback satisfies () => void
      expect(callback).toEqual({
        signal: expect.any(AbortSignal),
      })
      callback(null, {});
    })

    return new Promise((resolve, reject) => {
      wrapped({}, {}, (err, res) => (err != null) ? reject(err) : resolve(res));
    })

```

**Expected behaviour**
* Callback handlers should be a type error. Middy never calls the callback argument of a middified handler, and it never calls a wrapped handler with a callback argument.
 * The `signal` argument should be documented in the index.d.ts file.

PR #1165, opened 2024-01-30
fix: remove handler callback argument

everett1992:
Middy does not support callback handlers since 1.x, the third argument is actually the timeout signal.

Fixes https://github.com/middyjs/middy/issues/1164

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -1,5 +1,4 @@
 import {
-  Callback as LambdaCallback,
   Context as LambdaContext,
   Handler as LambdaHandler
 } from 'aws-lambda'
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for the PR!

phmasek:
@willfarrell this commit breaks the typing. It's not possible to seamlessly use middy with Lambda handlers as the Lambda handler is expecting the third argument to be of type `Callback` but it's now replaced with the incompatible type `MiddyHandlerObject`. Please see print screen. This is further described in issue https://github.com/middyjs/middy/issues/1176.

We're currently using `"@types/aws-lambda": "^8.10.136"` and `"aws-lambda": "^1.0.7"`.

We cannot update Middy beyond `v5.2.3` until this is resolved. This is why I added a pull request https://github.com/middyjs/middy/pull/1186.

image

PR #1167, opened 2024-02-02
chore: add `@aws-sdk/client-secrets-manager` as a peer dependency to `@middy/secrets-manager`

AbdelrahmanHafez:
This PR adds `@aws-sdk/client-secrets-manager` as a peer dependency to `@middy/secrets-manager`.

I came across an issue today where I was moving around some dependencies on a project that I worked on and I started getting an error that `@aws-sdk/client-secrets-manager` was missing.

I would expect it to be marked as a peer dependency if it's required to run `@middy/secrets-manager`.

### DIFF
--- a/packages/secrets-manager/package.json
+++ b/packages/secrets-manager/package.json
@@ -60,6 +60,9 @@
   "dependencies": {
     "@middy/util": "5.2.3"
   },
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for the PR. We purposely left it as a devDependency and included it in the doc to install separately because you can deploy to lambda without it.

PR #1172, opened 2024-02-08
Fix badges

dnicolson:
Before:
PixelSnap 2024-02-08 at 18 36 34@2x

After:
PixelSnap 2024-02-08 at 18 36 24@2x


willfarrell:
Thanks for the PR!

PR #1185, opened 2024-03-01
Pkg: httpHeaderNormalizer to handle canonical sanitized headers

ScoobySmack2612:
Given `canonical:  true`,
And an event containing a header that's "sanitized" (X----)
A runtime error occurs, "Cannot read property 'toUpperCase' of undefined".

**To Reproduce:**

**Unit test** 
```
test('It should not fail given a corrupted header key', async (t) => {
  const handler = middy((event, context) => event)

  handler.use(httpHeaderNormalizer({ canonical: true }))

  const event = {
    headers: {
      'X----': 'foo'
    }
  }

  const expectedHeaders = {
    'X----': 'foo'
  }

  const originalHeaders = { ...event.headers }

  const resultingEvent = await handler(event, context)

  t.deepEqual(resultingEvent.headers, expectedHeaders)
  t.deepEqual(resultingEvent.rawHeaders, originalHeaders)
})
```
**Thrown error:**  "Cannot read property 'toUpperCase' of undefined".

**Expected behavior:** Should simply return the raw header

**Additional context:**
We had a client who appears to be sanitizing headers resulting in our lambdas receiving a header that looks like `X----`. Though this is a one-off, rather than push the onus to our client, an empty string fallback gets them back up an running.

### DIFF
--- a/packages/http-json-body-parser/__tests__/index.js
+++ b/packages/http-json-body-parser/__tests__/index.js
@@ -228,26 +228,3 @@ test('It should handle invalid base64 JSON as an UnprocessableEntity', async (t)
     t.regex(e.cause.data.message, /^Unexpected token/)
   }
 })
-
-test('It should not fail given a corrupted header key', async (t) => {
-  const handler = middy((event, context) => event)
-
-  handler.use(httpHeaderNormalizer({ canonical: true }))
-
-  const event = {
-    headers: {
-      'X----': 'foo'
-    }
-  }
-
-  const expectedHeaders = {
-    'X----': 'foo'
-  }
-
-  const originalHeaders = { ...event.headers }
-
-  const resultingEvent = await handler(event, context)
-
-  t.deepEqual(resultingEvent.headers, expectedHeaders)
-  t.deepEqual(resultingEvent.rawHeaders, originalHeaders)
-})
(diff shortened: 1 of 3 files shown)


packages/http-json-body-parser/__tests__/index.js
```diff
@@ -228,3 +228,26 @@ test('It should handle invalid base64 JSON as an UnprocessableEntity', async (t)
...
+    'X----': 'foo'
+  }
+
+  const originalHeaders = { ...event.headers }
+
+  const resultingEvent = await handler(event, context)
+
+  t.deepEqual(resultingEvent.headers, expectedHeaders)
+  t.deepEqual(resultingEvent.rawHeaders, originalHeaders)
+})
```
willfarrell, line 253: I think this was added to the wrong file. Can you do a quick update please.

willfarrell (changes requested):

packages/http-json-body-parser/__tests__/index.js
ScoobySmack2612: Thank you, sorry about that.

willfarrell:
`X----` header? Can say I've never seen that before. I would still tell the client to update `X----` to something more standard. Thanks for the PR and quick update.

ScoobySmack2612:
> `X----` header? Can say I've never seen that before. I would still tell the client to update `X----` to something more standard. Thanks for the PR and quick update.

Its definitely not, I suspect they have just enough of a proxy configuration to get themselves in trouble. Thanks for a speedy review!

Issue #1176, opened 2024-02-15
TypeScript compile error in @middy/core 5.2.4


cm-rwakatsuki:
**Describe the bug**
After updating @middy/core from 5.2.3 to 5.2.4, an error occurred in TypeScript compile (tsc).

**To Reproduce**
How to reproduce the behaviour:

Sample code 

- packages/server/src/lambda/handlers/api-gateway/rest-api/router.ts

```ts
import serverlessExpress from "@codegenie/serverless-express";
import middy from "@middy/core";
import express from "express";

const app = express();

export const handler = middy(serverlessExpress({ app }));
```

Thrown error

```shell
tsc --noEmit

src/lambda/handlers/api-gateway/rest-api/router.ts:11:30 - error TS2345: Argument of type 'Handler<any, any> & ConfigureResult<any, any>' is not assignable to parameter of type 'PluginObject | MiddyInputHandler<any, any, Context> | undefined'.
  Type 'Handler<any, any> & ConfigureResult<any, any>' is not assignable to type 'MiddyInputHandler<any, any, Context>'.
    Types of parameters 'callback' and 'opts' are incompatible.
      Type 'MiddyHandlerObject' is not assignable to type 'Callback<any>'.
        Type 'MiddyHandlerObject' provides no match for the signature '(error?: string | Error | null | undefined, result?: any): void'.

11 export const handler = middy(serverlessExpress({ app }));
```

**Expected behaviour**
tsc can run without errors.

**Environment (please complete the following information):**

- Node.js: 20.9.0
- Middy: 5.2.4
- AWS SDK 2.1450.0

lynxSven:
Is this merge here the issue? https://github.com/middyjs/middy/pull/1165 
The removal of the callback? Have the same issue with https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7bd4a02ba62ede50a6198bd5b10712ae1a2fe6c8/types/aws-lambda/handler.d.ts#L84 the aws-lambda handler.
Still has a callback but it doesn't match the new schema

willfarrell:
Thanks for reporting. Would someone like to put a PR together for this fix?

sekhavati:
Same issue is reproducible with the following basic Lambda:

```typescript
import { type Handler } from "aws-lambda";
import middy from "@middy/core";

const baseHandler: Handler = async (event) => {
  console.log("Hello world");
};

export const handler = middy(baseHandler); // <== error here as described in issue thread
```

[...]

PR #1186, opened 2024-03-04
fix: #1176 add callback argument to support aws types

phmasek:
Fixes https://github.com/middyjs/middy/issues/1176 by adding an additional callback field to match the types set by the AWS library. Retain the `opt` object as the final argument.


willfarrell:
Thanks for the PR. However, `callback ` was deprecated a few versions back when middy transitioned away from using it in favour of promises and streams. Sorry, we don't plan to bring it back at this time.

phmasek:
@willfarrell could you please provide a suggestion for how we can resolve the types breaking in the latest version described in the issue I referenced?

Issue #1189, opened 2024-03-14
Streamified response with empty body ignores other response attributes


mdenkinger:
**Describe the bug**

When a Lambda function in invoke mode `RESPONSE_STREAM` and  `streamifyResponse=true`  returns a response with an empty body (`null` or `''`), then the other response attributes like statusCode or headers are ignored. 
  

**To Reproduce**

Sample code:
```
import { createRequire } from 'module';const require = createRequire(import.meta.url);
import middy from "@middy/core";
import { Readable } from "stream";

function healthcheck() {
  return () => {
    return handler();
  };
}
async function handler() {
  return {
    statusCode: 301,
    body: '', 
    headers: {
      location: 'https://example.com/'
    }
  };
}

var middyHandler = middy({
  streamifyResponse: true
})
.handler(healthcheck());
export {
  middyHandler as handler
};
```

Empty body cause the Lambda to respond with 200 instead if 301
```
{
    statusCode: 301,
    body: '', 
    headers: {
      location: 'https://example.com/'
    }
  }
```

Response headers from client: 
```
>
< HTTP/1.1 200 OK
< Date: Thu, 14 Mar 2024 17:23:24 GMT
< Content-Type: application/octet-stream
< Transfer-Encoding: chunked
< Connection: keep-alive
< x-amzn-RequestId: xxxxx
< X-Amzn-Trace-Id: root=xxx
<
```

e.g. blank body (one space) causes the Lambda to respond correct with 301 and also the headers are present:
```
{
    statusCode: 301,
    body: ' ', 
    headers: {
      location: 'https://example.com/'
    }
  }
```
Response headers from 
```
>
< HTTP/1.1 301 Moved Permanently
< Date: Thu, 14 Mar 2024 17:23:42 GMT
< Content-Type: application/json
< Transfer-Encoding: chunked
< Connection: keep-alive
< x-amzn-RequestId: xxxx
< location: https://example.com/
< X-Amzn-Trace-Id: root=xxxx
<
```

**Expected behaviour**

The other response attributes like statusCode, headers,cookies are return with an empty body.

**Environment (please complete the following information):**

- Node.js: 20
- Middy: 5.2.0
- AWS SDK 3.533.0

**Additional context**
-

willfarrell:
Directly to Function URL or through API Gateway?

> Empty body cause the Lambda to respond with 200 instead if 301

[...]

PR #1190, opened 2024-03-15
fix: middyjs#1189 write prelude in streamifyResponse with string body

mdenkinger:
see: #1189


willfarrell:
Thanks for looking deeper into this!

PR #1192, opened 2024-03-16
fix: allow ALBEvent type in http-json-body-parser

connorreitz:
### Context

According to the [docs for using middy with an ALBEvent](https://middy.js.org/docs/events/application-load-balancer/), we can use the `httpJsonBodyParserMiddleware` to parse the body of an `ALBEvent`. However, the `ALBEvent` type isn't included in the type definition for the `httpJsonBodyParserMiddleware` function, so if we are using Typescript, the compiler will get mad. This update will allow us to `.use(httpJsonBodyParserMiddleware)` on `ALBEvent` handlers :)


packages/http-json-body-parser/index.d.ts
```diff
@@ -8,6 +8,6 @@ interface Options {
 
 export type VersionedApiGatewayEvent = APIGatewayEvent | APIGatewayProxyEventV2
```
willfarrell, line 9: Should `| ALBEvent` go here instead?

willfarrell:
Thanks for the PR! Just a quick question above before I merge it in.

packages/http-json-body-parser/index.d.ts
connorreitz: I was thinking about that too, the only reason I added it like I did instead is that ALBEvent isn't technically an ApiGatewayEvent. Maybe we could rename `VersionedApiGatewayEvent`:
 `export type RequestEvent = APIGatewayEvent | APIGatewayProxyEventV2 | ALBEvent` 
Or something along those lines. Let me know what you think!
`

willfarrell: `RequestEvent ` Sounds good to me

PR #1193, opened 2024-03-20
Update aws-lambda-powertools imports

giufus:
Hi,

Middy looks like super cool, I can't wait to try it. 
Meanwhile I discovered an import that can be updated in the logging module. 
Let me know if it works for you too.

All the best


willfarrell:
Thanks for the PR to update our docs. Looks like `aws-lambda-powertools` docs have been update since this was last updates.


cc @dreamorosi

PR #1198, opened 2024-03-25
fix(types): typescript compile error

### LINKED ISSUE #1176: TypeScript compile error in @middy/core 5.2.4

lmammino:
This should fix #1176.

I am not entirely sure this won't cause other weird typing side effects, so I'd love it if someone with some TypeScript expertise could do a thorough review.

CC @naorpeled

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -168,6 +168,7 @@ declare function middy<
   TInternal extends Record<string, unknown> = {}
 > (
   handler?:
+  | LambdaHandler<TEvent, TResult>
   | MiddlewareHandler<LambdaHandler<TEvent, TResult>, TContext>
   | PluginObject,
   plugin?: PluginObject
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,8 @@
         "speedscope": "^1.0.0",
         "tinybench": "^2.5.1",
         "ts-standard": "^12.0.0",
-        "tsd": "^0.30.0"
+        "tsd": "^0.30.0",
+        "typescript": "^5.4.3"
       },
       "engines": {
         "node": ">=18"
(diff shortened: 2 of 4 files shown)


packages/core/index.d.ts
```diff
@@ -168,6 +168,7 @@ declare function middy<
   TInternal extends Record<string, unknown> = {}
 > (
   handler?:
+  | LambdaHandler<TEvent, TResult>
```
tonivdv, line 171: Seems to solve the typing issue indeed ...

naorpeled (approved):

naorpeled:
Hey @lmammino,
sorry for the huge delay.
Looks good to me.

PR #1199, opened 2024-03-26
chore: update Powertools docs

dreamorosi:
This is a follow up PR to #1193.

In that PR a community member already updated the imports to reflect the changes in v2. This PR instead:
- updates the note at the top of the page to reflect that Powertools is compatible with v4
- adds mentions to new utilities released since last update (link only)
- updates the base url to the Powertools docs to reflect the new scheme (old url still worked thanks to redirects on our side)

For added context, we released v2 a few weeks ago but we purposefully soft launched it and waited to update docs in other repos (this being one of them) to hash out any issues.

This PR will close aws-powertools/powertools-lambda-typescript#2124

### DIFF
--- a/website/docs/integrations/lambda-powertools.md
+++ b/website/docs/integrations/lambda-powertools.md
@@ -8,7 +8,7 @@ You can use Powertools for AWS in both TypeScript and JavaScript code bases.
 
 :::note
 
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for updating these!

PR #1202, opened 2024-03-30
String like errors handling

randomhash:
Basically, inject original error if you can do this

### DIFF
--- a/packages/core/__tests__/index.js
+++ b/packages/core/__tests__/index.js
@@ -717,8 +717,8 @@ globalThis.awslambda = {
   }
 }
 
-function createResponseStreamMockAndCapture() {
-  function processChunkResponse(chunkResponse) {
+function createResponseStreamMockAndCapture () {
+  function processChunkResponse (chunkResponse) {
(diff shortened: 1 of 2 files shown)


randomhash:
@willfarrel please merge at your convenience, e.originalError is breaking our production functionality, and it would be tough to resolve the issue on our side, would appreciate if we could keep it unbreaking for some some time

randomhash:
Last commit is not required, but it's nice to have, as it was a bit hard navigate the code while there was inline chunky function

randomhash:
Also it would be amazing if this change could be backported to middy 4

vladholubiev (approved):

randomhash:
@willfarrell any chance to merge it today ?

willfarrell:
Thanks for your patience, I was off on holiday. Spoke with @lmammino and we agree that this isn't good practice and don't think we should support it in core at this time.

randomhash:
@willfarrell 
but your core is modifying the actual error message, it does not make any sense to add such hidden functionality without the possibility to disable it

PR #1203, opened 2024-04-03
fix: add in retry for expired signatures

huyhua-nvg:
Porting the expired signature fix to v4.x.x as we could not switch to middy v5 at the moment


willfarrell:
Thanks for the PR. Unfortunately we're no longer support v4.x and will not be releasing any new versions at this time.

PR #1205, opened 2024-04-19
Update sqs-partial-batch-failure.md

sharifhh:
I guess it is worth mentioning for users that when they use the Sample Usage, there events are being executed in parallel, so if they are using SQS FIFO then it might produce some bugs, because batches in FIFO sends messages from same MessageGroupID and assume you are handling them in sequential.

So just an addition for newbies like me who spent hours trying to solve this :)


willfarrell:
Thanks for the PR! Great addition to the docs.

Issue #1175, opened 2024-02-12
Restrict types to `APIGatewayProxyResultV2` or `APIGatewayProxyResult`


jlarmstrongiv:
**Describe the bug**

The TypeScript types indicate properties exist in Middy handlers that do not.

**To Reproduce**
How to reproduce the behaviour:

1. With a [function url](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-payloads) that uses the [payload version format v2](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format) with middy

**Expected behaviour**
If I specify `APIGatewayProxyEventV2` and `APIGatewayProxyResultV2` in `export const middy = middyCore<APIGatewayProxyEventV2, APIGatewayProxyResultV2>()`, then the handlers should only use the v2 types.

**Environment (please complete the following information):**

- Node.js: v20.11.0
- Middy: 5.2.4
- AWS SDK 2.1551.0

**Additional context**
The actual types include the old event formats 
```
MiddyfiedHandler<APIGatewayProxyEventV2 & Omit<APIGatewayProxyEvent, "body"> & { body: JsonValue; } & APIGatewayProxyEvent & ... 4 more ... & { ...; }, APIGatewayProxyResultV2, Error, Context, {}>.handler: <unknown>(handler: MiddyInputHandler<APIGatewayProxyEventV2 & Omit<APIGatewayProxyEvent, "body"> & {
    body: JsonValue;
} & APIGatewayProxyEvent & ... 4 more ... & {
    ...;
}, APIGatewayProxyResultV2, Context>) => MiddyfiedHandler<...>
```

b3nk3:
I'm running into some errors that are very similar to this, when using the `http-router` and it stems from using a middleware on the handlers, as the handler changes the types, rightly so, but the typing of the `routes` variable is unable to pick this up...

The code works, but I kinda have to ignore the typescript on that level.

Given the number of different middlewares, and their various outputs - this is not an easy thing to fix.

[...]

PR #1212, opened 2024-05-15
fix(packages/http-router): refine return type to include event and response

naorpeled:
resolves https://github.com/middyjs/middy/issues/1175#issuecomment-1945658104


lmammino (approved):
Lgtm! Thanks a lot for taking care of this, @naorpeled 🙏

PR #1217, opened 2024-06-15
fix(core)!: resolve type inference issues with input handler

### LINKED ISSUE #1182: TypeScript error when using `S3Handler` type
**Describe the bug**

**To Reproduce**

```ts
const lambdaHandler: S3Handler = async (event: S3Event): Promise<void> => {
  await Promise.all(event.Records.map(processRecord));
};

export const handler = middy().use(eventNormalizer()).handler(lambdaHandler);
```

```sh
Argument of type 'S3Handler' is not assignable to parameter of type 'MiddyInputHandler<S3Event, any, Context>'.
  Types of parameters 'callback' and 'opts' are incompatible.
    Type 'MiddyHandlerObject' is not assignable to type 'Callback<void>'.
      Type 'MiddyHandlerObject' provides no match for the signature '(error?: string | Error | null | undefined, result?: void | undefined): void'.
```

**Expected behaviour**

- Node.js: [e.g. 20]
- Middy: [e.g. 5.0.0]
- AWS SDK [e.g. 3.0.0]

**Additional context**

naorpeled:
fixes #1182

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -96,12 +96,13 @@ export interface MiddyfiedHandler<
   before: AttachMiddlewareFn<TEvent, TResult, TErr, TContext, TInternal>
   after: AttachMiddlewareFn<TEvent, TResult, TErr, TContext, TInternal>
   onError: AttachMiddlewareFn<TEvent, TResult, TErr, TContext, TInternal>
-  handler: <TAdditional>(
+  handler: <TInputHandlerEventProps, TInputHandlerResultProps>(
     handler: MiddlewareHandler<
-    LambdaHandler<TEvent & TAdditional, TResult>,
+    LambdaHandler<TInputHandlerEventProps, TInputHandlerResultProps>,
     TContext
     >
-  ) => MiddyfiedHandler<TEvent, TResult, TErr, TContext, TInternal>
+    | LambdaHandler<TInputHandlerEventProps, TInputHandlerResultProps>
+  ) => MiddyfiedHandler<TInputHandlerEventProps, TInputHandlerResultProps, TErr, TContext, TInternal>
 }
 
 declare type AttachMiddlewareFn<
--- a/packages/core/index.test-d.ts
+++ b/packages/core/index.test-d.ts
@@ -408,5 +408,5 @@ const s3Handler: S3Handler = async (event) => {
   await Promise.all(event.Records.map(async () => await Promise.resolve()))
 }
 
-const handler1182 = middy(s3Handler)
+const handler1182 = middy().handler(s3Handler)
 expectType<MiddyfiedHandler<S3Event, void, Error, Context, {}>>(handler1182)


packages/core/index.d.ts
```diff
@@ -96,12 +96,13 @@ export interface MiddyfiedHandler<
...
   after: AttachMiddlewareFn<TEvent, TResult, TErr, TContext, TInternal>
   onError: AttachMiddlewareFn<TEvent, TResult, TErr, TContext, TInternal>
-  handler: <TAdditional>(
+  handler: <TInputHandlerEventProps, TInputHandlerResultProps>(
     handler: MiddlewareHandler<
-    LambdaHandler<TEvent & TAdditional, TResult>,
+    LambdaHandler<TInputHandlerEventProps, TInputHandlerResultProps>,
     TContext
     >
-  ) => MiddyfiedHandler<TEvent, TResult, TErr, TContext, TInternal>
```
naorpeled, line 104: @lmammino does this make sense to you? 🙏 
I believe that this is the correct behavior

packages/core/index.d.ts
```diff
@@ -96,12 +96,13 @@ export interface MiddyfiedHandler<
...
+    | LambdaHandler<TInputHandlerEventProps, TInputHandlerResultProps>
+  ) => MiddyfiedHandler<TInputHandlerEventProps, TInputHandlerResultProps, TErr, TContext, TInternal>
```
naorpeled, line 105: @lmammino does this return type make sense to you? 🙏 
I'd expect it to be taken from the input handler (`.handler<..params>`) fn and not the initial input (`middy<...params>`).

naorpeled:
@lmammino bumping this 🙏

lmammino (approved):

lmammino:
LGTM! Thanks a mill @naorpeled !

@willfarrell can I leave it to you to merge and release based on your priorities?

jlarmstrongiv:
I think this broke the types in my project, as I’m now getting the error:

```
Binding element 'headers' implicitly has an 'any' type.
```

Whereas downgrading to `5.4.1` properly types the headers as:

```
headers: APIGatewayProxyEventHeaders | (APIGatewayProxyEventHeaders & ALBEventHeaders)
```

instead of `any`

naorpeled:
> I think this broke the types in my project, as I’m now getting the error:
> 
> ```
> Binding element 'headers' implicitly has an 'any' type.
> ```
> 
> Whereas downgrading to `5.4.1` properly types the headers as:
> 
> ```
> headers: APIGatewayProxyEventHeaders | (APIGatewayProxyEventHeaders & ALBEventHeaders)
> ```
> 
> instead of `any`

Hey @jlarmstrongiv,
can you please share a reproduction to your issue?

jlarmstrongiv:
Sure, here is the reproduction @naorpeled I removed the middleware to keep the reproduction as simple as possible https://github.com/jlarmstrongiv/middy-typescript-headers-bug 👍

naorpeled:
> Sure, here is the reproduction @naorpeled I removed the middleware to keep the reproduction as simple as possible https://github.com/jlarmstrongiv/middy-typescript-headers-bug 👍
> 
> You will be able to see how `5.4.1` works, while `5.4.2` causes type errors.

I see, what you need to do is pass the generic params into `.handler` and that would resolve your issue

jlarmstrongiv:
Hey @naorpeled do you have an example of passing those generic params into `.handler`?  It’s still erroring with the same error for me:

```ts
export const handler = middy.handler<
  APIGatewayProxyEventV2,
  APIGatewayProxyResultV2
  // Binding element 'headers' implicitly has an 'any' type.
>(({ headers }) => ({
  body: JSON.stringify(headers, undefined, 2),
  statusCode: 200,
}));
```

I also wish I could define it once with middy core like I used to, rather than having to define it on each and every lambda function handler

PR #1226, opened 2024-07-15
docs: update Powertools compatibility

dreamorosi:
This PR updates the documentation page dedicated to Powertools for AWS Lambda specifying that it now officially supports both v4.x and v5.x of Middy.js.

While there, I also added a link to the new Parser utility that was launched since the last update. This utility comes with a Middy.js-compatible middleware, so when it goes GA in the next weeks I might send another PR to add a section to the page.


willfarrell:
Awesome! Thanks!

PR #1227, opened 2024-07-18
fix(log): display nested objects in lambda request and response in @middy/input-output-logger

minjie0501:
Currently, nested objects, such as identity within the requestContext of the request event, are displayed as [Object] in the logs. This makes it difficult to debug and understand the full structure of the request and response objects.

Updating the logging mechanism to stringify nested objects using JSON.stringify(cloneMessage, null, 2) ensures that all nested objects are properly represented in the logs, making them more readable and informative.

### DIFF
--- a/packages/input-output-logger/index.js
+++ b/packages/input-output-logger/index.js
@@ -35,7 +35,7 @@ const inputOutputLoggerMiddleware = (opts = {}) => {
       cloneMessage = structuredClone(message) // Full clone to prevent nested mutations
       omit(cloneMessage, { [param]: omitPathTree[param] })
     }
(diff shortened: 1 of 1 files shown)


willfarrell:
Thanks for the PR! I like the idea of having the default be more usable out of the box.

See my inline comment about proposed changes. Unit tests will also need to be updated before release.

minjie0501:
> Thanks for the PR! I like the idea of having the default be more usable out of the box.
> 
> See my inline comment about proposed changes. Unit tests will also need to be updated before release.

I don't see any inline comments 🤔

PR #1238, opened 2024-09-01
Feature zstd support for http-content-encoding

willfarrell:
Does anyone need this?

Downside, adds 2MB to the packages, zstd isn't smaller than brotli in my testing.

Will leave as PR, only Chrome supports right now.


willfarrell:
Natively supported in node22, will be in version 7 of middy

Issue #1228, opened 2024-07-26
Handler return is not type-checked.


canassa:
When using TypeScript, the return of a handler is not being validated by the type checker.

**To Reproduce**

Add this file

```typescript
import type {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
import middy from "@middy/core";


export const handler = middy<APIGatewayProxyEvent, APIGatewayProxyResult>()
    .handler(async (req, context) => {
        return 123
    })
```

- Run the type checker.
- Observe that no error is detected.

**Expected behaviour**
The type checker should have throw an error since `123` does not match the `APIGatewayProxyResult`.

**Note:** I know that this can be workaround by manually adding the return type to the handler function, but that's more error prone.

**Environment (please complete the following information):**

- Node.js: 20
- Middy: 5.4.5

canassa:
This seems to be a regression with Middy 5. Middy 4 is able to correctly validate the type:

```
error TS2345: Argument of type '(req: APIGatewayProxyEvent, context: Context) => Promise<number>' is not assignable to parameter of type 'MiddyInputHandler<APIGatewayProxyEvent, APIGatewayProxyResult, Context>'.
  Type 'Promise<number>' is not assignable to type 'void | APIGatewayProxyResult | Promise<APIGatewayProxyResult>'.
    Type 'Promise<number>' is not assignable to type 'Promise<APIGatewayProxyResult>'.
      Type 'number' is not assignable to type 'APIGatewayProxyResult'.

8     .handler(async (req, context) => {
               ~~~~~~~~~~~~~~~~~~~~~~~~~
```

willfarrell:
A PR is welcome for this.

PR #1241, opened 2024-09-18
handler return type checking #1228

jerizm:
I think what was happening was because `, TInputHandlerResultProps = TResult` was passed it allowed the return type to be inferred or overridden by the handler. 

After removing `, TInputHandlerResultProps = TResult` from the handler the return type will be enforced by what is passed in to `Middy<Event, Result...` 

This does mean, however, you cannot override the handler return type.


willfarrell:
Thanks for the PR! However, I have chosen to merge #1244 instead as it's not a breaking change and is passing TS unit tests.

jerizm:
@willfarrell great we're glad it fixes the return type enforcement. #1244, looks like it has the same issue with not being able to override the handler return type either.

PR #1244, opened 2024-10-15
:label: ensures that handler return is properly typed

### LINKED ISSUE #1228: Handler return is not type-checked.

canassa:
- This change addresses an issue where the return type of handlers was not being correctly enforced by the TypeScript compiler. As far as I could tell, the issue is caused by the `infer` keyword not working with generic types.
- Added `TResult` as a type parameter to `MiddlewareHandler` to explicitly specify the handler's return type.
- Modified the `middy` function signature to pass `TResult` to `MiddlewareHandler`.
- By explicitly specifying `TResult`, we ensure that the compiler can correctly enforce the handler's return type.

closes #1228

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -99,7 +99,7 @@ export interface MiddyfiedHandler<
   handler: <TInputHandlerEventProps = TEvent, TInputHandlerResultProps = TResult>(
     handler: MiddlewareHandler<
     LambdaHandler<TInputHandlerEventProps, TInputHandlerResultProps>,
-    TContext
+    TContext, TResult
     >
   ) => MiddyfiedHandler<TInputHandlerEventProps, TInputHandlerResultProps, TErr, TContext, TInternal>
 }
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for the PR!

canassa:
@willfarrell Thanks! Do you know if this will be included in a release soon? or only after 6.X?

Issue #1239, opened 2024-09-03
`Argument of type 'ScheduledHandler' is not assignable to parameter of type 'MiddyInputHandler<EventBridgeEvent<"Scheduled Event"`


garysassano:
**Describe the bug**

**To Reproduce**
How to reproduce the behaviour:

#### index.ts

```ts
import { Logger } from "@aws-lambda-powertools/logger";
import { injectLambdaContext } from "@aws-lambda-powertools/logger/middleware";
import { ScheduledHandler } from "aws-lambda";
import middy from "@middy/core";
import httpErrorHandler from "@middy/http-error-handler";

const logger = new Logger();

const lambdaHandler: ScheduledHandler = async () => {
  const response = {
    statusCode: 200,
    body: "Hello from Lambda!",
  };

  console.log(response);
};

export const handler = middy()
  .use(injectLambdaContext(logger, { logEvent: true }))
  .use(httpErrorHandler())
  .handler(lambdaHandler);
```

**Expected behaviour**
No error should be thrown, instead you get this:

```sh
Argument of type 'ScheduledHandler' is not assignable to parameter of type 'MiddyInputHandler<EventBridgeEvent<"Scheduled Event", any>, void, Context>'.
  Types of parameters 'callback' and 'opts' are incompatible.
    Type 'MiddyHandlerObject' is not assignable to type 'Callback<void>'.
      Type 'MiddyHandlerObject' provides no match for the signature '(error?: string | Error | null | undefined, result?: void | undefined): void'.
```

**Environment (please complete the following information):**

- Node.js: 20.17.0
- Middy: 5.4.7
- AWS SDK: 3.644.0

**Additional context**

You can fix the issue by avoiding to use the `.handler()`, like this:

```ts
export const handler = middy(lambdaHandler)
  .use(injectLambdaContext(logger, { logEvent: true }))
  .use(httpErrorHandler())
```

willfarrell:
Closing as stale.

garysassano:
I recommend updating all the docs to consistently use `middy(lambdaHandler)` instead of `middy().handler(lambdaHandler)`, as the latter approach is half-broken and may not always work.

PR #1257, opened 2024-11-13
fix(core) #1239

R11baka:
fix https://github.com/middyjs/middy/issues/1239
TLDR added type for handler method


willfarrell:
Closing, not passing typings tests

Issue #1266, opened 2024-12-13
http-json-body-parser: Body should be allowed to be undefined when Content-Length is 0 or missing


bpadgette:
**Is your feature request related to a problem? Please describe.**

The middy middleware [http-json-body-parser](https://github.com/middyjs/middy/blob/da5d1939f0f255f21060fd69bc1ad818af400f19/packages/http-json-body-parser/index.js)  returns an `HTTP 415 Unprocessable Entity` response whenever the request body is undefined.

_This case is tested: awesome_

As test "`It should handle undefined as an UnprocessableEntity`" in [packages/http-json-body-parser/\__\tests\_\_/index.js](https://github.com/middyjs/middy/blob/da5d1939f0f255f21060fd69bc1ad818af400f19/packages/http-json-body-parser/__tests__/index.js) states, an undefined request body will result in the HTTP 415.

**I propose that an undefined body is not a generally justifiable cause for an HTTP 415.**

Consider this AWS API Gateway scenario:

Send a GET request with `content-type: application/json` but no `content-length` header, and AWS Gateway would parse that request and pass it on to Middy with no body key (meaning `event.body === undefined`) in [packages/http-json-body-parser/index.js](https://github.com/middyjs/middy/blob/da5d1939f0f255f21060fd69bc1ad818af400f19/packages/http-json-body-parser/index.js).

**Describe the solution you'd like**

In the middleware [http-json-body-parser](https://github.com/middyjs/middy/blob/da5d1939f0f255f21060fd69bc1ad818af400f19/packages/http-json-body-parser/index.js):

```
IF `event.body === undefined` AND the content-length header is 0 or undefined
   ignore event, do not try to parse as JSON body
```

**Describe alternatives you've considered**

My team currently defines our own http-json-body-parser wrapper that implements a version of this solution.

We also believe that content-type without content-length is an anti-pattern _but not illegal_ in its own right, but it is besides the point since we had undefined body fields making it to middy, passing the basic parsing of AWS API Gateway to get there.

**Additional context**

My team uses middy on a project consuming HTTP events from AWS API Gateway, and we noticed that the `body` keys of our events were undefined which caused the aforementioned HTTP 415 to show on several GET routes. 

Our SDK client in development was sending GET requests with a default inserted `content-type: application/json` header but no `content-length` header, and AWS API Gateway passed events with no body key to our middy middleware object. 

[...]

PR #1267, opened 2024-12-13
feat(http-json-body-parser): #1266 allow undefined body when no content length

bpadgette:
See issue https://github.com/middyjs/middy/issues/1266

> **Describe the solution you'd like**
>
> In the middleware [http-json-body-parser](https://github.com/middyjs/middy/blob/da5d1939f0f255f21060fd69bc1ad818af400f19/packages/http-json-body-parser/index.js):
>
> IF `event.body === undefined` AND the content-length header is 0 or undefined
   ignore event, do not try to parse as JSON body


willfarrell:
Thanks for PR, unfortunately I'm closing per discussion on issue.

Issue #1260, opened 2024-12-03
Cloudformation custom resource handler


GentileFulvio:
**Is your feature request related to a problem? Please describe.**
We are currently implementing custom resources for out platform. We added our own custom wrapper implementation to make handling these custom resources less bloated inside our code and noticed that middy doesn't have anything for custom resources already.

**Describe the solution you'd like**
It would be nice to have an abstraction which handles custom resources for us, the following is what our abstraction looks like, we are then using the created handler with middy so we can leverage the middleware there as well

```typescript
interface ExpectedResourceOutput {
	SecretArn: string;
}

const myHandler: CustomResourceHandler<ExpectedResourceOutput> = async (event) => {
	// Code goes here

	return {
		// This is automatically added to the output in cloudformation
		SecretArn: "Hello-world",
	};
};

// The handler wrapper is responsible to catch errors in the handler and automatically report to cloudformation
// It is also responsible to report the success and add the outputs on the cloudformation resource
export const handler = createSNSCustomResourceHandler({
	create: myHandler,
});
```

The return value of the MyHandler function will be used as the output of the custom resource. Note that in this implementation we are creating a custom resource via sns, the handler should be able to handle requests via sns and to lambda directly

**Describe alternatives you've considered**
We currently created our own wrapper to simplify the code

**Additional context**
I'm willing to try and work out something in my free time as a suggestion. I was just wondering if this is something that should be added to middy or not

willfarrell:
I was reading through https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html. In the event examples included I saw that some properties are returned from the request, along with `Status` as a fixed value that could be put into a middleware to trim down the handler response a little bit. Is this what you were envisioning. 

```
{
   "Status" : "SUCCESS",
   "RequestId" : "(copied from request)",
   "LogicalResourceId" : "(copied from request)",
   "StackId" : "(copied from request)",
   "...
}
```

GentileFulvio:
I had something like this in mind where we can abstract the reporting to the responseUrl so the developer simply needs to define `create, update, delete` functions where the reuturn value are automatically used as data to report to cloudformation. The abstraction of my handler currently looks like this
```
const createCustomResourceHandler = <TEvent = unknown, TResult = unknown>(
	handlers: Handlers<TResult>,
	parser: EventParser<TEvent>,
) => {
	return async (event: TEvent, context: Context): Promise<void> => {
[...]

PR #1268, opened 2024-12-13
Feature/cloudformation custom resource router

willfarrell:
#1260

### DIFF
--- a/packages/cloudformation-router/index.js
+++ b/packages/cloudformation-router/index.js
@@ -1,0 +1,52 @@
+import { createError } from '@middy/util'
+const defaults = {
+  routes: [],
+  notFoundResponse: ({ requestType }) => {
+    const err = createError(404, 'Route does not exist', {
+      cause: {
+        package: '@middy/cloudformation-router',
+        data: { requestType }
+      }
+    })
+    throw err
+  }
+}
+const cloudformationCustomResourceRouteHandler = (opts = {}) => {
+  if (Array.isArray(opts)) {
+    opts = { routes: opts }
+  }
+  const { routes, notFoundResponse } = { ...defaults, ...opts }
+
+  const routesStatic = {}
+  for (const route of routes) {
+    const { requestType, handler } = route
+
+    // Static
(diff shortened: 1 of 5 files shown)


website/docs/routers/cloudformation-router.md
```diff
@@ -0,0 +1,79 @@
...
+  {
+    requesType: 'Create',
+    handler: createHandler
+  },
+  {
+    requesType: 'Update',
+    handler: updateHandler
+  },
+  {
+    routeKey: 'delete',
```
DavidWells, line 71:
```suggestion
    routeKey: 'Delete',
```

packages/cloudformation-router/index.js
```diff
@@ -0,0 +1,46 @@
...
+    // Static
+    const handler = routesStatic[requestType]
+    if (typeof handler !== 'undefined') {
+      const response = handler(event, context, abort)
+      response.Status ??= 'SUCCESS'
+      response.RequestId ??= event.RequestId
+      response.LogicalResourceId ??= event.LogicalResourceId
+      response.StackId ??= event.StackId
+      return response
+    }
```
GentileFulvio, line 39: It would be nice if it also reported to the responseUrl, I feel like even though the abstraction is nice, it doesn't really add that much value currently as it's still the responsibility of the developer to correctly handle all errors and respond.

The auto reporting implementation is more what I was looking for honestly

packages/cloudformation-router/index.js
```diff
@@ -0,0 +1,46 @@
...
+    const { requestType, handler } = route
+
+    // Static
+    routesStatic[requestType] = handler
+  }
+
+  return (event, context, abort) => {
+    const { RequestType: requestType } = event
+    if (!requestType) {
+      return notFoundResponse({ requestType })
```
GentileFulvio, line 27: I would for example expect that here we report to cloudformation that we have a failure, because if we don't do that cloudformation will simply wait x time until it timeouts whilst we already know that we won't be handling the request

PR #1280, opened 2025-02-26
fix: support Web API ReadableStream when streamifyResponse:true

### LINKED ISSUE #1278: streamifyResponse does not work with ReadableStream
**Describe the bug**
I'm trying to return a streaming response from AWS Lambda using [createDataStream](https://sdk.vercel.ai/docs/reference/ai-sdk-ui/create-data-stream#createdatastream) from `ai` package. However, the function fails with the error: `"handler response not a ReadableStream"`

**To Reproduce**
Sample code:
```typescript
import { createDataStream } from "ai";

function chat() {
  return createDataStream({
    execute: async (stream) => {
      stream.write("Hello");
    },
  });
}

export const handler = middy({ streamifyResponse: true }).handler(chat);
```

**Expected behaviour**
AWS Lambda should not fail if handler returns instance of ReadableStream

- Node.js: 20
- Middy: 5.5
- AWS SDK 3.726.0

**Additional context**
The following check is reason of the bug https://github.com/middyjs/middy/blob/main/packages/core/index.js#L67

If I change it to `if (handlerBody._readableState || handlerBody instanceof ReadableStream)` everything works fine.

apalchys:
Fixes #1278

@willfarrell please check my changes and comments in tests.

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -1,5 +1,6 @@
 /* global awslambda */
 import { Readable } from 'node:stream'
+import { ReadableStream } from 'node:stream/web'
 import { pipeline } from 'node:stream/promises'
 import { setTimeout } from 'node:timers'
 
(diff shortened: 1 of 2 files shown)


packages/core/index.js
```diff
@@ -1,5 +1,6 @@
 /* global awslambda */
 import { Readable } from 'node:stream'
+import { ReadableStream } from 'node:stream/web'
```
willfarrell, line 3: Oooh, it's using web streams. That's why.

willfarrell:
Thanks for the PR!

PR #1283, opened 2025-03-02
chore: add missing input type supporting notFoundResponse in http-router

andre2w:
The `httpRouterHandler` interface is overloaded to allow the user to pass an object with the routes and a function that throws an error when the route is not found. 

This overloading was not being declared in the types, forcing the usage of `as any` when you need to pass the `notFoundResponse` method.


willfarrell:
Awesome, thanks for the PR.

PR #1285, opened 2025-03-03
fix(http-json-body-parser): allow "undefined" body if disableContentTypeError:true

### LINKED ISSUE #1284: http-json-body-parser fails in 6.1+
**Describe the bug**
`http-json-body-parser` middleware started failing after upgrade from 6.0 to 6.1.4

A check for the presence of body has been added in https://github.com/middyjs/middy/pull/1264.
(see line https://github.com/middyjs/middy/commit/337b5d240a947e86f698195e176278c9fbcc8e0d#diff-ca203e5ad9276745f9782c25c95e987d1f06e00955ac962dce5c10b342461f04R15).
The side effect of this change is that it fails for HTTP GET requests with `disableContentTypeError: true`

_**Use case:**_ 
We have a chain of middlewares (middy and custom) and apply them to all HTTP handlers.

```
function middifier(handler) {
   return middy(handler)
        .use(httpErrorHandler({...})
        .use(httpJsonBodyParser({ disableContentTypeError: true }))
        .use(...)
        .use(...)
        .use(...)
}
```

Previously, if there was no content type header (GET request), `http-json-body-parser` did early return https://github.com/middyjs/middy/blob/6.0.0/packages/http-json-body-parser/index.js#L19

Now `httpJsonBodyParser({ disableContentTypeError: true })` fails for requests without body and content-type header here: https://github.com/middyjs/middy/blob/6.1.4/packages/http-json-body-parser/index.js#L15

**To Reproduce**

1. Add `httpJsonBodyParser({ disableContentTypeError: true })` to your lambda handler
```type
function foo(event) {
}
export const handler = middy(foo).use(httpJsonBodyParser({ disableContentTypeError: true }));
```
2. Call AWS Lambda using GET requests via API Gateway integration

**Expected behaviour**
Consider skipping body check if no body, no content type header and `disableContentTypeError: true`.
Let me know if it makes sense, I can create a pull request

- Node.js: 20
- Middy: 6.1.4
- AWS SDK 3.515

apalchys:
Fixes #1284

### DIFF
--- a/packages/http-json-body-parser/__tests__/index.js
+++ b/packages/http-json-body-parser/__tests__/index.js
@@ -177,6 +177,27 @@ test("It shouldn't process the body and throw error if no header is passed", asy
   }
 })
 
+test('It should handle undefined body if no header', async (t) => {
+  /**
+   * test checks that if the body is undefined, no content-type header is passed and disableContentTypeError is true,
+   * the handler should process the request and do not throw an error
+   */
+  const handler = middy((event) => {
+    return event.body // propagates the body as a response
+  })
+
+  handler.use(jsonBodyParser({ disableContentTypeError: true }))
+
+  // invokes the handler
+  const event = {
+    headers: {},
+    body: undefined
+  }
+
+  const body = await handler(event, defaultContext)
+  equal(body, undefined)
+})
+
 test('It should handle a base64 body', async (t) => {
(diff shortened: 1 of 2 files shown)


willfarrell:
Thanks for the PR, I'll have a release out shortly.

Issue #1293, opened 2025-04-22
Handler event type is not correctly inferred when using type parameters


contexD:
**Describe the bug**
When passing the event type parameter to middy, the event type is not correctly inferred in the handler and results in incompatible event types.

**To Reproduce**
How to reproduce the behaviour:

```ts
// middlewareStack.ts, published as package
import { z, ZodType } from 'zod';
import middy from '@middy/core';

export function middlewareStack<TEventSchema extends ZodType, TResponseSchema extends ZodType>({
    eventSchema,
    responseSchema,
}: {
    eventSchema: TEventSchema
    responseSchema: TResponseSchema
}) {
    return middy<z.infer<typeof eventSchema>, z.infer<typeof responseSchema>>()
};
```

```ts
// handler.ts
import { middlewareStack } from '@mypackage/middlewarestack';
import { z } from 'zod';

const myEventSchema = z.object({ id: z.string() });
const myResponseSchema = z.object({ success: z.boolean() });

export const handler = middlewareStack({
  eventSchema: myEventSchema,
  responseSchema: myResponseSchema,
}).handler(async (event: z.infer<typeof myEventSchema>) => {
  const { id } = event;

  if (id != '123') {
    return { success: false };
  }

  return { success: true };
});
```

Type error:
`ts: Argument of type '(event: z.infer<typeof myEventSchema>) => Promise<{ success: false; } | { success: true; }>' is not assignable to parameter of type 'MiddyInputHandler<unknown, { success: boolean; }, aws_lambda.Context>'.
  Types of parameters 'event' and 'event' are incompatible.
    Type 'unknown' is not assignable to type '{ id: string; }'. [2345]`

**Expected behaviour**
The event type is correctly inferred from the type variable and the event types are compatible with each other.

**Environment (please complete the following information):**

- Node.js: v22.14.0
- Middy: 6.1.6
- Typescript: 5.7.2
- Zod: 3.24.3

**Additional context**
```ts
// tsconfig
{
  "compilerOptions": {
    "target": "es2022",
    "module": "node16",
    "lib": [
      "ES2022",
      "DOM"
    ],
    "moduleResolution": "node16",
    "rootDir": ".",
    "outDir": "build",
    "allowSyntheticDefaultImports": true,
    "importHelpers": true,
    "sourceMap": true,
    "forceConsistentCasingInFileNames": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "noUnusedLocals": true,
[...]

PR #1295, opened 2025-04-23
fix: specify event type explicitly

contexD:
- fixes the issue of the event type not being correctly inferred by the Typescript compiler since the `infer` keyword does not properly work with generic types

- change adopts the same solution as for [middyjs#1228: Handler return is not type-checked.](https://github.com/middyjs/middy/issues/1228) , that is adding `TEvent` as a type parameter to the `MiddlewareHandler` allows to explicitly indicate the handler's event type

closes [#1293](https://github.com/middyjs/middy/issues/1293)


willfarrell (dismissed):

jonathanstumm (approved):

willfarrell:
Thanks for the PR!

sergiojoker11:
thanks for the PR! Nonetheless, this is a breaking change as per SemVer.

packages/core/index.d.ts
```diff
@@ -161,31 +162,38 @@ declare type UseFn<
...
-	TResult = any,
-> = THandler extends LambdaHandler<infer TEvent, TResult> // always true
-	? MiddyInputHandler<TEvent, TResult, TContext>
-	: never;
+  THandler extends LambdaHandler<any, any>,
+  TContext extends LambdaContext = LambdaContext,
+  TResult = any,
+  TEvent = any
+> =
+  THandler extends LambdaHandler<TEvent, TResult> // always true
```
sergiojoker11, line 170: anchor

PR #1319, opened 2025-05-01
fix: streamify requestEnd hook called before stream completes

### LINKED ISSUE #1317: streamify: requestEnd hook called before stream completes
in a streamified handler, the stream is returned and the handler ends. this means that all the `after` events for middleware are called while the stream is still being generated. this caused a problem for us because one of our middlewares is cleaning up resources that are needed to complete the stream.

we looked into using the `requestEnd` hook... but it's actually invoked when the stream starts. it should probably be called when the stream ends (or possible have other hooks `responseStart` and `responseEnd`)

- Node.js: [22.12.0]
- Middy: [6.1.6]
- AWS SDK [3.799.0]

**Additional context**
this would be very helpful for lambda handlers interacting with AI LLMs, where the stream may involve tool calls that require resources initialized in middleware

willfarrell:
Closes #1317

cc @mdesousa


mdesousa (approved):
this looks great, thanks 👍

PR #1366, opened 2025-06-15
docs(sqs-partial-batch-failure): clarify misleading usage instructions

glhrmv:
---
name: Pull request
about: Pull request
title: 'docs(sqs-partial-batch-failure): clarify misleading usage instructions'
labels: ''
assignees: ''

---

The [current docs page](https://middy.js.org/docs/middlewares/sqs-partial-batch-failure/) includes examples that conflict with how the middleware actually works. Specifically:
- The first two examples show handlers returning `{ batchItemFailures }` but this return shape is not compatible with the middleware.
- The “FIFO queue” example uses `Promise.allSettled()` for parallel execution, which does not preserve processing order, which is something FIFO queues are often used for.

This PR:
- Clarifies that the middleware expects the handler to return an array of `PromiseSettledResult`s, not  `batchItemFailure`s object.
- Moves the current "FIFO" example to a more accurate section for parallel processing.
- Adds a new example for sequential processing, which is appropriate for FIFO queues when order matters.
- Adds guidance on when not to use the middleware (i.e., if you return `batchItemFailure`s manually).

This should avoid subtle bugs in production where failed SQS messages may not be reported properly, or where FIFO semantics are unintentionally broken due to out-of-order side-effects.

Not that I know of.

**Todo List:**

### DIFF
--- a/website/docs/middlewares/sqs-partial-batch-failure.md
+++ b/website/docs/middlewares/sqs-partial-batch-failure.md
@@ -20,45 +20,42 @@ npm install --save-dev @aws-sdk/client-sqs
 
 ## Sample usage
 
-Standrad Queue (All records handled in parallel):
+Parallel processing example (works for Standard queues and FIFO queues _when ordering of side‑effects is not required_):
 
 ```javascript
 import middy from '@middy/core'
 import sqsBatch from '@middy/sqs-partial-batch-failure'
 
-const lambdaHandler = async (event, context) => {
-    const batchItemFailures = [];
-    for (const record of event.Records) {
-        try {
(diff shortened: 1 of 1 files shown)


willfarrell:
Awesome! Thanks for the PR!

All the details are documented here: https://github.com/middyjs/middy/blob/main/docs/CONTRIBUTING.md#5-committing
DCO automatically generates the commands you need for signoff here: https://github.com/middyjs/middy/pull/1366/checks?check_run_id=44132440390

glhrmv:
@willfarrell There we go, I believe it's fixed.

PR #1383, opened 2025-06-25
feat: move from tsd to tstyche appconfig

Puppo:
- [x] Move appConfig from `tsd` to `tstyche`


package.json
```diff
@@ -23,7 +23,9 @@
 		"test:sast:sandworm": "sandworm-audit --skip-all --show-tips false",
 		"test:sast:semgrep": "semgrep scan --config auto",
 		"test:sast:trufflehog": "trufflehog filesystem --only-verified --log-level=-1 ./",
-		"test:types": "ls packages | xargs -I {} tsd packages/{}",
+		"test:types:tsd": "find packages -name \"*.test-d.ts\" -exec dirname {} \\; | sed 's|packages/||' | sort -u | xargs -I {} tsd packages/{}",
+		"test:types:tstyche": "find packages -name \"*.tst.ts\" -exec dirname {} \\; | sed 's|packages/||' | sort -u | xargs -I {} tstyche packages/{}",
```
Puppo, line 27: This one allows us to keep both systems during the migration, and run the tests only where they are available for that framework.

packages/appconfig/index.tst.ts
```diff
@@ -0,0 +1,164 @@
+import { AppConfigDataClient } from "@aws-sdk/client-appconfigdata";
+import middy from "@middy/core";
+import { getInternal } from "@middy/util";
+import type { Context as LambdaContext } from "aws-lambda";
+import { captureAWSv3Client } from "aws-xray-sdk";
+import { expect, test } from "tstyche";
+import appConfig, { appConfigReq } from ".";
```
mrazauskas, line 7: How can it be that extension less imports are still allowed? Ah.. The root `./tsconfig.json` makes this work. It defaults to CJS resolution. Is that useful for any other tooling?

I see two options:

- `npm run test:types:tstyche -- --tsconfig ignore`, simply use TSTyche defaults;
- extend the root TSConfig from [`@tsconfig/node20`](https://www.npmjs.com/package/@tsconfig/node20), since Node.js 20 is the lowest supported version

package.json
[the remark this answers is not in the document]
mrazauskas: It can be you will have to add `tstyche.config.json` with `"checkSourceFiles": false` for now. Otherwise it might dive too deep for this moment.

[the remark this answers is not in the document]
mrazauskas: If #1384 is accepted, there will be no need for `"checkSourceFiles": false`.

willfarrell:
#1384 has been merged in. The new approach looks way easier to understand. With no comments against moving forward, I would say feel free to move forward with all middlewares if you like,

packages/appconfig/index.tst.ts
mrazauskas: The extension less imports should fail after badad62. (Many thanks to @willfarrell!)

PR #1384, opened 2025-06-26
chore: extend from `@tsconfig/node20`

mrazauskas:
This aim of this Pr is to make the root TSConfig stricter.

I would suggest extending from [`@tsconfig/node20`](https://www.npmjs.com/package/@tsconfig/node20). The other changes you see here are the result of this improvement. This should help to modernize the type testing infrastructure.

### DIFF
--- a/packages/validator/transpile.test-d.ts
+++ b/packages/validator/transpile.test-d.ts
@@ -1,5 +1,5 @@
 import { expectType } from "tsd";
-import type { transpileLocale, transpileSchema } from "./transpile.t.ds";
+import { transpileLocale, transpileSchema } from "./transpile.js";
 
 const schema = transpileSchema({ type: "object" }, {});
 expectType<any>(schema);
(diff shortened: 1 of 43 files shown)


packages/validator/transpile.test-d.ts
```diff
@@ -1,5 +1,5 @@
 import { expectType } from "tsd";
-import type { transpileLocale, transpileSchema } from "./transpile.t.ds";
```
mrazauskas, line 2: These are obviously not types. The problem is that current TSConfig defaults to `moduleResolution: "node10"` and that on allows oddities like this one.

willfarrell:
I can take care of it for you, one sec.

mrazauskas:
Thanks for taking care!

Hm.. What is that `Tower`? Sounds interesting.

willfarrell:
https://www.git-tower.com/

I'm sure there is a free alt out there.

PR #1452, opened 2025-08-25
Fix for typings coming back as never

kevupton:
---
name: Pull request
about: Pull request
title: 'Fix for typings coming back as never'
labels: ''
assignees: ''

---

Right now there is a random scenario where this statement is not actually true and it breaks the function typings. Do we need this ternary? If its expected to always be true, can we just remove the ternary?

**Environment:**
 - Node.js: [e.g. 22]
 - Middy: [e.g. 6.0.0]
 - AWS SDK [e.g. 3.999.0]

**Any other comments?**

**Todo List:**
- [x] All commits are cryptographically signed
- [ ] Updated relevant types
- [ ] Added tests (if applicable)
  - [ ] Unit tests
  - [ ] Fuzz tests
  - [ ] Type tests
  - [ ] Benchmark tests

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -163,9 +163,7 @@ declare type MiddlewareHandler<
 	TContext extends LambdaContext = LambdaContext,
 	TResult = any,
 	TEvent = any,
-> = THandler extends LambdaHandler<TEvent, TResult> // always true
-	? MiddyInputHandler<TEvent, TResult, TContext>
-	: never;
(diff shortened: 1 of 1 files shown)


willfarrell:
> Right now there is a random scenario where this statement is not actually true and it breaks the function typings. Do we need this ternary? If its expected to always be true, can we just remove the ternary?

great question, no idea. hopefully a TS person can comment. Looks fine otherwise. I may merge into the v7 alpha if there is a concern of it breaking things.

PR #1472, opened 2025-09-10
chore: use `test()` helper in type tests

mrazauskas:

This PR replaces comments with the `test()` helper in type tests. The helpers improve structure of the test files. `.only` and `.skip` can be used while debugging.

The diff is mostly white space change. 

*Tips.* When running a single file, TSTyche will print out the structure with passing / failing test cases: `npx tstyche utils`. You can use `--only` and `--skip` directly from CLI: `npx tstyche utils --only choose`.

*Note.* `packages/core/index.test-d.ts` was too complex for me. I left it for follow up PR.

**Todo List:**
- [x] All commits are cryptographically signed
- [ ] Updated relevant types
- [ ] Added tests (if applicable)
  - [ ] Unit tests
  - [ ] Fuzz tests
  - [ ] Type tests
  - [ ] Benchmark tests


packages/http-event-normalizer/index.test-d.ts
```diff
@@ -1,11 +1,13 @@
...
+	expect(middleware).type.toBe<middy.MiddlewareObj<Event>>();
+});
 
-// use with all options
-middleware = httpEventNormalizer();
-expect(middleware).type.toBe<middy.MiddlewareObj<Event>>();
+test("use with all options", () => {
+	const middleware = httpEventNormalizer();
+	expect(middleware).type.toBe<middy.MiddlewareObj<Event>>();
+});
```
mrazauskas, line 13: This one looks like a duplicate of the above test. Or I missed something?

PR #1611, opened 2026-04-08
fix: compatibility of CJS TypeScript with @middy/core exports

bilalq:

I have durable function lambdas I maintain that are using middy 7. Unfortunately, ESM support is still tricky with many setups, so we're using CJS build artifacts. At runtime, all our code works. However, because of the way type definitions load in, we get type errors.

Up till now, I've been bypassing this with a `@ts-ignore`, but that's kind of janky and would ideally not be needed.

TypeScript projects using `"module": "commonjs"` (which defaults to `moduleResolution: "node10"`) cannot resolve
`@middy/core/DurableContext` because the CJS resolution algorithm ignores the exports field entirely. It looks for a literal file at `node_modules/@middy/core/DurableContext.d.ts`, but the actual file was named `executionModeDurableContext.d.ts`.

This aliasing seems kind of unnecessary to begin with though. We could get CJS compatibility just by not doing it.

By renaming the files to match the subpath export names, both resolution strategies work:
- node10 (CJS): finds DurableContext.d.ts directly by filename
- node16/nodenext/bundler: resolves via the exports map as before

The public API of the exports remains unchanged. This should be fully backwards compatible without issue.

No

Example of failing type check:

```
lambda/handlers/my-workflow.ts:3:45 - error TS2307: Cannot find module '@middy/core/DurableContext' or its corresponding type declarations.
  There are types at '/Users/bilalq/example-project/node_modules/@middy/core/executionModeDurableContext.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.

3 import { executionModeDurableContext } from '@middy/core/DurableContext'
                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in lambda/handlers/my-workflow.ts:3
```

**Environment:**
 - Node.js: 24
 - Middy: 7.x
 - AWS SDK 3.1026.0

**Any other comments?**

Example tsconfig:

```json
{
  "compilerOptions": {
    "target": "ES2024",
    "module": "commonjs",
    "lib": ["es2024"],
    "esModuleInterop": true,
    "noEmit": true,
    "declaration": true,
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "noUnusedLocals": false,
    "noUnusedParameters": false,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": false,
    "inlineSourceMap": true,
    "inlineSources": true,
    "experimentalDecorators": true,
    "strictPropertyInitialization": false,
    "typeRoots": ["./node_modules/@types"],
    "skipLibCheck": true,
    "resolveJsonModule": true,
    "incremental": true,
    "tsBuildInfoFile": ".tsbuildinfo"
  },
  "exclude": ["node_modules", "cdk.out"]
}
```

**Todo List:**
- [ ] All commits are cryptographically signed
- [X] Feature/Fix fully implemented
- [X] Updated relevant types
- [X] Added tests (if applicable)
  - [X] Unit tests
  - [ ] Fuzz tests
  - [ ] Type tests
  - [ ] Benchmark tests


willfarrell:
Thanks for taking the time to open this PR. I appreciate your solution and ensure backward compatibility. However, we committed to not support cjs years ago (nudge everyone in a more performant direction and simplify maintenance). Having all the executionMode files start with the same prefix is purposeful.

> TypeScript projects using "module": "commonjs" (which defaults to moduleResolution: "node10")

This seems like a real concern. I would recommend submitting a PR to typescript to address this.

PR #1614, opened 2026-04-15
fix(http-urlencode-body-parser): preserve duplicate keys as arrays using getAll()

### LINKED ISSUE #1613: handle duplicate key in http-urlencode-body-parser
**Describe the bug**
This [commit](https://github.com/middyjs/middy/commit/14fcc90596e5105e1e813e75e9a3a85932c7654a#diff-9f54fdd2ca6721ab5e035b6f5daf45886bf92c5f828073f77605fd2dc9861a99) remove the usage of `node:querystring` module. This breaks the behaviour of how duplicated keys are parsed.

The node module parses duplicated keys into array (https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options), while the new Object.fromEntries logic only retains the last value and drop all other values

**To Reproduce**

```
test("It should decode duplicate keys in url encoded requests", async (t) => {
	const handler = middy((event, context) => {
		return event; // propagates the body as response
	});

	handler.use(urlEncodeBodyParser());

	// invokes the handler
	const body = "key=i&key=j";
	const event = {
		headers: {
			"Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
		},
		body,
	};

	const processedEvent = await handler(event, defaultContext);

	deepStrictEqual(
		processedEvent.body,
		Object.assign(Object.create(null), {
			"key": ["i", "j"],  // actual result is only "key": "j"
		}),
	);
});

```

**Expected behaviour**
For backward compatibility, it should have same behaviour as `querystring.parse`, or an option to opt-in / opt-out.

Or if it's intended then I think there should be updates to the doc / change log.

- Node.js: [e.g. 22]
- Middy: [e.g. 7.2.1]
- AWS SDK [e.g. 3.0.0]

**Additional context**

cloudyun888:
### Problem
The  middleware was using  to parse URL-encoded bodies. However,  with  only keeps the **last** value when a key appears multiple times.

For example, with , the result would be  instead of .

### Solution
Replace the single  call with  which returns an array of all values for each key.

This ensures duplicate keys are preserved as arrays, which is the expected behavior for form-encoded data with multiple values for the same key.

### DIFF
--- a/packages/http-urlencode-body-parser/index.js
+++ b/packages/http-urlencode-body-parser/index.js
@@ -28,7 +28,12 @@ const httpUrlencodeBodyParserMiddleware = (opts = {}) => {
 		}
 
 		const data = decodeBody(request.event);
-		const parsedBody = Object.fromEntries(new URLSearchParams(data));
+		const parsedBody = Object.fromEntries(
+    Array.from(new URLSearchParams(data).keys()).map((key) => [
+      key,
+      new URLSearchParams(data).getAll(key),
+    ])
+  );
 
 		// Check if it didn't parse
 		if (parsedBody?.[body] === "") {


willfarrell:
Thanks opening a PR. We've add in your fix in the latest.

PR #1620, opened 2026-04-17
fix(core): lazy-load durable SDK and mark as optional peer to fix bundler errors

### LINKED ISSUE #1548: Middy 7 uses @aws/durable-execution-sdk-js, but not marked as external package
**Describe the bug**
When I try to use Middy 7.0.0, I get this error:

```
X [ERROR] Could not resolve "@aws/durable-execution-sdk-js"

    node_modules/.pnpm/@middy+core@7.0.0/node_modules/@middy/core/executionModeDurableContext.js:1:37:
      1 │ import { withDurableExecution } from "@aws/durable-execution-sdk-js";
        ╵                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "@aws/durable-execution-sdk-js" as external to exclude it from the bundle,
  which will remove this error and leave the unresolved path in the bundle.
```

**To Reproduce**
- Simply upgrade from 6.x.x to 7.0.0 and build

**Expected behaviour**

I'm no Node Dependency expert, but I believe `@aws/durable-execution-sdk-js` should be marked as an peerDependencies.

- Node.js: 22
- Middy: 7.0.0
- AWS SDK 3.962.0

Suresh-2017:
… peer

---
name: fix(core): lazy-load durable SDK and mark as optional peer to fix bundler errors
about: fix(core): lazy-load durable SDK and mark as optional peer to fix bundler errors
title: ''fix(core): lazy-load durable SDK and mark as optional peer to fix bundler errors"
labels:  dependencies | GitHub_Actions
assignees: ''Suresh-2017"
Email: reachpadala@gmail.com

---

@middy/core v7 introduced Durable Functions support but imported @aws/durable-execution-sdk-js as a static top-level import and listed it as a hard dependency. This caused two problems for users who do not use Durable Functions:

Build time failure — bundlers (esbuild, webpack, etc.) fail with Could not resolve "@aws/durable-execution-sdk-js" because they try to follow every static import at build time, even if the code path is never used.
Unnecessary dependency — the SDK was installed for every @middy/core user via dependencies, adding bundle weight for users who never opt into Durable Functions.

Changes made:

executionModeDurableContext.js — replaced the static import with a lazy await import(...) inside a try/catch. If the SDK is absent at runtime, a clear actionable error is thrown pointing to the missing package.
package.json — removed @aws/durable-execution-sdk-js from dependencies, added it to peerDependencies with peerDependenciesMeta.optional: true so it is only installed by users who explicitly need it.
index.d.ts — added JSDoc @throws comment to executionModeDurableContext documenting the optional peer requirement.

Closes #1548

**Environment:**
 Node.js: 24
Middy: 7.2.3
AWS SDK: not applicable (issue is with @aws/durable-execution-sdk-js peer)

**Any other comments?**
Users who do use Durable Functions simply need to add @aws/durable-execution-sdk-js as a direct dependency in their own project. If they call executionModeDurableContext without installing it, they will now get a clear error message.

**Todo List:**
- [ ] All commits are cryptographically signed
- [Yes ] Feature/Fix fully implemented
- [ Yes ] Updated relevant types
- [Yes  ] Added tests (if applicable)
  - [Yes  ] Unit tests
  - [ ] Fuzz tests
  - [ ] Type tests
  - [ ] Benchmark tests
- [Yes  ] Updated relevant documentation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9912,7 +9912,8 @@
 			"workspaces": [
 				"tests/deps/*",
 				"tests/projects/*",
-				"tests/projects/workspace/packages/*"
+				"tests/projects/workspace/packages/*",
+				"tests/executionModeDurableContext.test.js*"
 			],
 			"peerDependencies": {
 				"vite": "^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0"
--- a/packages/core/executionModeDurableContext.js
+++ b/packages/core/executionModeDurableContext.js
@@ -1,7 +1,13 @@
 // Copyright 2017 - 2026 will Farrell, Luciano Mammino, and Middy contributors.
 // SPDX-License-Identifier: MIT
-import { withDurableExecution } from "@aws/durable-execution-sdk-js";
 import { executionContextKeys, lambdaContextKeys } from "@middy/util";
+
+let withDurableExecution;
+try {
+	({ withDurableExecution } = await import("@aws/durable-execution-sdk-js"));
+} catch {
+	withDurableExecution = undefined;
+}
 
 export const executionModeDurableContext = (
 	{ middyRequest, runRequest },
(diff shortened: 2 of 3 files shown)


willfarrell:
I don't understand what the purpose of this PR is. The SDK is already a peerDep, and is only needed when you import the related executionMode.

PR #1662, opened 2026-07-10
Fix: #1661

### LINKED ISSUE #1661: Allow .use() middlewares to wrap the handler invocation (async-context isolation)
**Is your feature request related to a problem? Please describe.**

Middlewares that need a per-invocation async context - `AsyncLocalStorage.run()`, `cls-hooked` `namespace.run()`, OpenTelemetry context, AWS X-Ray `captureAsyncFunc` - have no way to establish one around the handler. This became load-bearing with [Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances.html), where invocations are multiplexed concurrently into one execution environment and shared-slot state leaks across invocations. In Powertools for AWS Lambda's tracer middleware this misattributes trace annotations and response payloads across invocations - potentially across tenants (aws-powertools/powertools-lambda-typescript#5434, upstream aws/aws-xray-sdk-node#760).

The `before`/`after` hooks structurally cannot do this. In `runRequest` (v7.7.0 `index.js`), `await runMiddlewares(request, beforeMiddlewares, plugin)` creates its continuation *before* any hook body executes. Async-context machinery (both `AsyncLocalStorage` and `cls-hooked`) snapshots context at async-resource *creation* time, so when that continuation resumes to call `lambdaHandler(...)`, it restores the pre-hook snapshot and discards whatever the hook entered - for `enterWith()` (the workaround suggested in #1232) exactly as for `cls-hooked` `namespace.enter()`.

We verified this empirically on real AWS Lambda (Node.js 22, native async/await). A diagnostics middleware entered a fresh `cls-hooked` context in `before`; per-invocation CloudWatch logs show the handler running with the *pre-hook* context and the process-shared segment slot:

```
{"diag":"pre-tracer-before","activeCtx":"ctx#0","segmentName":"facade","isFacade":true}   <- hook enters new context here
{"diag":"handler-start",    "activeCtx":"ctx#0","segmentName":"facade","isFacade":true}   <- handler: entered context GONE
```

Note: transpiled test runners (vitest/jest) alter the async-resource topology and make this pattern *appear* to work - it only fails under native async/await, i.e. in production.

The plugin `beforeHandler`/`afterHandler` hooks don't help either: they're constructor-only (a `.use()` middleware can't register them), and they sit on opposite sides of the `await`, so paired enter/exit APIs corrupt the context stack (`cls-hooked` `exit()` splices the wrong frame when the active context differs).

**Describe the solution you'd like**

Let a middleware optionally provide a handler wrapper that middy composes around the `lambdaHandler(...)` call site (LIFO, innermost = last registered):

```js
const tracerMiddleware = () => ({
  before,
  after,
  onError,
  wrapHandler: (next) => (event, context, opts) =>
    namespace.runPromise(() => next(event, context, opts)),
})
```

Because the wrapper executes synchronously at the call site, the handler's async chain captures the entered context, and teardown happens inside the wrapped chain - no cross-hook bookkeeping. We validated this exact mechanism on real Lambda by pre-wrapping the handler before passing it to `middy()`: per-invocation contexts (`ctx#1`, `ctx#2`, `ctx#3` across three invocations, cold and warm) propagated correctly through the handler and instrumented calls, with the base context restored cleanly afterwards.

**Describe alternatives you've considered**

- `AsyncLocalStorage.enterWith()` in `before` (#1232 suggestion): fails under native async/await, verified on real Lambda (logs above)
- `cls-hooked` `namespace.enter()`/`exit()` split across `before`/`after`: same propagation failure, plus context-stack corruption
- plugin `beforeHandler`/`afterHandler`: not reachable from `.use()`, same across-`await` problem
- documenting "wrap your handler before `middy()`": works (it's how we validated), but the behavior can't ship as a middleware, which defeats the composition model

**Additional context**

<details>
<summary>Minimal reproduction - deployable function code for both experiments</summary>

Deploy on a Node.js 22.x Lambda function with X-Ray active tracing enabled. Only dependencies: `@middy/core` and `aws-xray-sdk-core`. Invoke a few times and read the JSON `diag` lines in CloudWatch.

**Experiment 1 - context entered in `before` (does NOT propagate):**

```js
// index.mjs
import xray from 'aws-xray-sdk-core'

const namespace = xray.getNamespace()
const ids = new WeakMap()
let n = 0
const ctxId = (c) => {
  if (!c) return 'none'
  if (!ids.has(c)) ids.set(c, `ctx#${n++}`)
  return ids.get(c)
}
const snap = (label) => {
  const segment = xray.getSegment()
  console.log(JSON.stringify({
    diag: label,
    activeCtx: ctxId(namespace.active),
    segmentName: segment?.name,
    isFacade: segment?.facade === true,
  }))
}

const contextMiddleware = () => ({
  before: (request) => {
    snap('before:pre-enter')
    const clsContext = namespace.createContext()
    namespace.enter(clsContext) // establish per-invocation context
    const facade = xray.getSegment()
    const subsegment = facade.addNewSubsegment(`## ${process.env._HANDLER}`)
    xray.setSegment(subsegment) // writes into the entered context
    request.internal.clsContext = clsContext
    request.internal.subsegment = subsegment
    snap('before:post-enter') // new ctx + '## handler' subsegment, as expected
  },
  after: (request) => {
    request.internal.subsegment?.close()
    namespace.exit(request.internal.clsContext)
  },
})

export const handler = middy(async () => {
  snap('handler') // pre-hook ctx + facade segment: the entered context is gone
  return { ok: true }
}).use(contextMiddleware())
```

Observed (every invocation):

```
{"diag":"before:pre-enter", "activeCtx":"ctx#0","segmentName":"facade","isFacade":true}
{"diag":"before:post-enter","activeCtx":"ctx#1","segmentName":"## index.handler","isFacade":false}
{"diag":"handler",          "activeCtx":"ctx#0","segmentName":"facade","isFacade":true}
```

Anything the handler (or auto-instrumented SDK/HTTP calls) does resolves the shared slot, not the invocation's subsegment. The same result is obtained with `AsyncLocalStorage.enterWith()` in place of `namespace.enter()`.

**Experiment 2 - wrapper composed at the handler call site (works):**

```js
// index.mjs
import xray from 'aws-xray-sdk-core'

// ...same ctxId/snap helpers as above...

const wrapHandler = (next) => (event, context, opts) =>
  xray.captureAsyncFunc(`## ${process.env._HANDLER}`, async (subsegment) => {
    snap('wrap-enter')
    try {
      return await next(event, context, opts)
    } finally {
      snap('wrap-exit')
      subsegment?.close()
    }
  })

export const handler = middy(
  wrapHandler(async () => {
    snap('handler') // per-invocation ctx + '## handler' subsegment
    return { ok: true }
  })
).use({
  before: () => snap('middleware-before'),
  after: () => snap('middleware-after'),
})
```

Observed (three invocations, cold and warm - a fresh context per invocation, hooks still on the base context, clean restore):

```
{"diag":"middleware-before","activeCtx":"ctx#0","segmentName":"facade","isFacade":true}
{"diag":"wrap-enter",       "activeCtx":"ctx#1","segmentName":"## index.handler","isFacade":false}
{"diag":"handler",          "activeCtx":"ctx#1","segmentName":"## index.handler","isFacade":false}
{"diag":"wrap-exit",        "activeCtx":"ctx#1","segmentName":"## index.handler","isFacade":false}
{"diag":"middleware-after", "activeCtx":"ctx#0","segmentName":"facade","isFacade":true}
```

The proposed `wrapHandler` middleware member would let this composition happen inside middy instead of requiring users to wrap their handler manually.

</details>

Prior art: #1232 requested `store.run` support and was closed with the `enterWith` workaround, which the evidence above shows to be unreliable in production.

Disclosure: this investigation was conducted with the help of an AI assistant, supervised by a Powertools for AWS Lambda maintainer; all findings were verified against deployed AWS Lambda functions and the logs above are from those deployments.

willfarrell:
Closes #1661

cc @dreamorosi  @svozza

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -151,10 +151,25 @@ const runRequest = async (
 		request.context.getRemainingTimeInMillis ||
 		request.context.lambdaContext?.getRemainingTimeInMillis;
 	const timeoutEarly = plugin.timeoutEarly && getRemainingTimeInMillis;
+	const beforeMiddlewareHook = plugin.beforeMiddleware;
+	const afterMiddlewareHook = plugin.afterMiddleware;
 
 	try {
-		if (beforeMiddlewares.length) {
-			await runMiddlewares(request, beforeMiddlewares, plugin);
+		for (let i = 0, len = beforeMiddlewares.length; i < len; i++) {
+			const nextMiddleware = beforeMiddlewares[i];
+			if (beforeMiddlewareHook) beforeMiddlewareHook(nextMiddleware.name);
+			let res = nextMiddleware(request);
+			if (res instanceof Promise) res = await res;
+			if (afterMiddlewareHook) afterMiddlewareHook(nextMiddleware.name);
+			// short circuit chaining and respond early
+			if (typeof res !== "undefined") {
+				request.earlyResponse = res;
+			}
+			// earlyResponse pattern added in 6.0.0 to handle undefined values
+			if ("earlyResponse" in request) {
+				request.response = request.earlyResponse;
+				break;
+			}
 		}
 
 		// Check if before stack hasn't exit early
@@ -213,8 +228,19 @@ const runRequest = async (
 			}
 
 			plugin.afterHandler();
-			if (afterMiddlewares.length) {
-				await runMiddlewares(request, afterMiddlewares, plugin);
+			for (let i = 0, len = afterMiddlewares.length; i < len; i++) {
+				const nextMiddleware = afterMiddlewares[i];
+				if (beforeMiddlewareHook) beforeMiddlewareHook(nextMiddleware.name);
+				let res = nextMiddleware(request);
+				if (res instanceof Promise) res = await res;
+				if (afterMiddlewareHook) afterMiddlewareHook(nextMiddleware.name);
+				if (typeof res !== "undefined") {
+					request.earlyResponse = res;
+				}
+				if ("earlyResponse" in request) {
+					request.response = request.earlyResponse;
+					break;
+				}
 			}
 		}
 	} catch (err) {
@@ -227,7 +253,20 @@ const runRequest = async (
 		request.response = undefined;
 		request.error = err;
 		try {
-			await runMiddlewares(request, onErrorMiddlewares, plugin);
+			for (let i = 0, len = onErrorMiddlewares.length; i < len; i++) {
+				const nextMiddleware = onErrorMiddlewares[i];
+				if (beforeMiddlewareHook) beforeMiddlewareHook(nextMiddleware.name);
+				let res = nextMiddleware(request);
+				if (res instanceof Promise) res = await res;
+				if (afterMiddlewareHook) afterMiddlewareHook(nextMiddleware.name);
+				if (typeof res !== "undefined") {
+					request.earlyResponse = res;
+				}
+				if ("earlyResponse" in request) {
+					request.response = request.earlyResponse;
+					break;
+				}
+			}
 		} catch (err) {
 			// Save error that wasn't handled. When an onError middleware rethrows
 			// `request.error`, err === request.error; attaching it to itself would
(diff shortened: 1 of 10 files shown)


packages/core/index.js
```diff
@@ -151,10 +151,25 @@ const runRequest = async (
 		request.context.getRemainingTimeInMillis ||
 		request.context.lambdaContext?.getRemainingTimeInMillis;
 	const timeoutEarly = plugin.timeoutEarly && getRemainingTimeInMillis;
+	const beforeMiddlewareHook = plugin.beforeMiddleware;
+	const afterMiddlewareHook = plugin.afterMiddleware;
 
 	try {
-		if (beforeMiddlewares.length) {
-			await runMiddlewares(request, beforeMiddlewares, plugin);
+		for (let i = 0, len = beforeMiddlewares.length; i < len; i++) {
```
svozza, line 158: Is the fix here inlining the `runMiddlewares` function? How does it work?

willfarrell: That is correct.

runMiddlewares used to always await middleware steps. To improve perf for when the middleware step isn't a promise we supported a conditional approach. I believe this introduced this regression that your team found. I don't understand why this issue is happening at a deep understanding level yet, but was able to prove and resolve with TDD, and that's good enough for me to include in a release. The removal of runMiddlewares does reduce maintainability a little, not to a level I'm concerned about.

willfarrell: I just want to make sure this fork does resolve your root issue.

PR #1679, opened 2026-08-19
feat: support v2 of protocol x402

### LINKED ISSUE #1678: http-x402: failed payments return 402 without a PAYMENT-RESPONSE header
`@middy/http-x402` sets the `PAYMENT-RESPONSE` header when settlement succeeds, but not on any of the failure paths. Under the v2 HTTP transport spec that is where a client is supposed to learn why its payment failed, so a client that reads only headers gets a bare 402 with no x402 information at all.

## What the code does

In `packages/http-x402/index.js`, the success path ends with:

```js
request.internal.x402 = { ...stored, payer: ..., transaction: ..., network: ... };
request.response.headers["PAYMENT-RESPONSE"] = encodeHeader(settleResult);
```

The settle-failure path returns before reaching it:

```js
if (!settleResult.success) {
  request.response.statusCode = 402;
  request.response.headers["Content-Type"] = "application/json";
  request.response.body = JSON.stringify({ x402Version: 2, error: settleResult.errorReason });
  return;
}
```

`respond402` (used for decode errors, scheme/network mismatch and verify failure) likewise sets only a status code, `Content-Type` and a JSON body. Counting occurrences in `index.js`: `PAYMENT-REQUIRED` is set once, `PAYMENT-RESPONSE` once, and that one is the success branch.

## Why that looks unintended rather than deliberate

The initial challenge already does the header form correctly, setting **both** the header and the body:

```js
request.response.headers["PAYMENT-REQUIRED"] = encodeHeader(paymentRequired);
request.response.body = JSON.stringify(paymentRequired);
```

So the package clearly intends header transport. It is only the failure branches that drop it.

[`specs/transports-v2/http.md`](https://github.com/x402-foundation/x402/blob/main/specs/transports-v2/http.md) is explicit about which one clients are entitled to rely on:

> Response bodies are a server implementation concern. All x402 protocol information is communicated through headers (`PAYMENT-REQUIRED`, `PAYMENT-SIGNATURE`, `PAYMENT-RESPONSE`).

and its **Example (Failure)** is exactly this case: a 402 whose body is `{}` and whose `PAYMENT-RESPONSE` header decodes to `{"success": false, "errorReason": "insufficient_funds", "transaction": "", "network": "...", "payer": "0x857b..."}`.

The value needed is already in hand at both sites, so the fix looks like one line in the settle-failure branch:

```js
request.response.headers["PAYMENT-RESPONSE"] = encodeHeader(settleResult);
```

The `respond402` paths are less clear cut, since verify failure has no `SettlementResponse` to report.

## Test coverage

`PAYMENT-RESPONSE` appears twice in `index.test.js`, both inside `test("verify and settle pass - adds PAYMENT-RESPONSE header")`. The failure tests are thorough about status code, `Content-Type` and body shape (`settle-failure 402 sets Content-Type application/json`, `verify-invalid 402 body is x402Version 2 with facilitator invalidReason`, and others) but none of them assert anything about response headers. So the current behaviour on the failure paths is unasserted either way.

## A smaller, more arguable point

The spec's error table maps `Invalid Payment` (malformed payment payload or requirements) to **400**, and reserves 402 for `Payment Required` and `Payment Failed`. `decodeHeader` throwing on a malformed payload currently produces a 402 `invalid_payment`. That may well be a deliberate choice to avoid leaking parse detail, so I mention it separately rather than as part of the report above.

## What I have not verified

- Runtime behaviour. I have read the source and tests, and have not run the package or a live facilitator.
- Whether `@x402/core`'s own client falls back to reading the JSON body when the header is absent, which would reduce this to a spec-conformance issue rather than a broken flow for that particular client.
- Which revision of the v2 spec this package targets, and whether you treat the error table as normative.

Filed by an automated agent. Every claim above was read out of the linked files rather than inferred, and I have tried to mark the places where I am reasoning about intent instead of reporting behaviour.

willfarrell:
close #1678

### DIFF
--- a/packages/http-x402/index.d.ts
+++ b/packages/http-x402/index.d.ts
@@ -1,8 +1,11 @@
 // Copyright 2017 - 2026 will Farrell, Luciano Mammino, and Middy contributors.
 // SPDX-License-Identifier: MIT
 import type middy from "@middy/core";
+import type { APIGatewayEvent, APIGatewayProxyEventV2 } from "aws-lambda";
 
-export interface Options {
+export type RequestEvent = APIGatewayEvent | APIGatewayProxyEventV2;
+
+interface BaseOptions {
 	FacilitatorClient?: new (config: {
 		url?: string;
 	}) => {
@@ -10,17 +13,25 @@ export interface Options {
 		settle(payload: unknown, requirements: unknown): Promise<unknown>;
 	};
 	facilitatorUrl?: string;
-	price: number;
+	versions?: readonly (1 | 2)[];
+	price?: number | string;
+	amount?: string;
 	decimals?: number;
 	network?: string;
 	payTo: string;
 	asset: string;
 	description?: string;
 	mimeType?: string;
-	human?: (request: middy.Request) => boolean;
+	extra?: Record<string, unknown>;
+	human?: (request: middy.Request<RequestEvent>) => boolean;
 }
 
-declare function httpX402(options: Options): middy.MiddlewareObj;
+export type Options = BaseOptions &
+	({ price: number | string } | { amount: string });
+
+declare function httpX402<EventType extends RequestEvent = RequestEvent>(
+	options: Options,
+): middy.MiddlewareObj<EventType, unknown, Error>;
 
 export declare function httpX402ValidateOptions(
 	options?: Record<string, unknown>,
--- a/packages/http-x402/index.fuzz.js
+++ b/packages/http-x402/index.fuzz.js
@@ -38,6 +38,23 @@ test("fuzz payment-signature header values never crash", async () => {
 	);
 });
 
+test("fuzz x-payment header values never crash", async () => {
+	const handler = middy(() => ({ statusCode: 200, body: "ok" })).use(
+		httpX402(fuzzOptions),
+	);
+
+	await fc.assert(
+		fc.asyncProperty(fc.string(), async (headerValue) => {
+			const response = await handler(
+				{ headers: { "x-payment": headerValue } },
+				defaultContext,
+			);
+			strictEqual(typeof response.statusCode, "number");
+		}),
+		{ numRuns: 10_000, examples: [] },
+	);
+});
+
 test("fuzz arbitrary event objects never crash", async () => {
 	const handler = middy(() => ({ statusCode: 200, body: "ok" })).use(
 		httpX402(fuzzOptions),
--- a/packages/http-x402/index.js
+++ b/packages/http-x402/index.js
@@ -9,6 +9,9 @@ const pkg = `@middy/${name}`;
 const defaults = {
 	FacilitatorClient: HTTPFacilitatorClient,
 	facilitatorUrl: "https://x402.org/facilitator",
+	// TODO remove in v8: drop 1 from the default versions so v1 acceptance
+	// becomes opt-in (smaller attack surface, no legacy facilitator path).
+	versions: [1, 2],
 	price: undefined,
 	amount: undefined,
 	decimals: 6,
@@ -17,6 +20,7 @@ const defaults = {
 	asset: undefined,
 	description: "",
 	mimeType: "application/json",
+	extra: undefined,
 	human: undefined,
 };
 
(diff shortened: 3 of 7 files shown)


Circadian-agent:
Tested as asked. Taking the v1-era framing first: that explains the `x402Version: 2` literal I was reading as a v2 implementation with defects, and "built for v1, v2 did not exist yet" is a better description of what I found than the one I filed under.

## What I checked in this branch, and it holds up

- **Top-level `resource`.** `buildPaymentRequired` emits `{ url, description, mimeType }` as a `ResourceInfo` object, and `accepts` is now the seven-field `PaymentRequirements` with the three v1 fields gone from it.
- **Every v2 pre-handler rejection carries the challenge.** `respondPaymentRequired` sets `PAYMENT-REQUIRED` and mirrors the body on the undecodable-header, version-mismatch, no-match, verify-throw and verify-invalid paths. That is the retry property I was arguing for, and the comment states it better than my issue did.
- **Settlement failure now reports.** Both the `!settleResult.success` branch and the throw branch set the response header, and the v1/v2 split onto `X-PAYMENT-RESPONSE` and `PAYMENT-RESPONSE` is right.
- **v1 is genuinely preserved,** not approximated: `maxAmountRequired`, `resource` as a string, `outputSchema`, and the challenge in the body rather than a header.
- The `isBase64Encoded = false` reset is a real bug I had not thought about, and it would have been ugly to debug.

## One thing I would change before this merges

The settle-throw branch hard-codes the hash while passing the reason through verbatim:

```js
const settleResponse = {
  success: false,
  errorReason: settleReason(error),   // returns error.errorReason unchanged
  transaction: "",
  network: storedRequirements.network,
};
```

The spec singles out one reason code where that combination is not allowed:

> **`settlement_pending`**: The settlement transaction was broadcast but its confirmation could not be established (e.g. a node/RPC error or timeout while waiting for the receipt). [...] A `SettleResponse` with this `errorReason` **MUST carry a non-empty `transaction`** (the broadcast hash) and `network` so the caller can reconcile on chain before deciding whether to retry.

A facilitator that broadcast, then timed out waiting for the receipt, is exactly the case that throws. `settleReason` forwards `settlement_pending` faithfully and the hash next to it is dropped, so the client is told the payment is in an unknown state **and** denied the one value that would let it find out. Of all the reason codes this is the one where a missing field has money attached: reconcile is impossible, and a retry may double-pay.

It is narrow. The `!settleResult.success` branch passes the facilitator's own object through and is fine; only the throw branch synthesises one. Something like:

```js
transaction: typeof error?.transaction === "string" ? error.transaction : "",
```

**Why it is easy to miss:** `settlement_pending` appears zero times in `index.test.js`, and the throw branch is covered through `unexpected_settle_error`, where an empty `transaction` is the correct answer. The defect hides behind the case where the hard-coded value happens to be right.

## What I have not verified

- I have not run the suite. This is read from the branch, and the failing case is one I reasoned to rather than observed.
- **Whether it is reachable today**: `settleReason` shows you expect thrown errors to carry `errorReason`, but I do not know whether `HTTPFacilitatorClient` also surfaces `transaction` on a throw. If it never does, this is a guard against a future facilitator rather than a live bug, and worth a comment either way.
- Whether you consider the non-terminal pending code in scope for this PR at all.

Circadian-agent:
Checked `7a2a074e`. That is the fix, and the tests pin it properly.

- The `settlement_pending` test asserts the hash off the **decoded** `PAYMENT-RESPONSE` header rather than the body, which is the surface that actually has to carry it.
- The non-string `transaction: 42` test is the one I would have asked for second: it pins the `typeof` guard, so removing that check fails a test instead of silently forwarding a number into a string field.
- The same construction is shared with the v1 branch, so `X-PAYMENT-RESPONSE` gets the hash too, without a second code path to keep in sync.

One residual, and I mention it only so nobody later reads it as an omission and "fixes" it:

If a facilitator throws `settlement_pending` **without** a hash, the response still carries `transaction: ""`, which is the state the spec forbids. There is nothing you can do about that from here, since the hash is not yours to invent, and the alternatives are worse. Rewriting the reason to `unexpected_settle_error` would hide a genuinely non-terminal state and could push a client into retrying a payment that is still pending, which is the double-spend this rule exists to prevent. Forwarding the facilitator's reason unchanged and letting the missing field be visibly missing is the right call.

Nothing else from my side. Thanks for turning this around so quickly, and for taking the v1 compatibility route rather than the cheap version bump.

**Not verified:** I still have not run the suite; this is read from the diff and the branch.

